Issue #268, opened 2026-08-27
`print.margin_grouping_spec()` emits half a line and then raises for a kind `cat()` cannot print


sayuks:
## Problem

`print.margin_grouping_spec()` hands the kind it read straight to `cat()`:

```r
kind <- grouping_spec_kind(x)
rule <- find_grouping_kind_rule(kind)
name <- if (is.null(rule)) kind else rule$constructor
cat("<marginplyr grouping specification: ", name, ">\n", sep = "")
```

`find_grouping_kind_rule()` answers `NULL` for anything that is not one
non-`NA` name, so the fallback prints the kind itself — and a kind `cat()`
cannot handle raises there, after the opening of the line has already been
written:

```r
print(structure(list(type = list("a")), class = "margin_grouping_spec"))
#> <marginplyr grouping specification: Error: argument 2 (type 'list') cannot be handled by 'cat'
#>   class: simpleError/error/condition
```

The partial line is the part that makes it worse than a plain refusal: the
reader is left with an unterminated line on their console and an untyped base-R
error naming `cat`, which says nothing about a Grouping specification.

## Relationship to #264

#264 established that the kind can be *read* before it is used, and routed this
site through `grouping_spec_kind()` so an object `$` cannot be asked prints
rather than raising. This is the other half of the same line, and it is not the
same defect: here the read succeeds and the *value* it produced is what `cat()`
refuses. `grouping_spec_kind()` cannot help — it answers with exactly what the
field held.

It is outside #264 for a second reason as well. #264's scope says "the wording
of the printed line for every object that prints today is deliberately
unchanged", and this shape does not print today, so deciding what it should
print is a decision that ticket ruled out making. Measured on this branch, the
behaviour is unchanged from before #264: `git show 51e145f:R/grouping-spec.R`
produces the same partial line and the same error.

## Reproduction / evidence

Measured on R 4.6.1 at `276d73b`, and at `51e145f` before it, with identical
results:

| `type` field | printed |
| --- | --- |
| `list("a")` | `<marginplyr grouping specification: ` then `argument 2 (type 'list') cannot be handled by 'cat'` |
| `1:3` | `<marginplyr grouping specification: 123>` |
| `c("a", "b")` | `<marginplyr grouping specification: ab>` |

The last two are the constraint on any fix, not further defects: they print
today, so #264's scope sentence applies to them and a fix that silences the
first by narrowing the fallback to a character scalar changes them too. Whether
[...]

PR #285, opened 2026-08-28
Bound what a review may raise about prose, and what an agent loads before it starts

### LINKED ISSUE #284: A review can raise a finding about prose that nothing bounds, and nothing bounds what an agent loads before it starts
## Problem

Two things this repository does not bound, both measured on `597f8bc`.

**A review's prose findings.** *Answering a review* in `design/architecture.md`
says what an answer must land with and nothing about what may be raised. A
reader given a diff it cannot fault reports a phrasing, the answer rewrites the
sentence, and the next round has the rewrite to measure. #282 recorded that a
termination condition cannot be stated and falsified three formulations of one;
what it did not try is bounding the finding rather than the round.

**The always-loaded context.** `CLAUDE.md` is `@AGENTS.md`, so 618 lines are
read before every task whether or not the task reaches them. Nothing in the
repository fails when that grows — prose costs no test and breaks no build —
which is how it reached 618 lines unmeasured. Of its eleven sections, one
governs a branch an ordinary commit enters: *Code comments* fires on 64 of the
last 100 non-merge commits. *Installation instructions* is 62 lines and fires
when `Config/marginplyr/cran-status` moves off `unpublished`, which no commit
in 412 has done.

A third thing is smaller and follows from the first. `mattpocock-skills:code-review`
looks for `docs/agents/issue-tracker.md`; here that path is inside altdoc's
generated site, gitignored and holding no tracked file, while the tracker doc
is `design/agents/issue-tracker.md`. Reviews have found it anyway, because
`AGENTS.md` names the real path and is loaded before the skill's search runs.
Nothing writes that down, so nothing stops `/setup-matt-pocock-skills` being
run and writing into what the next site build deletes.

## Evidence

Measured on `597f8bc`, reproducible from the tree:

- `AGENTS.md` is 37,043 bytes, read every turn.
- Verbatim duplication across the prose corpus is 6.1% (11,177 of 183,038
  words, 10-word shingles), so volume rather than duplication is what there is
  to act on.
- Between 2026-08-14 and 2026-08-28, R code grew 988 lines and prose 7,380;
  eight of the last twelve merged PRs changed no line of R.
- `design/architecture.md`'s test inventory named 18 of the 33 test files
  present, against 8 of the 9 there were when `957356c` wrote it.
- The ledger's 67 findings are each false, a drifting duplicate, a broken rule,
  or a required thing missing. None is of how a sentence reads.

## What this fixes

1. *Answering a review* gains what a prose finding may be, and two answers:
   deletion, and a ticket for what this branch's Acceptance does not reach.
2. `.github/scripts/verify-context-budget.R` derives the always-loaded closure
   from `CLAUDE.md` and fails when it passes a measured baseline.
3. `design/agents/code-review.md` maps the review skill's expectations onto
   this repository, beside the two mapping files already there.
4. The test inventory goes, since `tests/testthat/` is what it claimed to be
   and every seam exception in it is a second copy of the test's own comment.

## Acceptance

- A finding about how prose reads is out of scope, stated where any reviewer is
  bound and not only where this skill is described;
- the always-loaded closure has a baseline that a byte over fails, derived
  rather than listed, and asserting its own mechanism before its total;
- `/setup-matt-pocock-skills` is recorded as not run here, with the reason;
- nothing the test inventory held is lost that is not held elsewhere.

## Not in scope

`AGENTS.md`'s own Tier split, the prose #268 left in ADR 0008 and
`test-grouping-interface.R`, and the six verifiers whose workflow step can be
dropped without detection. Each is its own ticket.

sayuks:
Closes #284.

Three commits, +215 / -95. Two bounds and one mapping, in the layers this
repository already has.

## What lands

**`design/architecture.md`, *Answering a review*.** A finding about prose is
one of four — false, a drifting duplicate, a broken rule, or a required thing
missing — and how prose reads is not one. Two answers join the disposition list
for prose: a claim found false is deleted rather than restated, citing
*Code comments* for the rule #283 just gave it; and a finding this branch's
Acceptance does not reach becomes a ticket rather than prose added here.

The four are read off the ledger rather than invented. Its 67 findings are each
one of them — `:359` and `:363` false, `:352` duplicate, `:658` and `:674` a
broken rule, `:297` a required thing missing — and none is of how a sentence
reads. This is in the norm
layer because it binds any reviewer, not only the skill below.

**`.github/scripts/verify-context-budget.R`, run by `lint.yaml`.** The closure
`CLAUDE.md` pulls in is read before every task. It now carries a measured
baseline, and a byte over fails with the remedy in the message. It derives the
closure by following `@` references rather than naming `AGENTS.md`, and asserts
that it read at least one before trusting its total, since a reference reader
that stopped working would report one small file and pass.

It fired for real on this branch when #283's paragraph arrived from `main`. The
baseline moves to include it, and the commit says nothing paid for it: #283 was
merged before this gate existed, so the ratchet starts from the tree that holds
it.

**`design/agents/code-review.md`.** The adaptation layer already maps a skill's
vocabulary onto this repository — `triage-labels.md` has a
`Label in mattpocock/skills` column — and this joins it for
`mattpocock-skills:code-review`. It records that `/setup-matt-pocock-skills` is
not run here, because the `docs/agents/issue-tracker.md` it would write is
inside altdoc's generated site, and that a second run over the same fixed point
reads the first run's answers. `AGENTS.md` gains one pointer line, in the shape
the three beside it already have.

**`## Test seams` loses its inventory.** It named 18 of the 33 test files
present, against 8 of the 9 when `957356c` wrote it. Every seam exception it
held is a second copy — `test-margin-order.R:1`, `test-grouping-plan.R:319`
`:578` `:936`, `test-execution-conditions.R:148` `:611` `:758` `:917`,
`test-share.R:1946` `:2047` — so the rule stays and each reason stays beside
the assertion it licenses. Two ADR citations were not second copies, and moved
into the two tests that observe them.

## What does not land, and why

`## Module responsibilities` names 13 of 23 files and is left alone: the ten it
omits are four public-verb wrappers whose lifecycle `margin-operation.R` owns
and documents, two files that are roxygen with one line of code, and four
modules — 296 lines between them — that are a gap in a map rather than a
duplicate of one.

`AGENTS.md`'s own Tier split, the prose #268 left in ADR 0008 and
`test-grouping-interface.R`, and the six verifiers whose workflow step can be
dropped without detection are each their own ticket. #284 says so.

## Verification

Full suite on this branch: 5,104 passed, 0 failed, 0 skipped, `NOT_CRAN` set.
`lintr::lint_package()` clean. The budget gate was exercised against a `/tmp`
copy: one byte added to `AGENTS.md` fails, an unresolvable `@` reference fails,
and a reference reader that matches nothing fails naming itself rather than the
repository.

## Review

None yet. This is the change that introduces the scope rule, so reviewing it is
the first chance to see whether the rule reaches the reviewer at all — the
limitation the new `design/agents/code-review.md` does not claim to close, since
the skill builds its own sub-agent prompts.

No ledger entry: every changed line is a comment or a document, and the fix is
its own diff under `design/review-dispositions.md`'s prose exemption. The
verifier script is new rather than a fix.

### DIFF
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -1,0 +1,139 @@
+# Fails when the always-loaded agent context grows past the baseline below.
+#
+# `CLAUDE.md` is read before every task an agent runs here, and each `@path` in
+# it pulls that file in too, so the closure is loaded whether or not the task
+# reaches what it says. Nothing else in the repository fails when that closure
+# grows: prose costs no test and breaks no build, which is why it grew to 618
+# lines before anyone measured it. A budget is the only thing that makes an
+# addition cost something at the moment it is made.
+#
+# The baseline is a ratchet and not a target. It records what the closure
+# measured when it was last deliberately set, so an addition is paid for by a
+# deletion until someone decides otherwise and moves it in the same commit that
+# spends it. A number chosen for how much the closure ought to be would be
+# invented; this one is measured.
+#
+# The closure is derived from `CLAUDE.md` rather than listed, for the reason
+# every other gate here derives: a list would name what it named when it was
+# written, and a file added to the closure by a new `@path` would escape the
+# budget it was added to. A derivation that stopped working reports a closure of
+# one small file and passes, so the mechanism is asserted before the total is.
+
+source(".github/scripts/ci-helpers.R")
+
+# What the closure measured when it was last set, and the commit that set it.
+# Moving this means saying in the same commit what was added and what paid for
+# it.
+baseline_bytes <- 38396L
+baseline_note <- "this gate, its entry in AGENTS.md, and #283's paragraph"
+
+entry <- "CLAUDE.md"
+
+# Reads `@path` references the way the harness does: one per line, leading
+# whitespace allowed, the rest of the line taken as the path. A reference inside
+# a fenced block is still a reference for this purpose -- the budget is what the
+# file costs, and a reader that resolves it costs that whether the fence was
+# meant to quote it or not.
+references <- function(path) {
+  lines <- readLines(path, warn = FALSE)
+  found <- regmatches(lines, regexpr("^\\s*@\\S+", lines))
+  sub("^\\s*@", "", found)
+}
+
+closure <- function(root) {
+  seen <- character()
+  pending <- root
+  while (length(pending) > 0L) {
+    path <- pending[[1L]]
+    pending <- pending[-1L]
+    if (path %in% seen) {
+      next
+    }
+    if (!file.exists(path)) {
+      stop(call. = FALSE, sprintf(
+        paste0(
+          "%s names `@%s`, which no file answers. An agent reaches the ",
+          "closure through these references, so one that resolves to nothing ",
+          "is context the task was promised and did not get."
+        ),
+        entry, path
+      ))
+    }
+    seen <- c(seen, path)
+    pending <- c(pending, references(path))
+  }
+  seen
+}
+
+if (!file.exists(entry)) {
+  stop(call. = FALSE, sprintf(
+    "%s is absent, so nothing here can say what an agent loads before a task.",
+    entry
+  ))
+}
+
+files <- closure(entry)
+
+# The mechanism, before the total. `CLAUDE.md` is a pointer file -- it carries
+# one `@` line and no prose -- so a closure of itself alone is a reference
+# reader that stopped reading, not a repository that shrank.
+if (length(files) < 2L) {
+  stop(call. = FALSE, sprintf(
+    paste0(
+      "%s resolved to %d file(s), so no `@` reference was read. The budget ",
+      "below would pass on any repository. Fix the reference reader in this ",
+      "script before trusting what it reports."
+    ),
+    entry, length(files)
+  ))
+}
+
+sizes <- vapply(files, file.size, numeric(1))
+total <- sum(sizes)
+
+report <- sprintf("%8.0f  %s", sizes, files)
+
+# The remedy is part of the message, as `verify-suite-coverage.R`'s is: a gate
+# that only refuses invites the baseline to be raised instead of the argument
+# being moved.
+if (total > baseline_bytes) {
+  stop(call. = FALSE, paste(
+    c(
+      sprintf(
+        paste0(
+          "The always-loaded context is %.0f bytes against a baseline of %d ",
+          "(%s), over by %.0f."
+        ),
+        total, baseline_bytes, baseline_note, total - baseline_bytes
+      ),
+      "",
+      report,
+      "",
+      paste(
+        "An instruction belongs in this closure; the argument for it belongs",
+        "to the file that owns the decision -- an ADR, a workflow comment, a",
+        "verifier's header, or the ticket. Move the argument and cite it. If",
+        "the addition is an instruction that has to be loaded before the task",
+        "reaches it, move the baseline in this file in the same commit, and",
+        "say there what was added and what paid for it."
+      )
+    ),
+    collapse = "\n"
+  ))
+}
+
+write_step_summary(c(
+  "## Always-loaded agent context",
+  "",
+  sprintf(
+    "%.0f bytes of a %d-byte baseline (%s).",
+    total, baseline_bytes, baseline_note
+  ),
+  "",
+  as_summary_block(paste(report, collapse = "\n"))
+))
+
+cat(sprintf(
+  "Verified: the always-loaded context is %.0f bytes, within %d.\n",
+  total, baseline_bytes
+))
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -30,10 +30,9 @@ baseline_note <- "this gate, its entry in AGENTS.md, and #283's paragraph"
 entry <- "CLAUDE.md"
 
 # Reads `@path` references the way the harness does: one per line, leading
-# whitespace allowed, the rest of the line taken as the path. A reference inside
-# a fenced block is still a reference for this purpose -- the budget is what the
-# file costs, and a reader that resolves it costs that whether the fence was
-# meant to quote it or not.
+# whitespace allowed. A reference inside a fenced block is still a reference for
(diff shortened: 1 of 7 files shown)


sayuks:
## Review record

**Snapshot reviewed:** `14340ea3eaa6116ea079acf331785c4615e79422`
**Fixed point:** `41e848a` (merge-base with `main`)
**Reviewer:** `mattpocock-skills:code-review`, two axes in parallel, invoked per `design/agents/code-review.md` on this branch. Scope and dispositions per `design/architecture.md`'s *Answering a review* on this branch.
**Rounds:** 1. Answers were verified by executing each finding's evidence, not by a second invocation — `design/agents/code-review.md:21`.

---

### F1 — the test-suite division claim
**Line:** `design/architecture.md:550-551` — "The test suite divides supporting contracts by module; `tests/testthat/` is what says which file holds which."
**Contract alleged broken:** *Answering a review* category 1, the prose is false. Reported against `8c786f0`, which deletes ~90 lines of inventory and adds two file headers.
**Counterexample offered:** ten test files open on `test_that(` with no file-level statement of the contract they hold — `test-expand-operation.R`, `test-get-col-names.R`, `test-grouping-backends.R`, `test-grouping-interface.R`, `test-margin-id.R`, `test-nest-operation.R`, `test-retail-sales.R`, `test-share.R`, `test-summarize-operation.R`, `test-utils.R`; `test-grouping-plan.R`'s first comment is a header on `ambiguous_s_message`, not on the file.
**Scope:** in scope. A claim false for eleven of thirty-three files is not how a sentence reads.
**Disposition:** rejected with evidence.
**Evidence:** the counterexample establishes that no *header* says it, and the sentence does not claim one does. Nothing in the repository requires a per-file contract header: the companion sentence this commit adds, `design/architecture.md:551-552`, requires a reason only where a test asserts below the seam. What `tests/testthat/` says is its file names and the assertions under them, and for each headerless file named that is enough —

```sh
grep "^test_that" tests/testthat/test-get-col-names.R
grep "^test_that" tests/testthat/test-margin-id.R
```

give `get_col_names` metadata across four backends, and Grouping-set occurrence identifiers across all four verbs, which is what the deleted inventory said of them. The two headers `8c786f0` does add exist to rehome two ADR citations that were not second copies, not to open a convention. The fix the finding asks for — headers on eleven files — is prose added to answer prose, which is the second copy the commit removes.

---

### F2 — the reference reader's header ✅ fixed
**Line:** `.github/scripts/verify-context-budget.R:32-33` at `14340ea` — "one per line, leading whitespace allowed, **the rest of the line taken as the path**", above `regexpr("^\\s*@\\S+", lines)`.
**Contract broken:** *Code comments* in `AGENTS.md` — a comment describes the code it sits beside. Category 1, the prose is false.
**Counterexample:**

```r
sub("^\\s*@", "", regmatches("@my notes.md", regexpr("^\\s*@\\S+", "@my notes.md")))
#> "my"
```

`\S+` stops at the first space, so a reference naming a path with a space is truncated and then reported by the `stop()` at `:53` as a path no file answers — naming `my`, which `CLAUDE.md` never wrote.
**Scope:** in scope, and not bound by the prose rule in any case: this is a comment on executable code, false against the line below it.
**Disposition:** the verifier script. The false clause is deleted rather than corrected, as *Code comments* requires — it was a claim about where the path ends, and the two clauses that remain are true of the regex as written. The code is right and the comment was wrong: a bare `@` reference stops at whitespace in the harness too.
**Evidence:** `ac78c59`. `Rscript .github/scripts/verify-context-budget.R` exits 0, reporting 38396 bytes of a 38396-byte baseline — the edit is to the script, not to the closure, so the baseline is untouched.

---

### F3 — "Running it a second time" as a rule about answering
**Line:** `design/agents/code-review.md:19-21`, against the same file's `:5` ("What may be a finding, and how one is answered, is not here").
**Contract alleged broken:** category 3, it breaks a rule a repository document states — here the file's own. Reported as scope creep against #284's fix 3, "maps the review skill's expectations onto this repository", and against the Acceptance line "stated where any reviewer is bound and not only where this skill is described".
**Counterexample offered:** `:21`'s "Verify a finding against the evidence its answer landed with instead" reads as a rule about how a finding is answered, which `:5` sends to `design/architecture.md`.
**Scope:** in scope.
**Disposition:** rejected with evidence.
**Evidence:** `:21` governs the reviewer's second run, not the answer — its subject is that re-running the skill against the same fixed point re-reads answers already landed, which is a property of this skill's invocation and of nothing else. That is what the adaptation layer owns, and both siblings carry the same shape rather than a table alone: `design/agents/triage-labels.md:15` ("When a skill mentions a role, use the corresponding label string from this table") and `design/agents/issue-tracker.md`'s *Conventions* and *When a skill says "publish to the issue tracker"*. This review is the instance — the round count above is 1 because `:21` was followed.

---

### Verified and not findings

- All four Acceptance bullets land. The budget is derived, asserts its mechanism before its total, and fails a byte over: against a `/tmp` copy, `printf 'x' >> AGENTS.md` exits 1 reporting "over by 1". `wc -c CLAUDE.md AGENTS.md` = 11 + 38385 = 38396, matching `baseline_bytes`. It runs on a bare R, which the `context-budget` job needs since it installs no dependencies.
- Nothing the inventory held is lost. Every seam exception the PR body cites is at its cited line, and so are items it does not cite: ADR 0025's two-part Absorbing-backend reason at `tests/testthat/test-grouping-backends.R:118-132`, the bullet-marker case at `test-execution-conditions.R:832-855`, the Total-share principle at `test-share.R:1703-1706`.
- The skill-mapping claims check against the installed `SKILL.md`; `docs/` is gitignored and `git ls-files docs` is empty.
- Nothing in *Not in scope* was touched.
- No baseline smell has a site in this diff.

**One item for the deferred ticket, not a finding:** the new `context-budget` job joins the class #284 defers — "the six verifiers whose workflow step can be dropped without detection". Deleting `.github/workflows/lint.yaml:19-30` fails nothing. Consistent with the deferral; worth adding to that ticket rather than answered here.

---

**Counts:** 3 findings raised, 1 confirmed and answered, 2 rejected with evidence. 0 findings about how prose reads — the scope bound was passed to both sub-agents and neither returned one.

PR #286, opened 2026-08-28
Refuse a grouping kind that cannot be classified, in marginplyr's own words

### LINKED ISSUE #280: The guards raise an untyped error for a kind whose `is.na()` or `length()` method raises
## Problem

`validate_grouping_spec_early()` reads the kind through `grouping_spec_kind()`,
which catches a `$` that raises (#262), and then hands what it read to
`find_grouping_kind_rule()`:

```r
kind <- grouping_spec_kind(grouping_spec)
rule <- find_grouping_kind_rule(kind)
```

`find_grouping_kind_rule()` asks `is.character(kind)`, `length(kind)`, and
`is.na(kind)`. The last two are generic, so a kind carrying a class of its own
answers them with the object's own methods — and a method that raises instead
of answering raises out of the guard, in place of the
`Invalid grouping specification.` refusal below it. That is #262's defect at a
different hazard: there the object could not answer for its field, here it
answers with a value that cannot be classified.

## Reproduction / evidence

Measured on R 4.6.1 at `cc5846b`:

```r
registerS3method(
  "is.na", "raising_kind",
  function(x, ...) stop("is.na raises"),
  envir = asNamespace("base")
)
spec <- structure(
  list(type = structure("set", class = "raising_kind"), args = list()),
  class = "margin_grouping_spec"
)
inspect_grouping(data.frame(a = 1), .grouping = spec)
#> Error: is.na raises
```

| forged `type` | class raised | message |
| --- | --- | --- |
| `is.na()` method raises | `simpleError/error/condition` | `is.na raises` |
| `length()` method raises | `simpleError/error/condition` | `length raises` |
| `"nope"` (for comparison) | `marginplyr_error/rlang_error/error/condition` | `Invalid grouping specification.` |

Both reproduce through `inspect_grouping()` and through a Margin verb.

## Relationship to #262 and #268

#262 caught the read at both guards, and #264 caught it at
`print.margin_grouping_spec()`. #268 then closed the printer's other two
questions: it decides the classification inside a catch, for exactly the
methods above, and records in ADR 0008 that the guards are deliberately left
alone — what a guard does about a raising method is a change to the condition
a call raises rather than to a printed line, and #262's answer for the read
there was a refusal rather than a caught value.

So the printer completes its line where the guards raise, for every value that
answers or raises an error, and the decision this ticket has to make is the one
#262 made for the read: whether an object that cannot be
classified reaches `abort_invalid_grouping_spec()`, and whether that moves
ADR 0008's condition-class constraint again — from `simpleError` to
`marginplyr_error`, in the direction ADR 0015 already chose for a malformed
specification reaching that guard.

## Expected outcome

A forged `.grouping` whose kind raises on being classified is refused with
marginplyr's own condition, or the ticket records why the guards keep base R's.
Acceptance:

- the two shapes above raise `marginplyr_error` from the public verbs and from
  `inspect_grouping()`, blamed on the call the caller wrote, or the decision to
  leave them is written down where ADR 0008 says the guards were left alone;
- whatever is decided is asserted beside
  `test-grouping-plan.R`'s "a malformed grouping specification is refused by
  both guards", which covers the read hazard and not this one.

## Origin

Found by the Spec axis of the review of #268's branch, as a counterexample to
that ticket's totality claim for the printer. The printer's half is fixed
there; this is the half that ticket's scope does not reach.

sayuks:
Closes #280.

## What moved

`is.na()` and `length()` are generic, so a forged `.grouping` whose kind carries a method of its own took the guards' answer away from them: the object's error came out of `validate_grouping_spec_early()` in place of `Invalid grouping specification.`. `check_ambiguous_nested_name()` reached the same methods and one more — `%in%` dispatches through `as.character()` — and there an untyped error replaced a compiled call.

`grouping_kind_name()` answers a kind's name or nothing. It catches what classifying raises, exactly as `grouping_spec_kind()` catches what reading raises, and strips the class from the answer so nothing downstream dispatches on it again. The catch #268 put in the printer moves there, since every reader of an unvalidated kind now shares one classification. The answer is then classified a second time, with no class left on it to dispatch, because a catch answers only a method that *fails* to answer — one that answers *wrongly* passes it.

`check_ambiguous_nested_name()` is not named by the ticket and is fixed here anyway: it is the same reader's other consumer, and a fix holding at one site and not the other would read as a fix. Its answer differs — it declines where a guard refuses — and ADR 0008 records that, along with the one behaviour that moves there in the other direction.

## Decision record

ADR 0008's condition-class constraint moves a fifth time, in the direction #262 chose for the read. The amendment states what does not move with it, what the second classification is for, what the nested-name site does instead, and the evaluation counts at each of the three sites.

## Review

A two-axis review (`mattpocock-skills:code-review`) ran over `e2dbff3`, one round, both axes carrying `design/architecture.md`'s *Answering a review* alongside the standards sources as `design/agents/code-review.md` requires. Seven findings, none about how a sentence reads. Answers are in `9195644`; the full record is in the first comment below.

`design/review-dispositions.md` §12 carries the three the diff cannot settle by being read: a test refactor, an assertion that passes either way, and an alternative weighed and rejected.

## Gates

`NOT_CRAN=true testthat::test_local()` green; `lintr::lint_package()` after `pkgload::load_all(".")` reports no lints; `verify-context-budget.R` passes at 38396/38396; `roxygen2::roxygenise()` produces no diff and no generated file is touched.

### DIFF
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -9,16 +9,14 @@ validate_grouping_spec_early <- function(grouping_spec) {
     ))
   }
 
-  kind <- grouping_spec_kind(grouping_spec)
+  # Classified where it is read, so that what this guard holds from here on is
+  # a name and not a value that might still refuse to be one (#280).
+  kind <- grouping_kind_name(grouping_spec_kind(grouping_spec))
   # A catch of its own, because the two fields are not read together: an object
   # can refuse one while answering the other, and this is the only reader of
   # `args` before the guard below has run.
   args <- tryCatch(grouping_spec$args, error = function(cnd) NULL)
-  if (
-    !is.character(kind) ||
-      length(kind) != 1L ||
-      !is.list(args)
-  ) {
+  if (is.null(kind) || !is.list(args)) {
     abort_invalid_grouping_spec()
   }
 
@@ -392,8 +390,10 @@ preflight_grouping_spec <- function(grouping_spec, data_vars) {
 # class with no kind to read -- an atomic vector given the class offers no
 # field to read at all, which `grouping_spec_kind()` answers with `NULL`, and
 # a list without the field answers `NULL` too, neither of which is a kind this
-# position could admit. That reading is the guard's own, so the two cannot
-# disagree about what a kind is readable off. Both catches are narrow
+# position could admit. So is a kind that raises on being classified, which
+# `grouping_kind_name()` answers with `NULL` (#280). Both that reading and that
+# classification are the guard's own, so the two cannot disagree about what a
+# kind is or about what one is readable off. All three catches are narrow
 # in what they decide and not in what they swallow: everything they hide is a
 # failure to produce a specification of an admitted kind, which is a
 # specification reading that is not available. That is where the ADR-0015 line
@@ -429,12 +429,8 @@ check_ambiguous_nested_name <- function(arg, parent, rule, data_vars) {
   if (!inherits(value, "margin_grouping_spec")) {
     return(invisible(NULL))
   }
-  kind <- grouping_spec_kind(value)
-  if (
-    !is.character(kind) ||
-      length(kind) != 1L ||
-      !kind %in% admitted
-  ) {
+  kind <- grouping_kind_name(grouping_spec_kind(value))
+  if (is.null(kind) || !kind %in% admitted) {
     return(invisible(NULL))
   }
   abort_ambiguous_nested_name(name)
@@ -834,20 +830,44 @@ grouping_kind_rules <- local({
   }
 })
 
-# Whether a kind read off an object is a name at all: one string, and not the
-# missing one. The lookup below needs this much before it can be made, and
-# answers `NULL` both for a kind that fails it and for a name it does not know,
-# so a caller that has to tell those apart asks this instead of reading that
-# answer.
-is_grouping_kind_name <- function(kind) {
-  is.character(kind) && length(kind) == 1L && !is.na(kind)
+# The name a kind read off an object is, or `NULL` where it is none: a name is
+# one string, and not the missing one. Every caller holds a value nothing has
+# validated, which is why this answers with the name rather than with whether
+# there is one. Deciding reaches the value's own `is.na()` and `length()`
+# methods, and either can raise instead of answering; a raise is no answer, so
+# the value is no name, exactly as a value answering with two of them is none
+# (#280).
+#
+# The class is stripped from the answer because it is what carries those
+# methods, and the questions asked of a name are not done being asked here:
+# `%in%` reaches `as.character()`, which dispatches too, so a caller that
+# classified a kind and then matched what it classified would raise on the line
+# after this catch. The registry lookup below dispatches on the list and not on
+# the index, so it is inside for the other reason -- one place asks a kind
+# anything.
+#
+# The catch is narrow in what it decides and not in what it swallows, exactly
+# as the read's in `grouping_spec_kind()` is: a value no name can be got out of
+# has none, whatever stopped it. It catches an error and not a condition, so a
+# method that warns on its way to answering still answers.
+grouping_kind_name <- function(kind) {
+  tryCatch(
+    if (is.character(kind) && length(kind) == 1L && !is.na(kind)) {
+      unclass(kind)
+    },
+    error = function(cnd) NULL
+  )
 }
 
+# `NULL` both for a kind that is no name and for a name the registry does not
+# know, so a caller that has to tell those apart asks `grouping_kind_name()`
+# instead of reading this answer.
 find_grouping_kind_rule <- function(kind) {
-  if (!is_grouping_kind_name(kind)) {
+  name <- grouping_kind_name(kind)
+  if (is.null(name)) {
     return(NULL)
   }
-  grouping_kind_rules()[[kind]]
+  grouping_kind_rules()[[name]]
 }
 
 grouping_constructor_names <- function() {
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -832,31 +832,30 @@ grouping_kind_rules <- local({
 
 # The name a kind read off an object is, or `NULL` where it is none: a name is
 # one string, and not the missing one. Every caller holds a value nothing has
-# validated, which is why this answers with the name rather than with whether
-# there is one. Deciding reaches the value's own `is.na()` and `length()`
-# methods, and either can raise instead of answering; a raise is no answer, so
-# the value is no name, exactly as a value answering with two of them is none
-# (#280).
-#
-# The class is stripped from the answer because it is what carries those
-# methods, and the questions asked of a name are not done being asked here:
-# `%in%` reaches `as.character()`, which dispatches too, so a caller that
-# classified a kind and then matched what it classified would raise on the line
-# after this catch. The registry lookup below dispatches on the list and not on
-# the index, so it is inside for the other reason -- one place asks a kind
-# anything.
-#
-# The catch is narrow in what it decides and not in what it swallows, exactly
-# as the read's in `grouping_spec_kind()` is: a value no name can be got out of
-# has none, whatever stopped it. It catches an error and not a condition, so a
-# method that warns on its way to answering still answers.
+# validated, and every reader of one shares this, as they share the read in
+# `grouping_spec_kind()`. What it decides, why it answers with the name rather
+# than with whether there is one, and what the class coming off that answer
+# costs are ADR 0008's, in its amendment for a kind the guard could not
+# classify.
 grouping_kind_name <- function(kind) {
-  tryCatch(
+  # `is.na()` and `length()` are generic, so these are the value's own methods
+  # and one may raise instead of answering. A raise is no answer, so the value
+  # is no name -- the reading that leaves the guards their own refusal (#280).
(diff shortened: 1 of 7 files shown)


sayuks:
## Review record — #280

**Snapshot reviewed:** `e2dbff3` (diff `ac78c59...e2dbff3`). SHAs are post-rebase onto `main`.
**Answers landed in:** `9195644`.
**Rounds:** 1 — `mattpocock-skills:code-review`, Standards and Spec as parallel sub-agents. Both were passed the standards sources and `design/architecture.md`'s *Answering a review* four-way classification of a prose finding, as `design/agents/code-review.md` requires. The skill was not re-invoked; each finding was answered and then verified by running its own evidence.

---

### S1 / P1 — the classification returned a value that was not a name

Both axes found this independently; one defect, two IDs.

| | |
|---|---|
| **Line** | `R/grouping-plan.R:853-859` (`grouping_kind_name()`), consumed at `:433` (`!kind %in% admitted`) and claimed at `:833` and `:12-13` |
| **Contract violated** | The header's own `a name is one string`, and ADR 0008's new amendment where it says the untyped error is removed at the nested-name site |
| **Counterexample** | `length()` and `is.na()` were asked of the **classed** value while the **unclassed** value was returned. A `length()` method that *lies* rather than raises passes the catch: `structure(c("set","sets"), class = "lie")` with `length.lie` → `1L` and `is.na.lie` → `FALSE` returned two strings, and `\|\|` at `:433` raised `'length = 2' in coercion to 'logical(1)'` — a `simpleError` out of a public verb, one line past the catch added to remove exactly that |
| **Scope** | In. The acceptance is `marginplyr_error` from the public verbs, and this is the same untyped escape at a line the commit rewrote on the claim that classification now establishes scalarity |
| **Disposition** | Code + test |
| **Evidence** | The answer is classified a second time, with no class left on it to dispatch, so the return is one string by R's reading rather than a method's. `tests/testthat/test-grouping-plan.R` — *a kind whose classification lies is no more a name for it*. Deleting the second classification makes it fail with the message above: verified by reverting the block and running `NOT_CRAN=true Rscript -e 'pkgload::load_all("."); testthat::test_local(filter="grouping-plan")'` → `Error ('test-grouping-plan.R:511'): 'length = 2' in coercion to 'logical(1)'` |

A note on what the test does **not** prove: the same lying kinds already reached `marginplyr_error` at the guard before the fix, because `[[` answers nothing for an index of two elements and nothing for one of none. Those halves are characterization of an accident the second classification replaces with the guard's own reading, and the test comment says so rather than implying they are regression evidence.

I checked whether stripping the class *before* classifying would remove the catch entirely — `is.character()` and `unclass()` are interceptable by neither S3 nor S4 (`setMethod` refuses both: *must supply a function skeleton*), so it would be total by construction. Rejected: it stops asking the value's own methods at all, which reverses #268's decision for the printed line and breaks its two tests. That is a decision for its own ticket, not a side effect of this one.

### S2 — the header re-derived ADR 0008's argument

| | |
|---|---|
| **Line** | `R/grouping-plan.R:833-852` |
| **Contract violated** | `AGENTS.md` *Code comments* — "Re-deriving the ADR's own argument beside the citation is the second copy: it reads as support for the citation while nothing compares the two, so an amendment leaves it standing and wrong" |
| **Counterexample** | Paragraph 2 restated the ADR's *What the classification answers with* nearly sentence-for-sentence (`%in%` reaches `as.character()`; `[[` dispatches on the list not the index), paragraph 3 its handler paragraph — with no citation, so it was a free-standing second copy |
| **Scope** | In — prose finding, category 2 (duplicates an argument another file owns and can drift from it) |
| **Disposition** | Code (comment rewritten) |
| **Evidence** | The header now cites ADR 0008's amendment for what the ADR owns and stops there; the two inline comments state what only those lines know — which methods are reached, and which classification is R's rather than a method's |

### S3 — the printer comment narrated its own change history

| | |
|---|---|
| **Line** | `R/grouping-spec.R:252-253` — "#268 caught it here alone, and #280 moved the catch to the classifier" |
| **Contract violated** | `AGENTS.md` *Code comments* — "A comment describes the code it sits beside, in the present tense. What a review found, which round answered it ... are things the code does not hold" |
| **Counterexample** | Only the last sentence of the block described present code |
| **Scope** | In — prose finding, category 3 (breaks a rule a repository document states) |
| **Disposition** | Code (comment rewritten to present tense) |
| **Evidence** | The rule quoted above, against the rewritten block |

### S4 — `kind_answering` duplicated across three sites

| | |
|---|---|
| **Line** | `tests/testthat/test-grouping-interface.R:1250-1254`, `test-grouping-plan.R:449-457`, `test-grouping-plan.R:1000-1008` |
| **Contract violated** | None documented — Fowler *Duplicated Code*, a judgement call. Reported as such |
| **Counterexample** | Three near-identical builders of `registerS3method(..., asNamespace("base"))` plus a hand-rolled class name. The hazard is real rather than cosmetic: nothing unregisters a method, so two sites spelling one class name each get whichever was registered last, in whichever order testthat ran their files |
| **Scope** | In — accepted. The repo already keeps shared test fixtures in `helper-*.R` (`helper-namespace-walk.R`, `helper-diagnostic-sites.R`, and six more), so this is its own idiom rather than a new one |
| **Disposition** | Test |
| **Evidence** | `tests/testthat/helper-forged-kinds.R` holds `kind_answering()` and `raising_kind_method()`; all five call sites across the two files take them from there, and the class name is derived from a per-site `purpose` plus the generics. Full suite green |

### P2 — the ADR made a claim a test in the same commit contradicted

| | |
|---|---|
| **Line** | `design/adr/0008-...md:373-375` — "A kind whose `length()` or `as.character()` raises reached the caller as an untyped error and now decides nothing" |
| **Contract violated** | *Answering a review*, category 1 — the prose is false |
| **Counterexample** | An `as.character()`-raising kind now *does* decide: the class comes off before `%in%` is reached, so it classifies as `set` and is refused as ambiguous. `test-grouping-plan.R` asserted exactly that in the same commit |
| **Scope** | In |
| **Disposition** | ADR |
| **Evidence** | Registered `as.character.acr` raising, compiled a colliding `grouping_sets(s)` → `marginplyr_error` with the ambiguity message, and `grouping_kind_name()` answered `"set"`. The amendment now separates three moves at that site instead of two, and says the raising method never bore on the refusal |

### P3 — an acceptance clause was unasserted

| | |
|---|---|
| **Line** | `tests/testthat/test-grouping-plan.R:988-1026` |
| **Contract violated** | The acceptance list — "raise `marginplyr_error` ... **blamed on the call the caller wrote**" |
| **Counterexample** | None; the blame was already correct. Coverage gap, not a defect — and the neighbouring test it was told to sit beside omits it too |
| **Scope** | In |
| **Disposition** | Test |
| **Evidence** | `rlang::call_name(conditionCall(error))` is now asserted for both public routes; measured `conditionCall(e)` → `inspect_grouping(d, .grouping = spec)` before the assertion was written |

### P4 — the ADR's evaluation-count claim (reported at low confidence; confirmed)

| | |
|---|---|
| **Line** | `design/adr/0008-...md:381-383` — "How often classifying asks the value's own methods falls rather than rises" |
| **Contract violated** | *Answering a review*, category 1 — false at one of the three sites |
| **Counterexample** | It falls at the guard and at the printed line, but is flat at `check_ambiguous_nested_name()`: `length`+`as.character` (2) became `length`+`is.na` (2), with `is.na` rising 0→1 |
| **Scope** | In |
| **Disposition** | ADR |
| **Evidence** | Measured with counting methods registered for all three generics. Guard alone: `is.na=1 length=1` after, against `length=2 is.na=1` before (`find_grouping_kind_rule()` re-asked). Printer: `1`/`1` after, `2`/`2` before. Nested collide: `is.na=1 length=1 as.character=0` after, `length=1 as.character=1` before. The amendment now states both directions and names the site each holds at |

### R1 — found while answering, not by either axis

| | |
|---|---|
| **Line** | The comment on the new *lies* test |
| **Contract violated** | *Answering a review*, category 1 — the prose is false |
| **Counterexample** | It claimed both forged shapes were written because "`\|\|` refuses" them differently, but only one shape reaches `\|\|`; the other is asserted at the guard, which refuses either way. Found by running the revert-and-fail check the finding's own evidence called for |
| **Scope** | In |
| **Disposition** | Test comment |
| **Evidence** | Reverting the second classification and running the suite fails only the colliding assertion; both guard assertions pass without it, measured directly |

---

### Gates

`NOT_CRAN=true testthat::test_local()` green across the whole suite; `lintr::lint_package()` after `pkgload::load_all(".")` reports no lints; `Rscript .github/scripts/verify-context-budget.R` passes at 38396/38396; `roxygen2::roxygenise()` produces no diff, and no generated file is touched.

Neither axis found a violation of the one-backend-per-test rule, the lazy-query policy, or the context budget. The Spec axis additionally swept ~25 generics (`format`, `print`, `[`, `[[`, `c`, `Ops`, `xtfrm`, `match`, `names`, ...) as raising methods across the top-level, nested, colliding-parent, colliding-nested and printer routes and found no further untyped escape.

### Ledger

`design/review-dispositions.md` §12 (`4f39b83`) carries the three findings whose disposition a reader of the diff cannot check by reading it — S4, P3, and the rejected alternative. The other four are exempt under the file's two legs: S1/P1 under the first (the commit names it and lands a test that fails without it), S2/S3/P2/P4/R1 under the second (prose-only hunks).

Issue #288, opened 2026-08-28
Retire the review disposition ledger, or say what it buys that a pull request does not


sayuks:
## Problem

`design/review-dispositions.md` is the largest file in the repository at
75,869 bytes — 1.8× `design/architecture.md`, 2× `AGENTS.md`, 5× `CONTEXT.md`
— and it has grown at every single commit that touched it, from 25,183 bytes
at `ef4fce9` to its present size across 23 commits. It has never once shrunk.

It is also the only document here that nothing checks. No workflow reads it, no
`verify-*.R` script derives anything from it, and no test fails when it is
stale, wrong, or simply not written. That is not a hypothetical: the review of
#280's branch owed it three entries and was completed without them, with the
full suite, `lintr::lint_package()`, `verify-context-budget.R`, and
`document.yaml` all green. The omission was caught by hand, afterwards, and
only because someone went looking.

## Evidence

Measured at `4f39b83`:

| | |
| --- | --- |
| Size | 75,869 bytes |
| Growth | 25,183 → 75,869 across 23 commits, monotonic |
| Cited by | `design/architecture.md`, two lines in *Repository placement* |
| Cited by `R/`, `tests/`, `.github/`, `AGENTS.md`, any ADR | nothing |
| Enforced by | nothing |
| §1–§9, the 0.1.0 stabilization history | 47,437 bytes — 63% |
| Scope rules, before §1 | 5,286 bytes |
| Entries | 40 marked fixed, 13 marked rejected |

The growth has a shape worth naming. The two commits that *narrowed* what
reaches the file — `0b190b1`, resting the first exemption on a test that fails
without the fix, and `1d99523`, adding the prose exemption — added 786 and
7,669 bytes respectively. Each was written to stop entries being made, and
each cost more than the entries it prevents. The rules are now 5,286 bytes
deciding the fate of entries that run 100 to 400 bytes each.

That is the argument against the file in its own terms: a document whose
scope rule is more expensive than its contents is one whose contents were not
what it was for.

## What the file buys, and whether anything else buys it

Its stated job is in `design/architecture.md`:

> `review-dispositions.md` is the ledger that gives "review complete" an
> auditable meaning ... An ADR records what was decided; the ledger records
> what was reviewed and what happened to it.

Most of that is now bought elsewhere, by practices that did not exist as
practices when `ef4fce9` created it:

- a review record on the pull request, naming per finding the snapshot, the
  line, the contract, the counterexample, the scope, the disposition, and the
[...]

PR #297, opened 2026-08-28
Hand the reviewer its contract instead of pointing at it

sayuks:
`AGENTS.md` names `design/agents/code-review.md` in one line. A pointer is read
only if the reader goes looking, and what that file holds cannot be picked up
afterwards — each of its three constraints is spent before anything could fail,
so there is nothing to gate after the fact.

| the constraint | what is lost if the invocation does not carry it |
|---|---|
| pass what a finding about prose may be, alongside the standards sources | the skill composes both sub-agent prompts itself (`SKILL.md`, step 4), so the scope reaches neither axis |
| push the branch and open the PR before writing the record | the record names SHAs nobody can open — #286 had to move #280's for this |
| the fixed point is the merge-base | `main`'s gap is reported as the branch's |

## What lands

A `PreToolUse` hook on the `Skill` tool, in `.claude/settings.json`, running
`.claude/hooks/code-review-contract.sh`. The harness runs it, so it does not
depend on the model deciding to read a file first.

**It injects the file, not a copy of what the file says.** There is nothing here
to drift from the contract, and editing the contract is editing that file —
which the file now says of itself, since a reader deciding whether to change it
is deciding what every later review is handed.

It matches the built-in `code-review` skill as well as the plugin's: what the
contract governs is a review of this repository, not one reviewer's
implementation of it.

## Verification

Piped the payload the hook receives, against the exact command string in
`settings.json`:

| case | result |
|---|---|
| `mattpocock-skills:code-review` | injects the file |
| built-in `code-review` | injects the file |
| `tdd` | injects nothing |
| invoked from `R/` | resolves the same file as from the root |
| invoked outside a repository | exits 0, silent |

`jq -e '.hooks.PreToolUse[] \| select(.matcher == "Skill") \| .hooks[] \| select(.type == "command") \| .command'` exits 0.
`verify-context-budget.R` is unchanged at 38,383 of 38,396 — `.claude/` and
`design/agents/` are outside the always-loaded closure. `test-documentation.R`
passes.

**Not proved firing live.** `.claude/settings.json` did not exist when this
session started, so the settings watcher is not watching that directory; the
hook is written and pipe-tested but takes effect after `/hooks` or a restart.

## Also

`.claude/` was untracked, so `.claude/worktrees/` is ignored now that the
settings and hook beside it are not. `.Rbuildignore` already excluded the
directory, so nothing reaches the tarball.

## What does not land

Anything in `AGENTS.md`. The pointer there is the fallback for a reader whose
settings do not load, and stating the hook beside it would be a second copy of
what the hook's own header argues — at a cost inside the budget that gate holds.

No ledger entry: the ledger was retired in #288, and what this produced is the
hook and its pipe-tested behaviour.

### DIFF
--- a/.claude/hooks/code-review-contract.sh
+++ b/.claude/hooks/code-review-contract.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+# Hands `design/agents/code-review.md` to the session that is about to run a
+# review, as a `PreToolUse` hook on the `Skill` tool.
+#
+# `AGENTS.md` points at that file, and a pointer is read only if the reader goes
+# looking. What it holds is not advice: the skill composes its own sub-agent
+# prompts, so a scope the invocation does not carry reaches neither axis; a
+# record written before the branch is pushed names SHAs nobody can open (#286);
+# and a fixed point that is not the merge-base reports `main`'s gap as the
+# branch's. Each of those is spent before anything fails, so there is nothing to
+# gate afterwards. This is what makes the file arrive rather than be looked up.
+#
+# It injects the file rather than a copy of what the file says, so there is
+# nothing here to drift from it. Editing the contract is editing that file.
+#
+# Silence is the only failure mode: outside a repository, on any other skill, or
+# with the file absent, it exits 0 and adds nothing.
+
+set -u
+
+input=$(cat)
+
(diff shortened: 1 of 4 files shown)


sayuks:
Withdrawn. A hook is harness configuration and should not have been added without asking; the same three constraints are being solved in documentation instead.

Issue #304, opened 2026-08-29
count_backend_reads() is named for reads and counts entry-point invocations


sayuks:
> *This was generated by AI during the code review of #301.*

Raised on the Standards axis of #301's review and not answered there: the
finding is about a helper that branch did not introduce, and no assertion it
adds depends on the answer.

`count_backend_reads()` is named for reads and counts traced invocations on the
path of a read. One read reaches several entries, so the number it returns is
not a count of reads:

| expression | counted | reads |
|---|---|---|
| `dplyr::collect(<duckdb tbl>)` | 3 | 1 |
| `dplyr::collect(<lazy_dt>)` | 2 | 1 |
| `summarize_with_margins(<lazy_dt>, ...)` | 2 | 1, exemption 1 of ADR 0020 |

The duckdb 3 predates #300. #301 added the `tibble::as_tibble` entry, which is
what makes the dtplyr rows 2 rather than 1 — the entry is warranted on its own
path, and this is a consequence of it rather than an objection to it.

Nothing is wrong today. Every assertion built on the counter is a zero or a
greater-than-zero, and both readings survive a number that overcounts. What the
name costs is the next assertion: an equality against a positive number reads as
"this many reads happened" and would be measuring how many catalog entries
happen to sit on one path, which changes whenever the catalog does.

## What would answer it

Either the name says what it counts — `count_traced_entry_points()`,
`count_entry_point_invocations()` — or the helper returns reads and the counter
deduplicates per read, which needs a definition of "per read" that nothing here
has. The first is a rename in one test file with no behaviour change; the second
is the "rule about not counting a read already counted further out" that #301
weighed and rejected as more machinery than the gap warranted.

The rename is the one worth doing. It is also worth doing *with* something that
makes the property hard to forget: an assertion that the counter reports more
than one for a single read on at least one path, so a future change that made it
report exactly one would fail rather than pass quietly.

Related: #303 holds the two ways the counter reports *fewer* invocations than
the path carries.

Issue #303, opened 2026-08-29
The caused-read gate misses a bare entry-point call, and one nested in a subject test's argument


sayuks:
> *This was generated by AI during triage.*

Split out of #301's triage. Both findings are properties of the tracing
mechanism rather than of the catalog, so they are held apart from that issue,
which closes one catalog gap.

Neither lets an uncounted read through today. `R/` spells every entry point
qualified — `dplyr::collect()`, `as.data.frame()` — and nests none inside
another's argument, so every read marginplyr causes is still counted. What is
exposed is the gate's own assertions: an expectation written either way
measures less than it appears to, which is the shape `AGENTS.md` rules out for
a derived gate — a clean result that reads exactly like a package that causes
no read.

## 1. A bare entry-point call is not traced

`trace(where = asNamespace(pkg))` rewrites the namespace binding, and it also
updates the copy in an importing package's imports environment — measured, so a
read another package performs on marginplyr's behalf is still seen. What it
does not touch is the **attached** `package:*` binding, which is what a call
written bare at the top level of a test resolves to.

Measured against the merged gate, one session, arrow 25.0.1:

| expression | counted | rows read |
|---|---|---|
| `dplyr::collect(<Table>)` | 1 | 3 |
| `collect(<Table>)` after `library(dplyr)` | **0** | 3 |

```r
b <- get("as_tibble", envir = as.environment("package:tibble"))
trace("as_tibble", where = asNamespace("tibble"), tracer = function() NULL)
identical(b, get("as_tibble", envir = as.environment("package:tibble")))  # TRUE
inherits(get("as_tibble", envir = asNamespace("tibble")), "functionWithTrace")  # TRUE
```

`base::as.data.frame` is unaffected, because a bare call resolves to the base
namespace this traces. Every other entry in the catalog is a `dplyr::` or
`DBI::` one, and each is exposed. The existing positive control passes because
it is written `dplyr::collect(table)`; the same control written bare would
assert nothing while reading exactly as it does now.

## 2. An entry point inside a subject test's argument is suppressed

`base::.doTrace` calls `tracingState(FALSE)` for the duration of a tracer and
restores it on exit. A subject-tested tracer reads its subject, which forces the
caller's promise, so **any entry point invoked while that promise is forced goes
uncounted**.

```r
f <- function(x) x
g <- function(y) y + 1
trace("f", tracer = function() force(get("x", envir = parent.frame(), inherits = FALSE)))
[...]

Issue #254, opened 2026-08-25
A Margin summary over an Arrow table fails with an untyped error when Arrow cannot translate the expression


sayuks:
## Problem

A Margin summary over an Arrow table fails with an untyped internal error whenever the caller's summary expression is one Arrow cannot translate into its own compute kernels.

Arrow's documented behaviour for such an expression is to warn and evaluate it in R instead ("Expression not supported in Arrow → Pulling data into R"). `dplyr::summarize()` on the same table takes that route and returns the correct answer. A Margin verb on the same table, with the same expression, aborts with:

```
Error: object of type 'special' is not subsettable
```

The condition carries the classes `notSubsettableError/error/condition`. It names nothing the caller wrote, describes no defect they can act on, and offers no rewrite.

The affected class of expressions is not exotic. Every one of these is accepted by `dplyr::summarize()` on an Arrow table and rejected this way by a Margin verb:

| Summary expression | `dplyr::summarize()` | Margin verb |
|---|---|---|
| `sum(v)` | ok | ok |
| `stats::weighted.mean(v, v)` | ok | **error** |
| `paste(s, collapse = ",")` | ok | **error** |
| `sum(v[v > 1])` | ok | **error** |
| `rev(v)[1]` | ok | **error** |
| `sum(.data[["v"]])` | ok | **error** |

This is scoped to in-memory Arrow inputs — an Arrow `Table` or `RecordBatch`, where Arrow's R fallback applies. An Arrow `Dataset` has no fallback, so Arrow raises its own `arrow_not_supported` condition there and a Margin verb propagates it correctly, as an External condition should be.

The proximate defect is upstream, in how Arrow recovers the originating call before falling back: it reads that call out of its own frame with an inheriting lookup, so when the call is absent it silently finds `base::call` — an object of type `"special"` — and subsetting it raises. What decides whether the call is absent is the *shape* of the frame the dplyr verb was reached from. A Margin verb reaches `dplyr::summarize()` once per grouping set through a wrapper that forwards the caller's expressions in `...`, and that shape is the one Arrow cannot recover the call from. Calling the same verb directly, or through a wrapper that does not forward `...`, recovers it and falls back correctly.

So the failure is only reachable through this package, on a backend this package documents as supported.

## Expected behavior

Three repository contracts bear on this, and no disposition may leave all three as they are.

[...]

PR #305, opened 2026-08-29
Watch tibble::as_tibble in the caused-read gate

### LINKED ISSUE #301: The caused-read gate does not watch tibble::as_tibble(), so a dtplyr materialisation goes uncounted
Split out of #300's verification. The subject test added there is sound; this is
a materialisation it cannot see, found by asking which other verbs bring a lazy
input local.

## What #300 covers

`as.data.frame` is now traced with a subject test, and on an Arrow input that
turns out to be the chokepoint for a whole family of conversions. Each of these
reaches `base::as.data.frame` **with the Arrow object as the subject**, so each
is counted, and none of them was counted before:

| call on an Arrow input | entries reached | counted |
|---|---|---|
| `tibble::as_tibble()` | `base::as.data.frame(Table)` | yes |
| `tibble::as.tibble()` | `base::as.data.frame(Table)` | yes |
| `tibble::as_data_frame()` | `base::as.data.frame(Table)` | yes |
| `as.list()` | `base::as.data.frame(Table)` | yes |
| `data.frame()` | `base::as.data.frame(Table)` | yes |

On SQL the same calls route through `dplyr::collect` and the `DBI::` entries,
which were already watched.

## The gap

On a dtplyr input the conversion happens the other way round. dtplyr
materialises first and converts afterwards, so the traced call never receives
the lazy object:

```
as.data.frame(dt)   subjects seen: dtplyr_step_first, data.table   counted 1
as_tibble(dt)       subjects seen: data.table                      counted 0
```

The read in the second row happened — the result is three rows — and nothing
counted it. `as.tibble()` and `as_data_frame()` behave identically, since both
delegate to `as_tibble()`.

This is a hole in coverage rather than an unreported read, exactly as #257 was.
Nothing reaches it today: `R/` never calls `as_tibble()` on data, its only
appearance being a roxygen example in `summarize_with_margins.R`, and dtplyr is
not an Absorbing backend, so no third package calls it on marginplyr's behalf.

## What would close it

One catalog entry, `tibble::as_tibble` with a subject test. Measured on tibble
3.3.1: tracing that generic fires with `dtplyr_step_first` for `as_tibble(dt)`,
and both deprecated aliases delegate to it, so a single entry covers three
spellings. There is no `as_tibble` method registered for `dtplyr_step`, `Table`,
`arrow_dplyr_query` or `tbl_lazy`, so the generic is where it must be traced —
the same reason #254 gives for tracing `dplyr::collect` rather than Arrow's
method.

What needs deciding is whether the double count that follows is acceptable. On
an Arrow input `as_tibble()` would then be counted twice, once at `as_tibble`
and once at `as.data.frame` beneath it. Every assertion the gate makes is a zero
or a "greater than zero", so nothing breaks — but a counter that reports two for
one read is a counter whose number means less than it appears to, and the
alternative is a rule about not counting a read already counted further out,
which is more machinery than the gap warrants.

## What is deliberately not proposed

`tibble::tibble()` needs no entry. It is a constructor rather than a conversion,
and it refuses a lazy input outright — `tibble(x = <Table>)` raises `All columns
in a tibble must be vectors` — so it never receives one to read.

`Table$to_data_frame()` cannot be covered by this mechanism at all. It is an R6
method rather than a generic, so nothing dispatches to it and `trace()` on a
generic cannot see it. Nothing in `R/` can reach it either, since marginplyr
holds no Arrow object it calls methods on directly.

`as.matrix()` is not a path: it fails on an Arrow input rather than
materialising it.

sayuks:
Closes #301.

#300 gave the catalog a subject test and put `as.data.frame` behind one, which
counts every conversion reaching it while the caller's lazy object is still the
subject. One conversion path reaches it only after that has stopped being true:
`as_tibble()` on a dtplyr step evaluates the step and converts the resulting
`data.table`, so the entry sees a local table and the read is counted zero while
three rows come back. The generic that does see the step is
`tibble::as_tibble`, which the catalog did not name. One row added,
subject-tested; two test files, no change under `R/`.

## What the entry covers, and what it does not

The conversion, not the backend. Measured on both sides of this branch:

| expression | main | here | rows |
|---|---|---|---|
| `tibble::as_tibble(<lazy_dt>)` | 0 | 1 | 3 |
| `dplyr::as_tibble(<lazy_dt>)` | 0 | 1 | 3 |
| `as.tibble(<lazy_dt>)` | 0 | 1 | 3 |
| `as_data_frame(<lazy_dt>)` | 0 | 1 | 3 |
| `as.data.frame(<lazy_dt>)` | 1 | 1 | 3 |

The last row is why the header says conversion rather than backend: dtplyr
registers an `as.data.frame` method, so the base generic is reached with the
step and that spelling was never the gap. The subjects the generic actually
sees are `dtplyr_step_first, data.table` on that path and `data.table` alone on
the `as_tibble` one.

The three remaining spellings close from the one entry because the two
deprecated ones delegate to the generic from inside tibble's own namespace, so
the caller's spelling does not decide whether the traced binding is reached.

## Why the generic, and why a subject test

Traced at the generic for the reason #254 gives for `dplyr::collect`, not for
the reason #301's body gives. That body says no `as_tibble` method is registered
for `dtplyr_step`; `as_tibble.dtplyr_step` is registered, and tracing a method
would see only a by-name call. The generic runs before dispatch, which is where
the step is still the subject.

The subject test is what the entry needs rather than an option it takes.
Converting a dtplyr step invokes the generic twice — once on the step, once on
the local table it produced — so an unfiltered entry would report two for one
read, and would count every `as_tibble()` any package makes on a local frame.

## What holds it

Both directions, as the `as.data.frame` controls do: a subject test answering
`TRUE` for everything and one answering `FALSE` for everything leave every zero
in the file reading exactly as it does now. The negative control is the local
frame.

Every expectation names its package. `trace()` rewrites a namespace binding and
the copy an importing package's imports environment holds, but not the attached
`package:tibble` one, so a bare `as_tibble()` written in a test counts nothing
while reading exactly as these do. That is #303, along with the
nested-entry-point hole it shares a cause with; neither is answered here.

The dtplyr assertions sit in a `test_that()` of their own behind
`skip_if_suggest_absent("dtplyr")`, so no test requires two members of
`optional_backends()`. `verify-suite-coverage.R` passes with the new test
executing in exactly one of the five single-backend configurations.

## Not changed

The Arrow test, in every number it asserts — both positive controls, the schema
and local-frame negatives, the two verb zeroes, and the four refusal shapes —
measured with the entry present. The reach snapshot, since `R/` writes no
`as_tibble()` call on data; its three mentions are all roxygen.

ADR 0020 is untouched. Membership is what the snapshot records, and the
decision, its two exemptions, and its test strategy are unchanged.

`summarize_with_margins()` on dtplyr goes from 1 to 2 — exemption 1 of ADR 0020
counted at two entries on one path. Nothing asserts it, as nothing asserts the 3
`collect()` already reports on duckdb. #304 holds what that number's name costs.

### DIFF
--- a/tests/testthat/_snaps/query-policy.md
+++ b/tests/testthat/_snaps/query-policy.md
@@ -3,16 +3,17 @@
     Code
       lazy_execution_entry_points()
     Output
-        package            name subject_test
-      1   dplyr         collect        FALSE
-      2   dplyr         compute        FALSE
-      3   dplyr            pull        FALSE
-      4    base   as.data.frame         TRUE
-      5     DBI      dbGetQuery        FALSE
-      6     DBI     dbSendQuery        FALSE
-      7     DBI dbSendStatement        FALSE
-      8     DBI         dbFetch        FALSE
-      9     DBI     dbReadTable        FALSE
+         package            name subject_test
+      1    dplyr         collect        FALSE
+      2    dplyr         compute        FALSE
+      3    dplyr            pull        FALSE
+      4     base   as.data.frame         TRUE
+      5   tibble       as_tibble         TRUE
+      6      DBI      dbGetQuery        FALSE
+      7      DBI     dbSendQuery        FALSE
+      8      DBI dbSendStatement        FALSE
+      9      DBI         dbFetch        FALSE
+      10     DBI     dbReadTable        FALSE
 
 # marginplyr functions reaching an execution entry point
 
--- a/tests/testthat/test-query-policy.R
+++ b/tests/testthat/test-query-policy.R
@@ -70,26 +70,38 @@ find_local_assignment <- function(fn, var_name) {
 # ordinary query-building call. `dplyr::show_query()` is deliberately absent
 # -- ADR 0020 states that it runs nothing.
 #
+# `tibble::as_tibble` is a member because it is the generic that receives the
+# caller's lazy object on a conversion path where `as.data.frame` receives only
+# a local one: converting a dtplyr step evaluates the step first and converts
+# the resulting `data.table`, so the read goes uncounted (#301).
+# `as.data.frame()` written directly against that same input is counted, dtplyr
+# registering a method the base generic dispatches from, so what this entry
+# covers is the conversion and not the backend. It is traced at the generic
+# rather than at `as_tibble.dtplyr_step`, for the reason given below for
+# `dplyr::collect`.
+#
 # `subject_test` says whether counting an invocation requires knowing what the
 # call was applied to, and it is enumerated for the same reason the membership
 # is. `as.data.frame` needs one because its name is also how unrelated code
 # converts unrelated objects -- a verb reads its input's schema through it
-# repeatedly, and no such conversion reaches a backend. No `DBI::` entry takes
-# one: invoking it is itself a read, and its first argument is a connection
-# rather than the caller's data, so a subject test there would count nothing at
-# all.
+# repeatedly, and no such conversion reaches a backend. `as_tibble` needs one
+# for the same reason, and reaches it on the path this entry exists for:
+# converting a dtplyr step invokes the generic twice, once on the step and once
+# on the local table it produced. No `DBI::` entry takes one: invoking it is
+# itself a read, and its first argument is a connection rather than the caller's
+# data, so a subject test there would count nothing at all.
 lazy_execution_entry_points <- function() {
   data.frame(
     package = c(
-      "dplyr", "dplyr", "dplyr", "base",
+      "dplyr", "dplyr", "dplyr", "base", "tibble",
       "DBI", "DBI", "DBI", "DBI", "DBI"
     ),
     name = c(
-      "collect", "compute", "pull", "as.data.frame",
+      "collect", "compute", "pull", "as.data.frame", "as_tibble",
       "dbGetQuery", "dbSendQuery", "dbSendStatement", "dbFetch", "dbReadTable"
     ),
     subject_test = c(
-      FALSE, FALSE, FALSE, TRUE,
+      FALSE, FALSE, FALSE, TRUE, TRUE,
       FALSE, FALSE, FALSE, FALSE, FALSE
     ),
     stringsAsFactors = FALSE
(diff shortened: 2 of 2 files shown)


sayuks:
> *This review was run by AI.*

## Review round 1

(`git diff d3ae9c4...HEAD`), both axes, on `2a8fb85` — amended to `5b636b6` by
what the round produced.

Standards carried the `AGENTS.md` sections the diff reaches — *Linting*,
*Queries against a lazy input*, *Code comments*, *Release matrix*,
*Optional-dependency guards*, *Dependency metadata* — plus ADR 0020 and
*Answering a review*. Spec carried #301 and its agent brief.

### Standards

**The header's justification was false. Fixed in code.** Both copies said that
on a dtplyr input the traced `as.data.frame` is reached with the resulting
`data.table` and the read goes uncounted. Measured, `as.data.frame(<lazy_dt>)`
counts 1 on `main` and 1 here: dtplyr registers a method, so the base generic is
reached with `dtplyr_step_first` first. The generic sees `dtplyr_step_first,
data.table` on that path and `data.table` alone on the `as_tibble` one, so the
gap is the conversion and not the backend. The claim was deleted rather than
restated, per *Code comments*; what replaced it is the distinction the
measurement draws, and the commit message carried the same sentence and was
amended.

**Duplicated Code, judgement call. Answered by the same edit.** The entry's
rationale stood in full at the catalog and again at the test, and the false
sentence was in both copies with nothing comparing them — the cost the rule
names, realized. The argument now sits once, at the catalog header that owns
membership; the test header keeps only what that site knows.

**Mysterious Name, judgement call. Ticketed as #304.** `count_backend_reads()`
counts traced invocations on a read's path, not reads. This branch takes the
dtplyr `collect()` path from 1 to 2; duckdb's 3 predates #300. Nothing breaks —
every assertion is a zero or a greater-than-zero — and the branch's acceptance
does not reach the name, so it became a ticket rather than prose here.

Verified clean by that axis: the #303 attached-copy claim, the delegation of
both deprecated spellings, the correction to #301's body, and the absence of a
guard on `tibble::`/`dplyr::`, which `optional_suggests()` would refuse.

### Spec

No blocking findings. Every acceptance criterion was checked by measurement
rather than by reading, including criterion 5 — `main`'s Arrow block extracted
verbatim and evaluated against this branch's catalog: 10 successes, 0 failures,
no number moved.

One observation, deliberately left alone: the local-frame negative control sits
inside the dtplyr-guarded block, so it skips in every `backend` job but that
one. It is the control for the assertions it sits with, and the Arrow block
places its own controls the same way; diverging would need more explaining than
it buys.

### Gates

Full suite green. `lintr::lint_package()` clean after `load_all()`.
`verify-suite-coverage.R` exit 0 — 628 tests, this one executing in exactly one
of five single-backend configurations. Snapshot diff is the one new row; the
reach snapshot did not move.

Issue #257, opened 2026-08-25
The caused-read gate does not watch as.data.frame()


sayuks:
Split out of #254's review. The gate added there is sound for what it watches;
this is the bound it states rather than a defect in it.

## What the gate watches

`test-query-policy.R` gained a reading that ADR 0020's other readings cannot
make: they walk `R/`, so they are blind to a read another package performs on
marginplyr's behalf. The new one counts reads as they happen, tracing the entry
points in `lazy_execution_entry_points()`.

It traces only the *qualified* entries — `dplyr::collect`, `dplyr::compute`,
`dplyr::pull`, and the `DBI::` ones. The catalog's one bare entry,
`as.data.frame`, is deliberately excluded, and the reason is in the comment
there: it is a base generic that unrelated code calls while any expression
runs, so counting it would report a read that never reached a backend. A gate
with false positives is worse than the narrower one.

## The gap

A caused read that goes through `as.data.frame()` on a backend object is
therefore not counted. Nothing reaches it today — Arrow's fallback uses
`dplyr::collect()` — so this is a hole in coverage rather than an unreported
read.

## What would close it

Trace `as.data.frame` and filter in the tracer by whether the object being
converted is a backend object rather than an ordinary one, so the unrelated
calls are not counted. `arrow_input_classes()` in `R/grouping-backend.R` and
`is.data.frame()` are most of that predicate already; what needs deciding is
how the tracer reaches the argument, and whether the same filter should cover
the `DBI::` entries, which are counted without one today because no Arrow input
can reach them.

Worth doing when a second Absorbing backend arrives, or sooner if a path
through `as.data.frame()` appears. Not urgent: the gate's positive control
asserts it counts a real read, so it cannot silently stop working for the
entries it does watch.

sayuks:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Trace `as.data.frame` in the caused-read gate, counted only when the object it converts is a lazy backend input.

**Current behavior:**

The runtime caused-read counter derives its traced set from the ADR 0020 execution catalog by keeping only the qualified `pkg::name` entries, so the catalog's one bare entry, `as.data.frame`, is never traced. Nothing in the package reaches an `as.data.frame` read today — the Absorbing-backend fallback uses `collect()` — so this is a hole in coverage rather than an unreported read.

[...]

PR #307, opened 2026-08-29
State each review-record argument once, and record a review that did not run

### LINKED ISSUE #306: A review record's routing is stated twice, and nothing distinguishes a skipped review from a clean one
Two defects in the rule #287 introduced, both measured against what has merged since.

## 1. The routing is stated in two files, and nothing compares them

`design/agents/code-review.md` routes a disposition by kind in a six-row table. `design/architecture.md`'s *Repository placement* restates five of those six rows in prose — the commit, the test, and the diff for what a review produced, and *Considered options*, an `investigation/` note, and a comment at the site for what it decided without producing anything — and then points at the table anyway.

#290 put it there deliberately: its body says *Repository placement* "now states the same thing in one paragraph instead of pointing at a file". The result is that it does both. A row added to or removed from the table leaves the paragraph stale, and nothing reads either — the second copy this repository refuses for a comment re-deriving an ADR, at a document.

What *Repository placement* owns is the placement claim: a review gets no file under `design/`. The routing is the other file's.

It has already drifted, which is how a second copy fails rather than a prediction of how it might: the paragraph says the ledger held the dispositions "a third time", while `design/agents/code-review.md` and this tracker's #288 both say a second.

## 2. A skipped review and a review that found nothing are the same absence

Nothing says a branch owes a record, only where one goes once written. So no record carries two readings, and both merged the same night:

| | round run | record |
|---|---|---|
| #300 — two test files, +149 / -39 | none | none |
| #305 — same shape | one, three findings | posted |

Nothing in #300 says which it was. It opened 81 minutes after #299 merged, moving the review instructions into the always-loaded closure, so this is not a branch that could not reach the rule — the rule has no clause it could have reached.

This is the objection this repository makes to every gate it has built: an option that asserts nothing reports nothing, which reads exactly like the thing it was supposed to catch.

## What would close this

The routing stated once, and a clause making a branch that ran no review say so. The clause belongs in the always-loaded closure — an agent that never invokes the skill never opens `design/agents/code-review.md`, which is exactly the case it governs — so it has to be paid for.

`AGENTS.md`'s *Code review* holds a second copy of the sub-agent-prompt argument, and that is where the payment should come from.

sayuks:
Closes #306. Three files, all `.Rbuildignore`d — `lint.yaml`'s context-budget gate is the only CI path any of them reaches.

## The routing, stated once

`design/architecture.md`'s *Repository placement* restated five of the six rows of `design/agents/code-review.md`'s disposition table and then pointed at the table. #290 put it there on purpose — "now states the same thing in one paragraph instead of pointing at a file" — and the result does both, so a row added to either goes stale with nothing reading them together. It already had: the paragraph said the ledger held its dispositions "a third time" where the table's file and #288 both say a second.

That section keeps the claim it owns, which is a placement claim and not a routing one: a review gets no file under `design/`. Every route the paragraph named is in the table already, checked row by row.

## The clause the rule lacked

#287 said where a record goes and what state the branch must be in. It never said a branch owes one, so no record has two readings:

| | round run | record |
|---|---|---|
| #300 — two test files, +149 / -39 | none | none |
| #305 — same shape | one, three findings | posted |

#300 opened 81 minutes after #299 merged, moving the review instructions into the always-loaded closure, so this is not a branch that failed to reach the rule — the rule had no clause it could have reached. A branch that ran no review now records that, which is one sentence and makes a skip a decision.

It goes in the closure rather than behind the pointer for the same reason #299's three do: an agent that never invokes the skill never opens `design/agents/code-review.md`, and that agent is precisely the case this governs.

## What paid for it

A deletion in the same section, and the round on this branch is what found it. `AGENTS.md` said the skill "composes its own sub-agent prompts, so what an invocation does not carry does not reach either axis"; `design/agents/code-review.md` said the same as "nothing else reaches them". Both are false — the always-loaded closure reaches a sub-agent whether or not the invocation names it — so both are deleted rather than restated, per *Answering a review*. The instruction to pass the standards sources stands without either, and `SKILL.md` step 4 still says what composes the prompts.

The baseline does not move: **38,753 of 38,810**, verified by `.github/scripts/verify-context-budget.R`.

## Not done here

The #300 case is now recorded rather than gated. Nothing fails when a pull request carries no record, consistent with how this repository treats the comment policy and the dependency audit — and unlike the closure, a record's presence has no derivation that could stand in for reading it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -626,13 +626,12 @@ This is a single-context repo. Read the root `CONTEXT.md` and relevant ADRs unde
 
 ### Code review
 
-Reviews run through `mattpocock-skills:code-review`, which composes its own
-sub-agent prompts, so what an invocation does not carry does not reach either
-axis. Pass what a finding about prose may be — *Answering a review* in
-`design/architecture.md` — alongside the standards sources. Review against the
-merge-base with `main`, and push the branch and open the pull request before
-writing the record, which goes there. `design/agents/code-review.md` maps the
-rest of what the skill expects onto this repository.
+Reviews run through `mattpocock-skills:code-review`. Pass what a finding about
+prose may be — *Answering a review* in `design/architecture.md` — alongside the
+standards sources. Review against the merge-base with `main`, and push the
+branch and open the pull request before writing the record, which goes there. A
+branch that ran no review records that instead. `design/agents/code-review.md`
+maps the rest of what the skill expects onto this repository.
 
 ### Investigation notes
 
--- a/design/agents/code-review.md
+++ b/design/agents/code-review.md
@@ -2,11 +2,13 @@
 
 Reviews here run through that skill. It reviews the diff since a fixed point on two axes — Standards and Spec — as parallel sub-agents. This file maps what it expects onto what this repository has.
 
-What may be a finding, and how one is answered, is not here. `design/architecture.md`'s *Answering a review* governs that, for this reviewer and for any other. `AGENTS.md` carries what the invocation has to hand this skill, because `SKILL.md` step 4 composes both sub-agent prompts from the standards sources the call names and nothing else reaches them.
+What may be a finding, and how one is answered, is not here. `design/architecture.md`'s *Answering a review* governs that, for this reviewer and for any other. `AGENTS.md` carries what the invocation has to hand this skill, because `SKILL.md` step 4 composes both sub-agent prompts from the standards sources the call names.
 
 ## What is reviewed, and where the record goes
 
 The subject is a diff, so the record goes where the diff is: the pull request. An issue gets a pointer and not a copy — the issue is where the ticket is argued, and a second copy of the record is one more thing to keep in step.
+
+`AGENTS.md` has a branch that ran no review record that, because a missing record reads exactly like a round that raised nothing. #300 merged two test files with no round run and no record; #305, of the same shape, ran one that raised three findings and posted them. Nothing in #300 tells the two apart, and it opened 81 minutes after #299 put the review instructions in the always-loaded closure — so what it was missing was a clause and not the reach to find one. One sentence is the whole record when no round ran, and it is what makes a skipped review a decision rather than the absence of one.
 
 A record names the snapshot it read, which is why `AGENTS.md` has the branch pushed and the pull request open before one is written: a SHA on an unpushed branch resolves for nobody. #280's review was posted to the issue while its branch was still local, so every SHA in it cited a commit no reader could open, and the record had to be moved (PR #286). For the same reason, rebase before the review and not after — a rebase rewrites the SHAs a published record already named.
 
--- a/design/architecture.md
+++ b/design/architecture.md
@@ -672,14 +672,10 @@ material. They are separate from the generated package site in `docs/`;
 packages. Internal architecture is not duplicated into the README or
 vignettes.
 
-A review is not written down a second time. What a review produced records it:
-the commit that names a finding, the test that fails without the fix, the diff
-that changed the prose. What a review decided without producing anything goes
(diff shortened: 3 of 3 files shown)


sayuks:
> *This review was run by AI.*

## Review round 1

(`git diff 9e5215f...HEAD`), both axes, on `6227bc9` — amended to `2e18f8c` by
what the round produced. Both sub-agents were passed *Answering a review*, which
bounds a finding about prose to the four kinds it names.

Standards carried `AGENTS.md`'s *Always-loaded context*, *Code review*, and
*Code comments*, `design/architecture.md`'s *Repository placement* and
*Answering a review*, and `design/agents/code-review.md` in full. Spec carried
#306 and this pull request's body.

### Standards

**The sentence the branch rewrote was false, and the rewrite broadened it.
Deleted in both places.** `AGENTS.md` said the skill "composes its own sub-agent
prompts, so what an invocation does not carry does not reach either axis", and
this branch had compressed that to "reaches only what an invocation carries".
The Standards axis reported from inside its own context that `CLAUDE.md` and
`AGENTS.md` were in it and that the invocation had not carried them: the
always-loaded closure reaches a sub-agent whichever way the call is written.
`design/agents/code-review.md` held the same claim precisely, as "and nothing
else reaches them", and it is false there too. Both are deleted rather than
restated, per *Answering a review*. Two intermediate rewrites were tried and
discarded first — "reaches only what an invocation carries" and "neither axis
reads a file it was not given" — the second false in the other direction, since
a sub-agent has tools and reads what it decides to.

The deletion is also what pays for the new clause, replacing the compression
this branch opened with. The instruction it sat in front of stands without it.

**A count that disagreed with its own source. Deleted.**
`design/architecture.md` said #288 retired a ledger holding the dispositions
"a third time"; `design/agents/code-review.md` and #288 itself both say a
second. The branch had kept the disputed word while deleting the prose around
it. The claim carries nothing, so it is gone rather than adjudicated.

**Two timings that do not hold. One deleted, one measured.** "#305 ran one an
hour later" — the record was posted 83 minutes after #300 merged. The commit
message said #300 merged "70 minutes after" #299, and no pairing of those
timestamps yields 70: #299 merged `23:42:07Z`, #300 opened `01:03:32Z` (81
minutes) and merged `01:15:20Z` (93). The decorative one is deleted; the
load-bearing one — that #300 could reach the closure and found no clause in it —
is now stated as the 81 minutes it is, in the file, the commit message, #306,
and this body.

### Spec

**The fix reproduced the duplication it removes.** `design/agents/code-review.md`
restated the new `AGENTS.md` rule instead of citing it, where the paragraph two
below it uses the citing form ("which is why `AGENTS.md` has the branch
pushed"). Rewritten to cite. This is the branch's own subject, found by the axis
that was given the branch's own ticket.

Verified holding by that axis, each by command: #300 has zero reviews and no
record on either the pull request or its closing issue #257; #305's record
raises exactly three Standards findings; `design/architecture.md` restated five
of the six table rows and not more; `git diff 9e5215f...HEAD -- .github/` is
empty, so the baseline was not moved to make room; both of #306's asks are
delivered and nothing beyond them is.

### Verified clean

Every route the deleted enumeration named survives in the disposition table —
commit and test, prose diff, *Considered options*, dated `investigation/` note,
comment at the site. The sixth row was never in `design/architecture.md`, so
nothing was lost. `git grep` over `review-dispositions`, *Repository placement*,
and "routes them" finds no file left stale by the deletion.

### Gates

`verify-context-budget.R` passes at 38,753 of 38,810 — 57 bytes under, where the
branch opened at 38,800. All three files are `.Rbuildignore`d and no test reads
any of them, so the budget gate in `lint.yaml` is the only CI path the diff
reaches.

Issue #261, opened 2026-08-26
A leading or interior empty argument in a Grouping specification constructor reports an rlang argument name


sayuks:
## Problem

A leading or interior empty argument in a Grouping specification constructor
reports an rlang argument name the caller never wrote:

```
Error: argument "expr" is missing, with no default
```

`expr` is the local `grouping_arg_spec()` binds from
`rlang::quo_get_expr(arg)`; `conditionCall()` is `grouping_arg_spec(arg,
data_vars)`. The caller wrote a comma and nothing.

This is the shape #168 and #174 fixed elsewhere, at a site those tickets did
not reach.

## Expected behavior

An empty argument is a call the caller can rewrite, so it is refused with a
Package condition naming what is empty — or it is accepted as the empty
selection it resembles, which `dplyr::select()` does.

## Reproduction / evidence

```r
d <- data.frame(region = c("a", "b"), v = c(1, 2))

inspect_grouping(d, .grouping = grouping_sets(, region))
#> Error: argument "expr" is missing, with no default

inspect_grouping(d, .grouping = rollup(, region))
#> Error: argument "expr" is missing, with no default
```

Every constructor reaches it, because every one captures with
`rlang::enquos(...)` and every argument passes through `grouping_arg_spec()`.
An empty argument arrives there as R's missing marker, and the local binding is
what raises: `expr <- rlang::quo_get_expr(arg)` binds the marker, and forcing
`expr` at the next read — `is.symbol(expr)` — raises `missingArgError`. The
shape test never answers.

A *trailing* empty argument is already harmless, because `rlang::enquos(...)`
captures one argument for `f(a, )` and two for `f(, a)`:

```r
inspect_grouping(d, .grouping = grouping_sets(region, ))   # a normal plan
```

The condition is not one of this package's: it carries no `marginplyr_error`
class, so `?marginplyr`'s contract does not cover it, and ADR 0015's boundary
puts it on the wrong side — this is a call a caller can avoid by rewriting it.

For comparison, dplyr accepts the same shape in a selection:

```r
dplyr::select(d, , region)
#>   region
#> 1      a
#> 2      b
```

The repository already records how this family is answered. `R/utils.R`'s
comment above `nameable_call()` states the rule — "The shape test comes before
anything is bound to a local … `expr <- ...` would bind the missing marker and
raise `missingArgError` on the next read of it (#168, #174)" — and
`R/contextual-helpers.R`'s `static_spelling_reference_name()` implements it by
[...]

Issue #260, opened 2026-08-26
The two shapes a nested position recognizes as a specification have no asserted evaluation count


sayuks:
## Problem

ADR 0008 holds the number and timing of caller-quosure evaluations fixed
"without a separately accepted decision", and `design/architecture.md` repeats
it for this position. Of the shapes a nested Grouping specification argument can
take, two have their evaluation count asserted, and they are two the position
does *not* recognize as a specification. The two it does recognize — a nested
constructor call, and a name bound to a specification — are each evaluated three
times on the name-only path, and nothing asserts either.

`tests/testthat/test-grouping-plan.R`'s *recognizing a nested specification
adds no caller evaluation* pins the two:

- a caller's own function whose result is refused, at 1;
- a name-only selection helper, at 2, "once per compilation pass".

The two recognized forms cost one more than a selection, because
`preflight_grouping_spec()` evaluates the argument to find out what it is, and
then each compilation pass evaluates it again — every expansion site calls
`grouping_arg_spec()` for itself (`resolve_grouping_set()`,
`expand_grouping_sets()`, `resolve_grouping_units()`,
`expand_grouping_product()`). That is 1 + 2 = 3 where a selection costs 2.

So the test's title states a property for the two shapes where recognition does
not happen, and the two where it does are the ones it does not measure.

## Expected behavior

The count for each recognized form is either asserted at whatever it is
intended to be, or reduced to what the design intends and then asserted. Either
way the comments describing it say the number that holds.

## Reproduction / evidence

The binding has to be read from the environment the call is evaluated in, so
both probes below install it there.

```r
d <- data.frame(region = c("a", "b"), grade = c("x", "y"), v = c(1, 2))

# A name bound to a specification.
count <- 0L
makeActiveBinding("s", function() { count <<- count + 1L; rollup(v) }, environment())

count <- 0L; inspect_grouping(d, .grouping = grouping_sets(s)); count
#> [1] 3

# With a predicate elsewhere in the specification the plan is not settled by
# names alone, one compilation pass is withheld, and the count drops with it.
count <- 0L
inspect_grouping(d, .grouping = grouping_sets(s, where(is.character)))
count
#> [1] 2

# A nested constructor call, which is the other recognized form, costs the
# same. The caller's own binding is what runs, which ADR 0019 records.
calls <- 0L
rollup <- function(...) { calls <<- calls + 1L; marginplyr::rollup(...) }

[...]

Issue #190, opened 2026-08-15
A nested specification position refuses a genuine Grouping specification a caller's function returns


sayuks:
## Parent

Part of #23. Split out of #172, which settled the name-resolution rule and
deliberately left this diagnostic alone: the rule is about which spellings carry
package semantics, and this is about what a nested position does with a value
that already has them.

## What is wrong

A caller function that returns a genuine `margin_grouping_spec` is accepted at
the top level of `.grouping` and refused when nested, with a diagnostic that
contradicts itself — it reports the object as the wrong kind for a position
where a Grouping specification is exactly what belongs.

```r
d <- data.frame(region = c("E", "E", "W"), grade = c("a", "b", "a"), units = c(1, 3, 6))
my_spec <- function(...) rollup(...)

summarize_with_margins(d, t = sum(units), .grouping = my_spec(region))
#> works

summarize_with_margins(d, t = sum(units),
                       .grouping = grouping_sets(my_spec(region), grade))
#> Error: Can't select columns with `my_spec(region)`.
#> x `my_spec(region)` must be numeric or character, not a <margin_grouping_spec> object.
```

## Cause

`grouping_arg_spec()` (`R/grouping-plan.R:701-735`) evaluates a nested argument
only when its call head is a known constructor spelling, when it is a symbol, or
when it is not language. Anything else returns `NULL` and falls through to
tidyselect, which reports the specification as an unusable selection.

## Why the gate is not the thing to remove

A nested argument is ambiguous between a tidyselect selection and a nested
Grouping specification. Evaluating every nested call would run selections such
as `starts_with("re")` outside a selection context, and ADR-0008 holds the
number and timing of caller-quosure evaluations fixed. The gate is what
separates the two, so the fix is the diagnostic and the documented contract, not
the gate.

## What to build

State the contract — a nested specification position recognizes a constructor
spelling or a symbol — and make the failure say so. A `margin_grouping_spec`
reaching the selection path should be reported as a specification built by a
call the position does not evaluate, and should name the workaround, which was
measured to work:

```r
s <- my_spec(region)
summarize_with_margins(d, t = sum(units), .grouping = grouping_sets(s, grade))
#> works
```

## Acceptance criteria

- [ ] A nested argument whose value is a `margin_grouping_spec` fails with a
      `marginplyr_error` that names the position, the recognized forms, and the
      bind-to-a-symbol workaround, rather than with a tidyselect subscript error.
[...]

Issue #178, opened 2026-08-13
Treat redundant parentheses as transparent around recognized Margin calls


sayuks:
## Parent

Part of #23. This ticket records a P1 PARTIAL / REGRESSION finding from the repository-wide review of `a473e5f3560e4e7572ad02302d7090a369fd8239`.

## What to build

Give semantically equivalent direct and redundantly parenthesized calls the same static meaning throughout Margin planning. Recognition of supported summary helpers, grouping constructors, tidyselect predicates, and dplyr context helpers must depend on callable identity rather than a single surface spelling.

Apply the helper-resolution decision made in #172. Keep the conservative safety policy from #130 for genuinely computed or unresolved call heads.

## Acceptance criteria

- [ ] Direct and redundantly parenthesized forms of `pick()`, `cur_group_id()`, `where()`, nested grouping constructors, and supported contextual helpers have identical public behavior.
- [ ] Bare and namespace-qualified forms follow the resolution rule established in #172.
- [ ] Parentheses do not leak grouping-plan columns, change selected columns, or bypass a prohibited-context error.
- [ ] Genuinely computed or unresolved call heads remain subject to the conservative #130 safety rule.
- [ ] Formula and quosure environments and other identity-relevant attributes remain intact.
- [ ] One shared callable-identity rule protects sibling analyzers instead of enumerating only the demonstrated spellings.
- [ ] End-to-end tests cover local results, lazy planning where applicable, and typed error behavior.

## Blocked by

- #172

PR #308, opened 2026-08-29
Read a nested argument through its redundant parentheses

### LINKED ISSUE #259: A nested specification position does not read through redundant parentheses when recognizing a name, or when deciding whose diagnostic a refusal gets
## Problem

A nested Grouping specification position reads a redundantly parenthesized
*call* as the call it wraps in one place and not in another, so two spellings
of one argument are read two ways, and neither difference was decided.

`?grouping_set` states the rule the position implements:

> A nested Grouping specification is recognized by how it is written: a call to
> one of these constructors, or a name bound to a specification.

#178 established that a recognized spelling is read through redundant
parentheses, so that `(f)(x)`, `(f(x))`, and `f(x)` are one call written three
ways. `R/utils.R` owns that reading — `unparenthesized_value()`,
`unparenthesized_name()`, `unparenthesized_call()`, and `is_redundant_parens()`
— and `design/architecture.md`'s *Static expression reading (`R/utils.R`)*
section records that it lives there so that "every analysis above" gets the
property at once.

Two readings in `R/grouping-plan.R` do not have it:

1. **Recognition of a name.** `grouping_arg_spec()` asks `is.symbol(expr)`, and
   `(s)` is a call to `(`, not a symbol, so a parenthesized name bound to a
   specification is never read as a name.
2. **The label comparison that decides whose diagnostic a refused selection
   gets.** `is_grouping_spec_subscript()` compares the refused subscript with
   `rlang::as_label()` of the argument, so that a specification written *inside*
   a selection keeps tidyselect's report. Neither label is unparenthesized, so a
   parenthesized argument never matches itself, and #190's diagnostic is
   withheld from it.

## Expected behavior

A redundantly parenthesized argument is read as the argument it wraps, in every
reading that decides what a nested position does with it — or the reference
page says which readings it is not read through, and why.

## Reproduction / evidence

```r
d <- data.frame(region = c("a", "b"), v = c(1, 2))
s <- rollup(v)
f <- function(...) rollup(...)

# A parenthesized constructor call is recognized (#178).
inspect_grouping(d, .grouping = grouping_sets((rollup(region))))
#> # A tibble: 2 × 6
#>   set_id fixed included omitted  grouping_bits grouping_id
#>    <int> <chr> <chr>    <chr>    <chr>               <int>
#> 1      1 ()    (region) ()       region=0                0
#> 2      2 ()    ()       (region) region=1                1

# The same binding, written bare, is recognized.
inspect_grouping(d, .grouping = grouping_sets(s))
#> # A tibble: 2 × 6
#>   set_id fixed included omitted grouping_bits grouping_id
#>    <int> <chr> <chr>    <chr>   <chr>               <int>
#> 1      1 ()    (v)      ()      v=0                     0
#> 2      2 ()    ()       (v)     v=1                     1

# Parenthesized, it is not, and tidyselect reports it.
inspect_grouping(d, .grouping = grouping_sets((s)))
#> Error:
#> ! Can't select columns with `s`.
#> ✖ `s` must be numeric or character, not a <margin_grouping_spec> object.
```

The second reading is visible without any binding at all. #190 exists so that a
specification arriving where a selection is expected is reported in
marginplyr's own words, and a redundant pair of parentheses is enough to
withhold it:

```r
inspect_grouping(d, .grouping = grouping_sets(f(region)))
#> Error:
#> ! `f(region)` is a grouping specification, but a nested position recognizes
#>   one only when it is a call to `grouping_set()`, `grouping_sets()`,
#>   `rollup()`, `cube()`, or `grouping_spec()`, or a name bound to a
#>   specification.
#> ℹ Anything else is read as a column selection.
#> ℹ Assign the specification to a name first, then use that name here.

inspect_grouping(d, .grouping = grouping_sets((f(region))))
#> Error:
#> ! Can't select columns with `f(region)`.
#> ✖ `f(region)` must be numeric or character, not a <margin_grouping_spec> object.
```

The condition carries `subscript_arg = "f(region)"` while
`rlang::as_label(quote((f(region))))` is `"(f(region))"`. The labels differ, the
condition is re-raised as it arrived, and the caller receives a bare
`vctrs_error_subscript_type`.

Where the input also has a column of that name, the parenthesized name selects
that column silently, which is #255's failure mode in the one spelling a fix
gating on `is.symbol(expr)` cannot reach — and every disposition #255 offers
that changes behaviour gates there.

## Decision required

Two dispositions are available and they produce different behaviour, so which
one is right is a maintainer's call:

- **Read through, in both readings.** Consult `is_redundant_parens()` and
  `unparenthesized_value()` before asking `is.symbol()`, and unparenthesize both
  sides of the label comparison. Gives every reading of a nested argument one
  parenthesis rule, which is the argument #178 made.
- **Read through in the label comparison only, and document the name.** State on
  the reference page that only a call is recognized through parentheses, so that
  `(s)` is a column selection by rule rather than by omission — and still fix the
  comparison, since `(f(region))` is a caller's own function whichever way the
  name is decided.

Both dispositions change the label comparison; they differ over the name.
Recording the decision on this ticket is what converts it to `ready-for-agent`.

## What to build

Whichever disposition is chosen, `grouping_sets((s))` and `grouping_sets(s)`,
and `grouping_sets((f(region)))` and `grouping_sets(f(region))`, are each
predictable from one written rule, and both cases are covered by a test rather
than left to which of two shapes the argument happens to have.

## Acceptance criteria

- [ ] `grouping_sets((s))`, for a name bound to a specification, resolves to the
      same Grouping plan as `grouping_sets(s)` — or the reference page says that
      only a call is recognized through parentheses.
- [ ] `grouping_sets((f(region)))`, for a caller's own function returning a
      specification, raises the same `marginplyr_error` as
      `grouping_sets(f(region))`.
- [ ] The behaviour is the same in every nested position, derived from
      `grouping_kind_rules()` rather than listed.
- [ ] A parenthesized constructor call keeps the reading #178 gave it, and
      multiple redundant parentheses behave as one, as they do for a call.
- [ ] A specification written inside a selection still keeps tidyselect's own
      report, parenthesized or not — `c(s, region)` and `c((s), region)` alike.
      This is the property `is_grouping_spec_subscript()` exists for and the one
      a change to it can break.
- [ ] The number of times a caller's nested quosure is evaluated is unchanged
      for every argument whose reading does not change. A `(s)` that becomes a
      recognized name gains the count a recognized name has, which #260
      describes.

## Testing seam

`inspect_grouping()`, the public reading of a resolved Grouping plan (ADR 0013),
beside the nested-position tests in `test-grouping-plan.R`. The withheld
diagnostic is asserted on the complete message, as the #190 case beside it is.

## Scope boundaries

Deliberately unchanged:

- The reading of a bare, unparenthesized name, which is #255's.
- Tidyselect's own precedence inside a selection.
- `is_redundant_parens()` and `unparenthesized_value()` themselves, which #178
  decided. This ticket asks further callers to use them; it does not change what
  they answer.

## Related Issues

- **#178** — established the read-through for a recognized spelling, and the
  argument this ticket applies to two readings that did not get it. The
  `(share_of_total)` reasoning is recorded in `R/contextual-helpers.R`'s comment
  above `static_spelling_reference_name()`.
- **#190** — added `abort_nested_grouping_spec()` and
  `is_grouping_spec_subscript()`, which is what withholds it here.
- **#255** — the same position's other unwritten rule. Independent: #255 is
  about which reading a bare name gets, this is about whether a parenthesized
  argument is that argument at all.
- **#260** — the evaluation counts of the recognized forms. A `(s)` this ticket
  makes recognized gains those counts, so the two tickets must not both pin
  them.
- **#261** — touches the same two lines of `grouping_arg_spec()` in the opposite
  direction: the obvious spelling of this ticket's read-through binds a local
  before testing the shape, which is exactly what #261 says must not happen.
  Satisfying both means the recursion idiom `static_spelling_reference_name()`
  uses.

Searched the open set for this root cause under parenthesis, spelling,
recognition, and nested specification terminology before filing. Nothing covers
it.

## Blocks

- #40

## Review origin

Found while investigating #255, and verified independently of it: reproduced
from a clean session on `main`, with and without a colliding column, against
both recognized forms and against a caller's own function, with the label
comparison read from the raised condition. No temporary change to the package
was needed.

sayuks:
Closes #259.

## What this changes

Three readings a Nested specification position makes now read through a
redundantly parenthesized argument, as every reading taken from `R/utils.R`'s
shared readers already did (#178, ADR 0019):

| written | before | after |
|---|---|---|
| `grouping_sets((s))`, `s` bound to a specification | column selection | the specification, as `grouping_sets(s)` |
| `grouping_sets(((!!s)))` | column selection | the specification |
| `grouping_sets((s))`, `s` also a column | selects the column, silently | ADR 0026's ambiguity refusal, as `s` gets |
| `grouping_sets((f(region)))`, `f` a caller's function | tidyselect's subscript error | #190's `marginplyr_error`, as `f(region)` gets |
| `grouping_sets((rollup(region)))` | the specification | unchanged |
| `grouping_sets((!!s))` | the specification | unchanged — `enquos()` drops one pair around `!!` at capture |
| `grouping_sets(c((s), region))` | tidyselect's report | unchanged |

## Decision

#259 required a maintainer's call between two dispositions. The decision is
[recorded on the ticket](https://github.com/sayuks/marginplyr/issues/259#issuecomment-5460036820):
read through in every reading, rather than documenting the name as the one
position the rule does not reach. It rests on two arguments — the rule the
reference page states is about how an argument is written, and `(` is not a way
of writing anything; and `(s)` selected a colliding column silently while `s`
was refused, which is the reading ADR 0026 exists to remove. A third argument
the comment gave was measured false in review and is
[withdrawn](https://github.com/sayuks/marginplyr/issues/259#issuecomment-5460277179).

## How

- `grouping_arg_spec()` restarts on the argument the pair wraps, by recursion
  rather than by rebinding `expr` — the rule `R/utils.R` states and
  `static_spelling_reference_name()` follows, since what a pair wraps may be
  R's empty argument.
- All three sites take the reading from `nested_arg_expr()`, which is
  `unparenthesized_value()` put to a quosure. Nothing re-implements it.

## Evaluation counts

A parenthesized argument gains the count of the argument it wraps; no bare
argument's count moves. Measured against `main`'s `R/` through
`inspect_grouping()`, with the binding an active binding, and re-measured
independently in review over 5 constructors × 10 shapes × 0/1/2 pairs:

| `grouping_sets(…)` | main | branch |
|---|---|---|
| `s` | 3 | 3 |
| `(s)` | 1 | 3 |
| `((s))` | 1 | 3 |

ADR 0019's third amendment records that as the decision it is, and reaches the
ticket's sixth criterion: the criterion blesses the `(s)` that becomes a
recognized name, and the shape between — a parenthesized name bound to
something that is *not* a specification, whose plan is unchanged while the gate
now reads the binding to establish that — is accepted with the criterion's
second sentence rather than against its first.

What the branch *asserts* is the equality of the two spellings' counts, since
pinning each form's number is #260's and nothing may pin it twice.

## #261

Its diagnostic's message and `conditionCall()` are byte-identical to `main`.
The condition's class is not, once `grouping_arg_spec()` is byte-compiled —
`evalError` becomes `getvarError`. That is the compiler's reading of a longer
body and not the raise line, which was checked by moving it. Both are base R's
untyped errors on the path #261 exists to replace, and `?marginplyr`'s contract
covers neither.

## Verification

- `testthat::test_local()` with `NOT_CRAN=true`: full suite passes; the new
  assertions fail against `main`'s `R/`.
- `lintr::lint_package()`: no lints. `spelling::spell_check_package()`: none.
- `verify-suite-coverage.R`: 630 tests, none executing in zero configurations.
- `verify-context-budget.R`: unchanged at baseline.
- All 26 CI checks passed on 4b930cc and again on 634fcf8.

## Review

Three rounds of `mattpocock-skills:code-review`. Round 1 ran both axes and
raised eight findings, answered in 8f1a491. Round 2 verified those answers
against the evidence they landed with and reviewed the executable behaviour
round 1 had not seen — it found the branch's own evaluation-count claim false,
answered in 634fcf8. Round 3 fact-checked every claim the branch's prose makes,
after two measurement errors, and found two more, answered in e40d16d. All
three records are in the comments below.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -20,7 +20,9 @@ An argument of a Grouping specification constructor, where a nested Grouping
 specification and a column selection are both allowed. Which one is meant is
 decided by how the argument is written: a call to a constructor, or a name
 bound to a specification, is a nested specification, and every other argument
-is a column selection. A spelling the position does not recognize is never
+is a column selection. Redundant parentheses are transparent to that reading,
+around the name or around the whole call, because `(` is the identity
+function. A spelling the position does not recognize is never
 evaluated to find out, because a selection such as `starts_with("re")` has no
 meaning outside a selection context. Both readings claim one argument at once:
 a bare name that is a column of the input and is also bound, in the caller's
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,7 +15,11 @@
   a column of your input that is also bound to a nested specification the
   position accepts — is refused rather than resolved by whichever the data
   happens to have, and the refusal names the spelling for each reading:
-  `all_of("s")` for the column, `!!s` for the specification.
+  `all_of("s")` for the column, `!!s` for the specification. Redundant
+  parentheses are transparent to every one of these readings, as they are
+  elsewhere, because `(` is the identity function: `(s)`, `(!!s)`, and
+  `(rollup(region))` are the arguments they wrap, however many pairs deep, and
+  both refusals reach a parenthesized argument as they reach a bare one.
 * Added contextual `grouping_bit()` and `grouping_id()` summary helpers.
 * Added the contextual `share_of_parent()` summary helper, which divides a
   preceding numeric scalar summary by the same measure one `rollup()` level
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -359,6 +359,14 @@ preflight_grouping_spec <- function(grouping_spec, data_vars) {
   list(spec = grouping_spec, name_only = name_only)
 }
 
+# The spelling a nested position reads an argument by: the caller's expression
+# with its redundant parentheses removed (#178, #259). The argument may be one
+# a caller left empty, so the answer may be R's empty argument, and a reader
+# passes it on as a value rather than binding it (#168, #174).
+nested_arg_expr <- function(arg) {
+  unparenthesized_value(rlang::quo_get_expr(arg))
+}
+
 # Refuses the one argument both readings claim: a bare name that is a column of
 # the input and is bound to a specification of a kind this position admits.
 # ADR 0026 holds that decision -- why such a name is refused rather than
@@ -370,6 +378,12 @@ preflight_grouping_spec <- function(grouping_spec, data_vars) {
 # "selection" for a symbol, asked again here because the gate reports which
 # reading it took and not why. A name nothing binds has no second reading, and
 # a name the data does not hold is resolved by the gate itself.
+#
+# The name is read through the gate's own parenthesis reading, so a refusal
+# `s` gets is a refusal `(s)` gets. Reading the caller's expression here
+# instead would withhold it from the parenthesized spelling, which is the
+# reading ADR 0026 refuses reached by a pair the gate has already dropped
+# (#259).
 #
 # The kinds this position admits are asked before the binding is read: where a
 # position admits none, the answer is already known, and reading a caller's
@@ -410,10 +424,10 @@ preflight_grouping_spec <- function(grouping_spec, data_vars) {
 # lookup twice. It is derived from the parent's kind, so the pair handed to the
 # memo below carries nothing that memo's key does not.
 check_ambiguous_nested_name <- function(arg, parent, rule, data_vars) {
-  if (!is_name_part(rlang::quo_get_expr(arg))) {
+  if (!is_name_part(nested_arg_expr(arg))) {
     return(invisible(NULL))
   }
-  name <- rlang::as_string(rlang::quo_get_expr(arg))
+  name <- rlang::as_string(nested_arg_expr(arg))
   if (
     !name %in% data_vars ||
       !rlang::env_has(rlang::quo_get_env(arg), name, inherit = TRUE)
@@ -879,6 +893,28 @@ grouping_constructor_names <- function() {
 
 grouping_arg_spec <- function(arg, data_vars) {
   expr <- rlang::quo_get_expr(arg)
+  # A pair of redundant parentheses is read through here as it is everywhere
+  # else (#178, ADR 0019): `(` is the identity function, so `(s)` and `s` are
+  # one argument written two ways and every reading below answers the two
+  # alike. The shared readers give a *call* that property already, which is why
+  # `(rollup(region))` was recognized while `(s)` and `(!!s)` were not -- a
+  # shape test put to the caller's own expression reaches none of them (#259).
+  #
+  # By restarting on the argument the pair wraps rather than by rebinding
+  # `expr`, which is the rule `R/utils.R` states and
+  # `static_spelling_reference_name()` follows: what the pair wraps may be R's
+  # empty argument, and binding the missing marker raises `missingArgError` on
+  # the next read of it (#168, #174). `is_parenthesized()` answers `FALSE` for
+  # a constructed pair holding one, so the restart cannot repeat.
+  if (is_parenthesized(expr)) {
+    return(grouping_arg_spec(
+      rlang::new_quosure(
+        unparenthesized_value(expr),
+        rlang::quo_get_env(arg)
+      ),
+      data_vars
+    ))
+  }
   if (
     is.symbol(expr) &&
       is_name_only_expr(
@@ -943,7 +979,7 @@ resolve_grouping_selection <- function(arg, data_proxy) {
       on_rename = abort_grouping_rename
     ),
     error = function(cnd) {
-      label <- rlang::as_label(rlang::quo_get_expr(arg))
+      label <- rlang::as_label(nested_arg_expr(arg))
       if (!is_grouping_spec_subscript(cnd, label)) {
         stop(cnd)
       }
@@ -963,6 +999,16 @@ resolve_grouping_selection <- function(arg, data_proxy) {
 # they have already made. Comparing the labels is what separates the two, and
 # both are written by `rlang::as_label()` from the same expression when the
 # argument as a whole is what was refused.
+#
+# Both sides of that comparison are unparenthesized, which is what makes them
+# comparable at all. tidyselect descends into a `(` call before it refuses
+# anything, so `(f(region))` reaches here reported as `f(region)`, and a label
+# taken from the caller's expression matched neither itself nor the argument it
+# wraps -- so #190's diagnostic was withheld from the parenthesized spelling
+# and the caller received the tidyselect report this exists to replace (#259).
+# `nested_arg_expr()` is what the label is written from for that reason. The
+# sub-selection case is unaffected: `c((s), region)` labels the whole argument,
+# which is not the `s` tidyselect refused inside it.
 #
 # Where a column shares the name, tidyselect refuses nothing and none of this
 # runs: `c(s, region)` selects the column `s`, which is the one reading a
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -893,25 +893,21 @@ grouping_constructor_names <- function() {
 
 grouping_arg_spec <- function(arg, data_vars) {
   expr <- rlang::quo_get_expr(arg)
-  # A pair of redundant parentheses is read through here as it is everywhere
-  # else (#178, ADR 0019): `(` is the identity function, so `(s)` and `s` are
-  # one argument written two ways and every reading below answers the two
-  # alike. The shared readers give a *call* that property already, which is why
-  # `(rollup(region))` was recognized while `(s)` and `(!!s)` were not -- a
-  # shape test put to the caller's own expression reaches none of them (#259).
+  # A redundantly parenthesized argument is read as the argument it wraps, as
+  # it is wherever a reading is taken from `R/utils.R`'s readers (#178,
+  # ADR 0019). The tests below take none: they ask `is.symbol()` and
+  # `is.language()` about the caller's own expression, so this site restarts on
+  # what the pair wraps rather than inheriting the reading (#259).
   #
-  # By restarting on the argument the pair wraps rather than by rebinding
-  # `expr`, which is the rule `R/utils.R` states and
-  # `static_spelling_reference_name()` follows: what the pair wraps may be R's
-  # empty argument, and binding the missing marker raises `missingArgError` on
-  # the next read of it (#168, #174). `is_parenthesized()` answers `FALSE` for
-  # a constructed pair holding one, so the restart cannot repeat.
+  # By restart rather than by rebinding `expr`, which is the rule `R/utils.R`
+  # states and `static_spelling_reference_name()` follows: what a pair wraps
+  # may be R's empty argument, and binding the missing marker raises
+  # `missingArgError` on the next read of it (#168, #174). `is_parenthesized()`
+  # answers `FALSE` for a constructed pair holding one, so the restart cannot
+  # repeat.
   if (is_parenthesized(expr)) {
     return(grouping_arg_spec(
-      rlang::new_quosure(
-        unparenthesized_value(expr),
-        rlang::quo_get_env(arg)
-      ),
+      rlang::new_quosure(nested_arg_expr(arg), rlang::quo_get_env(arg)),
       data_vars
     ))
   }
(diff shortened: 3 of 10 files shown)


sayuks:
## Review round 1 — `mattpocock-skills:code-review`, base 9e5215f, subject 4b930cc

Two axes, parallel sub-agents. Answered in 8f1a491.

### Standards — 5 findings

| # | finding | disposition |
|---|---|---|
| 1 | `is_parenthesized()`'s header claims "the three walks that restart share one test"; this branch is a fourth caller and not a walk | fixed in prose — the count is deleted rather than moved to four, since a count is what goes stale again |
| 2 | Two comments carry past-tense defect history, which `AGENTS.md`'s *Code comments* puts in the commit message and the ADR | fixed in prose — rewritten in the present tense |
| 3 | The restart's comment restates ADR 0019's own premise beside the citation to it | fixed in prose — what is left is which way this site falls under the ADR |
| 4 | One `label` is both the comparison key and the subject the refusal names, so `(f(region))` is refused in the words `f(region)` is — undispositioned against ADR 0024 | kept, dispositioned in ADR 0019's third amendment; the site cites it in one line |
| 5 | `nested_arg_expr()`'s header over-claims while the position's own gate does not call it | fixed in code — the gate's restart now builds its quosure from it, so one helper carries the reading and the header holds |

Finding 4 is kept on the consequence the first amendment already accepts — "a diagnostic dplyr raises about that argument echoes it without the pair" — and because the ticket asks for *the same* `marginplyr_error` as the bare spelling. What ADR 0024 was written for is a value whose bytes cli would otherwise collapse, which no parenthesis rule touches.

### Spec — 3 findings, one of which corrected a claim this branch made

| # | finding | disposition |
|---|---|---|
| 1 | `expect_identical(count_reads(quote(s)), 2L)` pins a number #260 owns, while the test's own comment and the ADR both say nothing here pins one | fixed in code — the pin is gone; the mechanism check is a positive and a zero |
| 2 | The count equality covers only shapes already equal by construction; the shapes whose count the read-through moved are counted nowhere | fixed in code and in prose — see below |
| 3 | `outcome()` compares `$sets` while criterion 1 asks for the plan | fixed in code — it compares the whole plan; free, the sets having agreed for all forty comparisons |

Finding 2 is the one worth reading. Measuring what to put in its place found **this branch's own claim wrong**. It said a literal gains one evaluation, because `(1)` is now read as `1` is. Measured against the pre-#259 package over eight argument shapes, each written bare and wrapped in one pair and in two, with the binding an active binding and the caller's function counting its own calls: **every count is the count it was**. Nothing a caller can observe moved, in either direction. Criterion 6 is met as literally written rather than in the weaker reading the decision comment offered, the ADR says so, and the shapes the gate newly evaluates without recognizing — a literal, an injected object — are constants, which is why they cannot be counted rather than why they cost nothing.

The count test now covers both branches the gate evaluates on and both it declines on, including a name bound to something that is not a specification, which the previous pair did not reach.

No scope creep found on either axis; `R/utils.R`'s readers, the bare-name reading, and tidyselect's precedence inside a selection are unchanged, and #261's diagnostic is byte-identical to `main` — message, `conditionCall()`, and class.

All 26 checks passed on 4b930cc, including every `backend` job, `depends-only`, `document`, `altdoc`, and `Suite coverage by construction`.

sayuks:
## Review round 2 — verification of round 1's answers, plus the behaviour round 1 did not see

Scoped as `design/agents/code-review.md` directs: verify each answer against the evidence it landed with, and review only executable behaviour the first run did not see. Answered in 634fcf8.

### Part 1 — all eight answers hold

Verified at the sites: `is_parenthesized()`'s header claims no count and all four callers restart; both comments are present-tense; the `(`-is-identity derivation is gone from the restart's comment; ADR 0019's third amendment decides the single-`label` question rather than restating it; the `2L` pin is gone and what replaced it collides with none of #260's numbers; `outcome()` is not vacuous — 21 of 40 bare outcomes are plans.

Two observations, both dispositioned without code changes:

- **ADR 0024 carried no back-reference**, so its text read unconditionally while ADR 0019's third amendment scopes it. Fixed in prose — 0024 now says what it is about and where a normalized expression is decided.
- **`is_name_only_selection()` reads the caller's expression, not `nested_arg_expr()`.** Rejected. The two answer different questions: `is_name_only_expr()` recurses into `(` wherever it appears in a selection tree, which is what a name-only test needs and strictly more than the outer pair the gate reads. The reading test holds them together — `name_only` decides how the plan compiles, so a divergence changes a plan or a condition in one of its forty comparisons.

### Part 2 — and the finding that matters

The restart's swap to `nested_arg_expr(arg)` is identical, verified including for a constructed `(` holding R's empty argument. The revised tests still fail against `main`'s `R/`.

**The central claim `8f1a491` made is false, and I had measured it wrong.** The probe behind "no count moves" reverted `R/` with `git stash push -- R/`, which reverts to `HEAD` — and `HEAD` was already `4b930cc`, the commit carrying the fix. Both columns were the same package. Measured against `main`'s `R/` through `inspect_grouping()`:

| `grouping_sets(…)`, `s` an active binding | main | branch |
|---|---|---|
| `s` | 3 | 3 |
| `(s)` | **1** | **3** |
| `((s))` | **1** | **3** |

— for a binding that is a specification, a character vector, or a number alike. So: **no bare argument's count moves, and a parenthesized name gains what the bare name costs.** That is what this branch said originally, before `8f1a491` "corrected" it; the correction was the error.

The restatement is not only arithmetic. The false claim said the read cost nothing, so nothing weighed it. The amendment now records it as the decision it is, and reaches the acceptance criterion the false claim let the branch skip: criterion 6 blesses the recognized `(s)` that gains a count, and the shape between — a parenthesized name bound to something that is *not* a specification, whose plan is unchanged while the gate now reads the binding to establish that — is accepted with the criterion's second sentence rather than against its first.

Nothing in `R/` or in the tests changed. The equality the count test asserts was already the right assertion and already fails against `main`.

The correction is also [on the ticket](https://github.com/sayuks/marginplyr/issues/259), where the decision comment carried the same wrong claim.

sayuks:
## Review round 3 — fact-check of every claim the branch's prose makes

Scoped to that alone, because two measurement errors had already been made on this branch. Answered in e40d16d, which changes no code and no test.

### Two claims were false

**1. `(!!s)` was never a column selection.** ADR 0019's third amendment said `(s)` and `(!!s)` both were, and 4b930cc added that the second is the spelling `abort_ambiguous_nested_name()` prints as the remedy for the first — which would make that remedy something a pair of parentheses could withhold. `rlang::enquos()` drops one pair around `!!` as it captures:

```r
rlang::expr((!!1))    #> 1
rlang::expr(((!!1)))  #> (1)
```

Measured against a pristine `main`: `grouping_sets((!!s))` compiled to the specification there; `grouping_sets(((!!s)))` was the selection. The injected spelling was reached by the *second* pair, not the first. Fixed in prose; the corresponding argument in the ticket's decision comment is [withdrawn rather than repaired](https://github.com/sayuks/marginplyr/issues/259#issuecomment-5460277179), since a second pair withholding a remedy is a different claim and not one this decision would have been written on. The decision rests on the other two arguments, either of which settles it alone. What the branch does is unchanged and still right: `((!!s))` is recognized now and was not.

**2. #261's condition is not class-identical.** Message and `conditionCall()` are byte-identical to `main`; the class is not, once `grouping_arg_spec()` is byte-compiled — `evalError` becomes `getvarError`. It is not the raise line: I moved the restart below the name test so the first read of `expr` is the read that was always first, and the class still changed, so what differs is the compiler's reading of a longer body. That ordering is not kept — it bought nothing and its comment claimed what it did not do.

Nothing is written about this where the code can be read, because there is nothing there to describe: both are base R's untyped errors on the path #261 exists to replace, `?marginplyr`'s contract covers neither, and what a caller reads does not move. This record is where the measurement lives.

### Everything else checked holds

Independently re-measured, 5 constructors × 10 shapes × 0/1/2 pairs: no bare count moves anywhere, `grouping_sets((s))` goes 1→3 for a specification, a character vector, and a number alike, and the wrapped count equals the bare count for **every** shape on the branch, including ones the amendment does not name. A parenthesized constructor call is 0/0/0 on both sides and was already recognized on `main`; a parenthesized caller's call is 1/1/1 on both.

Also confirmed: `main`'s tidyselect reports `(f(region))` as `f(region)`; the branch's refusal is byte-identical to the bare spelling's; `c((s), region)` keeps tidyselect's report before and after; `nested_arg_expr()` returns the empty argument without error and is never bound; ADR 0026, ADR 0024, and #259's sixth criterion say what the branch attributes to them.

All 26 checks passed on 634fcf8.

PR #309, opened 2026-08-29
Read a recognized nested specification argument once

### LINKED ISSUE #260: The two shapes a nested position recognizes as a specification have no asserted evaluation count

sayuks:
Closes #260.

## Disposition

The ticket required a maintainer's decision between *Assert what is there* and *Reduce it first*. **Reduce it first** was chosen and recorded on the ticket (#260 comment); the reasoning is there rather than repeated here.

## What changed

The two shapes a Nested specification position recognizes as a specification — a nested constructor call, and a bare name bound to one — were each evaluated three times where the plan is settled by names alone and twice where one compilation pass is withheld: once by `preflight_grouping_spec()` to find out what the argument is, and again by every expansion site, each of which called `grouping_arg_spec()` for itself.

The preflight now records what each nested argument resolved to, parallel to `spec$args`, and expansion reads that record instead of taking a reading of its own. A recognized argument is evaluated once per call, whichever passes run. `data_vars` leaves the whole expansion chain with the reading — it was what the gate decided a bare name against, and nothing else there asked for it.

Unchanged: which arguments are recognized and by what spelling; a column selection at one resolution per compilation pass; the colliding name ADR 0026 refuses, at one; the caller's own function #190 reports on, at one.

## Where the counts are written down

ADR 0008's new amendment *one reading of a recognized nested argument*, as a table over the four shapes. `design/architecture.md` and both comments in `R/grouping-plan.R` that describe the shared preflight cite it and state the number for the position they are about, rather than re-deriving it.

ADR 0026 gains *the spelling gate no longer runs per pass*: its *Where the decision is made* and *Decide in the spelling gate* sections both reject deciding a collision in the gate on a per-pass count the gate no longer has. The placement stands on the rest of those sections, which is untouched.

## Tests

`test-grouping-plan.R` gains *a recognized nested specification is evaluated once*, covering both forms on both paths. It derives the parent positions from `grouping_kind_rules()` via `admitted_nested_kinds()`, so a sixth kind is covered without editing it — four positions today (`grouping_sets`, `rollup`, `cube`, `grouping_spec`); `grouping_set` admits no nested kind and is excluded by the derivation, and the set is asserted non-empty before anything iterates over it. Each counter is shown to report a reading before a count is concluded from it. A counted name-only selection travels beside the recognized argument in every specification it compiles, because a count of one is also what a call that compiled nothing would produce: it reads 2 on the name-only path and 1 where a predicate withholds a pass, so the paths are shown to differ while the recognized count does not. A `dbplyr::simulate_postgres()` input covers the name-only path; no test added here needs an optional backend.

The two counts already asserted in the sibling tests still hold. *a colliding nested name is read once, in the preflight* had two assertions of the old 3 and 2 for a non-colliding bound name; both are now 1, with the comments corrected. *recognizing a nested specification adds no caller evaluation* is renamed to *an unrecognized nested argument is evaluated as a selection is*, which is what it measures — the ticket's own observation about that title.

Verified the new test fails against the pre-change implementation (3 and 2 where 1 is expected), so it is load-bearing.

## Local gates run

- full `testthat::test_local()` — pass
- `lintr::lint_package()` — no lints
- `Rscript .github/scripts/verify-suite-coverage.R` — 631 tests, all execute in at least one single-backend configuration
- `Rscript .github/scripts/verify-context-budget.R` — 38753 / 38810 bytes, unchanged

No roxygen or `README.Rmd` change, so nothing generated needs regenerating.

## Out of scope, observed

`test-diagnostic-pluralization.R`'s `pluralizing_coverage()` table carries line references into `test-grouping-plan.R` that were already stale at the merge-base — `:915 and :768` for `compile_grouping_spec_impl`, whose two arms are actually near `:1591` and `:1738` there, and the four `abort_by_rename` / `abort_grouping_rename` references likewise. The table's own comment says a line reference is review surface and unchecked. Not touched here; worth a ticket.

### DIFF
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -173,11 +173,8 @@ prepare_grouping_plan <- function(.data,
       }
       # Preflighted once and handed to every compilation pass -- both of them
       # where the plan is settled by names alone, and the one below otherwise.
-      # This is the only reading of a nested argument the call makes, so a
-      # recognized one is evaluated once however many passes run; preflighting
-      # per pass would evaluate it once per pass. ADR-0008 holds the number and
-      # timing of evaluations fixed, and its amendment for a recognized nested
-      # argument is what sets it at one.
+      # This is the call's only reading of a nested argument, so a recognized
+      # one is evaluated once whichever passes run (ADR-0008).
       preflight <- preflight_grouping_spec(grouping_spec, data_vars)
       if (preflight$name_only) {
         # Reject name-only plan errors before acquiring typed metadata. The
@@ -335,17 +332,17 @@ grouping_name_proxy <- function(data_vars) {
   stats::setNames(as.list(seq_along(data_vars)), data_vars)
 }
 
-# The one site that reads a nested argument. What it returns is the
-# specification together with the reading each argument got, so the compilation
-# passes below take none of their own: `args` is parallel to `spec$args`, and
-# holds this function's own answer for a nested specification and `NULL` for a
-# column selection. ADR 0008's amendment for a recognized nested argument is
-# what accepts the reading moving here, and states the counts that hold.
+# The only site that reads a nested argument. It answers with the
+# specification, whether names alone settle the plan, and one record per
+# argument: `quo` is the caller's quosure and `nested` is this function's
+# answer for it, a preflight of the same shape for a nested specification and
+# `NULL` for a column selection. Expansion below reads that answer instead of
+# taking one, which is what holds a recognized argument to one evaluation
+# (ADR 0008).
 #
-# An empty `args` beside a non-`NULL` `spec` is a specification with no
-# arguments, not a specification nothing was read for: the two are told apart
-# by `spec`, which is `NULL` only for the absent `.grouping` that
-# `compile_grouping_spec_impl()` fills in.
+# `spec` is `NULL` only for the absent `.grouping`, which
+# `compile_grouping_spec_impl()` fills in. An empty `args` beside any other
+# `spec` is a specification a caller wrote no arguments for.
 preflight_grouping_spec <- function(grouping_spec, data_vars) {
   stopifnot(is.character(data_vars), !anyNA(data_vars))
   validate_grouping_spec_early(grouping_spec)
@@ -363,13 +360,14 @@ preflight_grouping_spec <- function(grouping_spec, data_vars) {
     if (is.null(nested)) {
       check_ambiguous_nested_name(arg, grouping_spec, rule, data_vars)
       name_only <- name_only && is_name_only_selection(arg, data_vars)
+      args[[i]] <- list(quo = arg, nested = NULL)
       next
     }
 
     nested_preflight <- preflight_grouping_spec(nested, data_vars)
     rule$validate_nested(grouping_spec, nested_preflight$spec)
     name_only <- name_only && nested_preflight$name_only
-    args[[i]] <- nested_preflight
+    args[[i]] <- list(quo = arg, nested = nested_preflight)
   }
   list(spec = grouping_spec, args = args, name_only = name_only)
 }
@@ -536,10 +534,8 @@ admitted_nested_kinds <- local({
 #
 # `preflight` carries the reading each nested argument got, so a caller running
 # more than one pass over one specification hands the same preflight to each
-# and no pass reads a caller's nested quosure at all. A recognized nested
-# argument is therefore evaluated once per call, whichever passes run
-# (ADR-0008). A column selection is not: each pass resolves one for itself,
-# against its own proxy.
+# and no pass reads a nested quosure of its own. A recognized nested argument
+# is therefore evaluated once per call, whichever passes run (ADR-0008).
 compile_grouping_spec <- function(
   .grouping,
   data_vars,
@@ -565,8 +561,8 @@ compile_grouping_spec <- function(
   )
 }
 
-# `preflight` rather than a specification, because expansion below reads each
-# nested argument's reading off it instead of taking one (ADR 0008).
+# Takes the preflight rather than the specification inside it, because
+# expansion below needs the reading each nested argument got (ADR 0008).
 compile_grouping_spec_impl <- function(preflight,
                                        data_vars,
                                        data_proxy,
@@ -602,9 +598,9 @@ compile_grouping_spec_impl <- function(preflight,
     )
   }
 
-  # The absent `.grouping`, which stands for the empty grouping set. It has no
-  # arguments, so there is nothing for the preflight to have read and the
-  # substitute is complete.
+  # The absent `.grouping` stands for the empty grouping set. Preflighted
+  # rather than written out here, so that one function answers what a record
+  # holds.
   if (is.null(preflight$spec)) {
     preflight <- preflight_grouping_spec(
       new_grouping_spec("set", list()),
@@ -698,12 +694,11 @@ compile_grouping_spec_impl <- function(preflight,
   )
 }
 
-# Expansion takes a preflight rather than a specification, so that a nested
-# argument's reading is read off `preflight$args` rather than taken again
-# (ADR 0008). `data_vars` is gone from the whole chain with the reading: it was
-# what `grouping_arg_spec()` decided a bare name against, and nothing else here
-# asked for it. A column selection is still resolved per pass, against the
-# proxy the pass was given.
+# The families a specification expands to. Every function below it takes the
+# preflight the caller holds, and reads each argument's reading off the record
+# beside that argument's quosure rather than taking one (ADR 0008). A column
+# selection is resolved here and not there, so it is resolved once per pass,
+# against the proxy that pass was given.
 expand_grouping_family <- function(preflight, data_proxy) {
   rule <- find_grouping_kind_rule(preflight$spec$type)
   if (is.null(rule)) {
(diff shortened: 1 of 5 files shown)


sayuks:
## Review record

One round, `mattpocock-skills:code-review`, both axes against the merge-base with `main` (`git diff origin/main...HEAD`). Read at `e9466c6`; answered at `4a579ef`.

Standards sources passed: the `AGENTS.md` sections the diff reaches, `design/architecture.md` (including *Answering a review*), `design/agents/code-review.md`, ADRs 0008, 0026, 0019, 0015, 0020, `CONTEXT.md`. Spec source: #260, with the maintainer decision in its comment naming *Reduce it first* as the disposition under review.

### Standards — 5 findings, 3 smells

| # | Finding | Disposition |
|---|---|---|
| 1 | `expand_grouping_family()`'s header written in the past tense about the removed `data_vars` parameter; `compile_grouping_spec_impl()`'s header the same shape | fixed in prose — both rewritten in the present tense, the changelog sentence deleted |
| 2 | Three comments in `R/grouping-plan.R` re-derive the count argument ADR 0008's amendment owns | fixed in prose — the re-derivations deleted; each site keeps the number that holds there and the citation |
| 3 | `design/architecture.md` states a row of that table one sentence before naming the table as its single home | fixed in prose — the row deleted, the pointer kept |
| 4 | The test's derivation claim is false as written: `admitted[[1L]]` covers a sixth kind as a parent only | fixed in code — iterates every admitted kind, 12 pairs where there were 4 |
| 5 | Four expansion sites gained an unwritten precondition that `preflight$args` parallels `preflight$spec$args` | fixed in code — each argument now carries its reading beside its quosure, so the precondition does not exist |
| smell | Duplicated Code / parallel arrays across five sites | fixed in code, same change as 5 |
| smell | Primitive Obsession — the preflight record is an unclassed list | rejected: it is built, read, and discarded inside `R/grouping-plan.R` and reaches no consumer outside it (`grep -rn 'preflight' R/ \| grep -v grouping-plan.R` finds only `preflight_shares()`, an unrelated name). That is what separates it from `margin_grouping_spec`, which a caller holds and `print()` renders, and `margin_grouping_plan`, which executors and adapters consume |
| smell | Minor indirection — the empty-grouping-set substitute calls the full preflight | rejected: the alternative writes the record's fields in a second place. One function answers what a record holds |

Finding 4's fix exposed a defect the narrower derivation had hidden: where a position nests a kind inside itself, one shadowed constructor intercepted the caller's own parent call as well as the nested one and counted both, so `forced` read 2. The parent is now namespace-qualified and the nested spelling stays bare, which is the spelling the gate reads. The comment at that site says so.

### Spec — 1 finding

| # | Finding | Disposition |
|---|---|---|
| 1 | Same as Standards 4, plus: the lazy-input block ran `bind_name` only, not both recognized forms | fixed in code — both forms, in the same loop |

No scope creep and no incorrect implementation found. The round verified independently that the counts are 1 in all four cases, that patching `expand_grouping_sets()` to re-read per pass restores 3 (so the assertions are load-bearing), that `grouping_arg_spec()` is down to one non-recursive call site from five, that both compilation passes receive the same `data_vars` — so freezing the preflight's reading is behaviourally equivalent and not merely count-equivalent — and that `dbplyr` is an Import rather than a member of `optional_backends()`.

It also read `ADR 0008`'s amendment table against #260's *Scope boundaries* ("this ticket does not restate" the colliding name's count) and found the two reconcile: *What to build* requires all four shapes stated in one place, and ADR 0026's own amendment hands that row over rather than keeping a second copy.

### After the answers

- full `testthat::test_local()` — pass
- `lintr::lint_package()` — no lints
- `Rscript .github/scripts/verify-suite-coverage.R` — 631 tests, all covered
- the new test still fails against the pre-change implementation (10 failures)

PR #310, opened 2026-08-29
Run an Arrow Dataset through the caused-read gate

### LINKED ISSUE #302: The caused-read gate never runs an Arrow Dataset, the one shape whose reads may leave this process
Split out of #300's verification, and a sibling of #301: the same sweep that
found #301's gap found a shape the caused-read gate never runs.

## What is unasserted

`test-query-policy.R`'s Arrow test walks four input shapes on the refusing
path — a `Table`, a `RecordBatch`, and a query over each — and asserts a zero
for every one. ADR 0020's amendment counts five Arrow shapes, and
`arrow_input_classes()` names `Dataset` alongside the rest. Neither a `Dataset`
nor a query over one is in that loop.

`test-grouping-backends.R` does cover them, in *an Arrow Dataset keeps Arrow's
own refusal*, but it asserts the **diagnostic**: that the caller gets
`arrow_not_supported` and not a `marginplyr_error`. Whether anything was read
on the way to that refusal is asserted nowhere.

## Why the four shapes do not cover the fifth

The zero is reached by a different mechanism, which is the whole reason the
existing loop gives its four shapes rather than one.

A `Table` or `RecordBatch` **absorbs**: Arrow warns and would read the whole
input, and marginplyr's handler intercepts that warning before the collect. The
zero those shapes assert is a statement about the handler firing in time.

A `Dataset` does not absorb. Arrow refuses the expression itself, so the handler
never fires and the zero comes from Arrow's own refusal happening before any
scan. That is a claim about Arrow's behaviour on a shape marginplyr does not
control, and it is exactly the kind of claim that stops being true quietly —
Arrow gaining a kernel, or falling back to a scan for an expression it currently
rejects, would turn this into a read of an input that may live in object
storage.

That last point is what makes this shape worth asserting rather than the
least. ADR 0020's amendment draws the `Table` / `Dataset` distinction precisely
because a `Table` is in this process's memory by definition and a `Dataset` may
not be; the reads that cost money are the ones on this shape.

## Measured

Both Dataset shapes report zero today, so this adds an assertion rather than a
fix:

| shape | summary | expand | refusing path |
|---|---|---|---|
| `InMemoryDataset` | 0 | 0 | 0 |
| query over one | 0 | 0 | 0 |

A `FileSystemDataset` written to `tempdir()` behaves the same.

## What it would take

`arrow::InMemoryDataset$create()` builds one from a `Table` with no file I/O,
and is already how `test-grouping-backends.R` and `test-get-col-names.R` reach
the class, so the addition costs no fixture and no new Suggest.

One thing needs deciding rather than doing. The existing loop's variable is
named `absorbing` and its comment reads "over every input class that absorbs
rather than over the one" — which is true and load-bearing, since what those
four shapes assert is that the handler fires. A `Dataset` does not absorb, so
appending it to that list would make the name and the comment false. It wants
its own loop asserting the same zero for the opposite reason, or a rename that
keeps both reasons legible.

sayuks:
Closes #302.

The caused-read gate's Arrow reading walked four input shapes on the refusing
path — a `Table`, a `RecordBatch`, and a query over each — and asserted a zero
for every one. All four absorb: Arrow warns and would collect the whole input,
and marginplyr's handler intercepts before the collect, so what those zeros
assert is that the handler fires in time.

The fifth shape `arrow_input_classes()` names, `Dataset`, and the query over
one were in no reading of it. `test-grouping-backends.R` covers them by
condition class and diagnostic; whether anything was read on the way to that
refusal was asserted nowhere.

## What this adds

A reading of its own rather than two more entries in the loop above, whose
name and comment say the shapes in it absorb. A `Dataset` does not: Arrow
refuses the expression at query-build time, the handler never fires, and what
holds the zero is Arrow refusing before it scans — a claim about the one shape
whose rows need not be in this process's memory, which is the distinction ADR
0020's amendment draws a `Table` against.

The drift it sees is the refusal moving to after a scan: same
`arrow_not_supported` class, same text, raised later. Both the refusal test
and the block asserting Arrow still absorbs pass unchanged through such a
move, the latter because it takes a `Table`.

A positive control comes with it, for the reason the readings beside it carry
one: both new expectations are zeros, and a counter blind to this class would
report what a shape that was not read reports. `dplyr::collect()` of the
dataset and of the query each count 1.

The comment states the bound the zero does not. Every reading is an R-level
entry point, so a C++-side scan that never pulls into R goes uncounted; what
makes the assertion worth having anyway is that Arrow's existing mechanism for
what it cannot translate is a fallback into R, so the likely form of the move
is counted.

No change to `R/` — both shapes report zero today, so this adds an assertion
rather than a fix.

## Local checks

- `testthat::test_local()` with `NOT_CRAN=true`: all files pass.
- `Rscript .github/scripts/verify-context-budget.R`: within budget (unchanged).
- `Rscript .github/scripts/verify-suite-coverage.R`: the local run was aborted
  part-way when its working directory was removed out from under it, so it
  gave no verdict. The `Suite coverage by construction` job on this PR is what
  answers it.

### DIFF
--- a/tests/testthat/test-query-policy.R
+++ b/tests/testthat/test-query-policy.R
@@ -386,7 +386,7 @@ test_that("no Arrow read happens while a Margin verb runs", {
 # query-build time, so marginplyr's handler never fires and what holds the zero
 # is Arrow refusing before it scans -- a claim about a shape whose rows need
 # not be in this process's memory, which is the distinction ADR 0020's
-# amendment draws a `Table` against and the shape where a read costs money.
+# amendment draws a `Table` against.
 #
 # The drift this sees is that refusal moving to after a scan: same
 # `arrow_not_supported` class, same text, raised later. The refusal test in
--- a/tests/testthat/test-query-policy.R
+++ b/tests/testthat/test-query-policy.R
@@ -378,6 +378,67 @@ test_that("no Arrow read happens while a Margin verb runs", {
   }
 })
 
+# The fifth Arrow shape and the query over it, which the loop above cannot
+# reach: a `Dataset` does not absorb, so appending it to a list named for
+# absorption would make that name and its comment false (#302).
+#
+# The zero is the same and the claim is not. Arrow refuses the expression at
+# query-build time, so marginplyr's handler never fires and what holds the zero
+# is Arrow refusing before it scans -- a claim about a shape whose rows need
+# not be in this process's memory, which is the distinction ADR 0020's
+# amendment draws a `Table` against and the shape where a read costs money.
+#
+# The drift this sees is that refusal moving to after a scan: same
+# `arrow_not_supported` class, same text, raised later. The refusal test in
+# `test-grouping-backends.R` asserts the condition and passes unchanged through
+# such a move, and so does the block asserting Arrow still absorbs, which takes
+# a `Table`. Nothing else in the suite is positioned to see it.
+#
+# The bound, which the zero does not state on its own: every reading here is an
+# R-level entry point, so a scan Arrow performs without pulling into R reaches
+# none of them and goes uncounted. What makes the assertion worth having anyway
+# is that Arrow's existing mechanism for evaluating what it cannot translate is
+# a fallback into R -- measured on arrow 25.0.1, an absorbed summary on a
+# `Table` counts 1 through it -- so a release extending that fallback to
+# Datasets, the likely form of the move, is counted.
+test_that("no Arrow Dataset read happens while a Margin verb runs", {
+  skip_if_suggest_absent("arrow")
+  data <- data.frame(
+    k = c("E", "E", "W"),
+    v = c(1, 2, 3),
+    s = c("a", "b", "c"),
+    stringsAsFactors = FALSE
+  )
+  # `InMemoryDataset` reaches the class with no file I/O. What a `Dataset` may
+  # be behind is not what is asserted here; that it is told apart from a
+  # `Table` is, and the class is what tells it apart.
+  dataset <- arrow::InMemoryDataset$create(arrow::Table$create(data))
+  query <- dplyr::select(dataset, dplyr::all_of(names(data)))
+
+  # The mechanism on this shape, asserted before anything is concluded from it,
+  # for the reason the readings above assert it on a `Table`: those controls
+  # were measured against a different class, and a counter blind to this one
+  # would report exactly what a shape that was not read reports.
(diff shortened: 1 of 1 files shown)


sayuks:

## Review record

with `main` (`git diff origin/main...HEAD`, merge-base `a222eeb`). The round
read `d51ac0f`; `272d7bb` on this branch is its answer.

Standards sources passed: `AGENTS.md`'s *Code comments*, *Release matrix*,
*Optional-dependency guards*, *Queries against a lazy input*, and *Linting*;
`design/architecture.md`'s *Answering a review*; ADR 0020. Spec source: #302
and its comments, the agent brief being the contract.

### Standards — 3 raised, 1 hard

| finding | disposition |
|---|---|
| **Hard.** The ADR 0020 citation carried a re-derivation of that ADR's cost reasoning beside it — `and the shape where a read costs money` — which *Code comments* refuses. The site is the exposed one: the amendment cited exists to withdraw a supporting claim about this same Arrow pair, and the clause is contradicted 24 lines later by the comment's own `What a Dataset may be behind is not what is asserted here`. | fixed in code, `272d7bb` — deleted, not restated |
| Duplicated Code: the fixture reproduces `refusing_arrow_inputs()` from `test-grouping-backends.R`. | ticket, #313 |
| Duplicated Code (weaker): the `data.frame(k, v, s)` fixture and the loop body repeat the block above. | ticket, #313 — the loop divergence is deliberate and stated at both sites, so #313 excludes it |

The two duplication findings are one subject and go to one ticket. The
duplication is forced today — those builders live inside a test file, which
testthat does not source into another's environment — and two of the three
copies pre-date this branch. #313 records what makes it worth fixing anyway:
this file's comments argue from what the other file asserts, so the copies
agreeing is load-bearing while nothing checks it.

Every factual claim the added comments make was verified on arrow 25.0.1,
including that a Dataset refuses in `arrow:::abandon_ship`'s
`query_on_dataset()` branch, before the `dplyr::collect()` fallback that the
comment's stated bound turns on.

### Spec — 0 raised

All five acceptance criteria met; nothing out-of-scope asserted; one file
changed.

The round did not stop at reading. It shimmed the registered
`summarise.Dataset` / `summarise.arrow_dplyr_query` methods to scan before
raising the same refusal — the exact drift the brief names, with no `R/` change
— and confirmed both new zeros fail under it while *an Arrow Dataset keeps
Arrow's own refusal* and the block asserting Arrow still absorbs both pass
unchanged. That is the branch's whole justification, checked rather than
argued. An `R/`-level mutation collecting the Dataset before summarizing also
failed both zeros.

Issue #181, opened 2026-08-14
An empty argument to a grouping helper is reported as a column named ``


sayuks:
## Parent

Part of #23. Found while implementing #174, by the audit of the empty-argument
invariant that ticket records. It is not #174 and not a regression from it: the
`across()` planning path #174 covers does not reach `grouping_helper_vars()`,
and this reproduces unchanged on `main` (`bda1d89`).

---

**File / line:** `R/grouping-context.R:206` (`grouping_helper_vars()`).

## What is wrong

R's empty argument is a symbol whose name is `""`, so it passes the bare
`is.symbol()` test that decides whether a `grouping_id()` or `grouping_bit()`
argument is a bare column, and `as.character()` then puts `""` into the column
vector. The diagnostic that follows names a column the caller never wrote, and
two empty arguments name the same one:

```r
d <- data.frame(
  region = c("East", "East", "West"),
  store = c("a", "b", "c"),
  value = c(1, 3, 6)
)

summarize_with_margins(
  d,
  b = grouping_id(region, ),
  .grouping = rollup(region, store)
)
#> Error: Column `` is not part of `.by` or `.grouping`.

summarize_with_margins(
  d,
  b = grouping_id(, ),
  .grouping = rollup(region, store)
)
#> Error: `grouping_id()` does not accept duplicate columns.
```

Both conditions are `marginplyr_error`, so this is not the untyped
`missingArgError` of #168 and #174 — the caller does get a Package condition.
What they do not get is one they can act on: nothing in either message names
the empty argument, and the second describes a duplicate that exists only
because both empty arguments were read as the same name.

The right answer is the one the same function already gives a non-column:

```r
summarize_with_margins(
  d,
  b = grouping_id(1),
  .grouping = rollup(region, store)
)
#> Error: `grouping_id()` only accepts bare grouping columns.
```

An empty argument is not a bare grouping column, so that is the message it
should reach, ahead of both the membership test and the duplicate test.

`grouping_bit()` is unaffected in its one-argument form only by accident: its
arity check runs first, so `grouping_bit(region, )` is refused for having two
columns rather than for the empty one, and `grouping_bit()` with a single empty
argument would take the same wrong path if the arity check ever moved.

## What to build

Read the arguments of a grouping helper by subscript and ask
`is_name_part()` (`R/utils.R`) whether each names a column, rather than
`is.symbol()`. That helper exists as of #174 and answers exactly this question:
a symbol, and not the empty argument.

## Acceptance criteria

[...]

PR #312, opened 2026-08-29
Refuse an empty argument in a nested specification position

### LINKED ISSUE #261: A leading or interior empty argument in a Grouping specification constructor reports an rlang argument name

sayuks:
Closes #261.

## What changed

A leading or interior empty argument to a Grouping specification constructor
reported an rlang argument name the caller never wrote:

```r
inspect_grouping(d, .grouping = grouping_sets(, region))
#> Error: argument "expr" is missing, with no default
```

`expr` is the local `grouping_arg_spec()` binds from
`rlang::quo_get_expr(arg)`; forcing it at the next read raises base R's untyped
`missingArgError`, and `conditionCall()` was `grouping_arg_spec(arg,
data_vars)`. It now reads:

```r
#> Error in `inspect_grouping()`:
#> ! Argument 1 of `grouping_sets()` is empty.
#> i Remove the comma, or write the columns that position selects.
```

## The disposition

Refusal, [recorded on the ticket](https://github.com/sayuks/marginplyr/issues/261#issuecomment-5460611345)
rather than the empty selection `dplyr::select()` reads the same shape as.
`select()`'s arguments are one selection between them, so an argument selecting
nothing is the identity and there is nothing further to decide. A
`grouping_sets()` argument is one grouping set, so "selects no column" does not
say whether the caller asked for the empty grouping set — which
`grouping_set()` already spells — or for no set at all. And the position
refuses this argument elsewhere already: `grouping_helper_vars()` refuses an
empty one ahead of its arity checks (#181), so that the diagnostic describes
what the caller wrote rather than whatever a later check reached first.

## Where the refusal sits

In `preflight_grouping_spec()`'s loop rather than in the reader below it,
because that is the frame holding what the diagnostic names. The argument has
no spelling for ADR 0024's rule to quote, so the constructor and the count of
arguments before it are the only things that point at the comma.

`rlang::quo_is_missing()` is asked of the quosure, before anything reads the
expression out of it, which is the rule `R/utils.R` states. That makes "`arg`
carries an expression" a contract `grouping_arg_spec()` holds rather than
re-establishes, so its local binding stands and gains a header saying who holds
it; `check_ambiguous_nested_name()`'s comment, which named this ticket as the
reason no empty argument reached it, is corrected to say what now keeps one
away.

## Tests

The constructors are derived from `grouping_kind_rules()`. Both positions
`rlang::enquos(...)` captures an empty argument for are covered, plus one
nested a level in, where the refusal names the inner constructor. The message
is asserted in full and `conditionCall()` is read, since a match on the part
that was already right could see neither defect.

A second test holds the four readings that must not move: a trailing comma, an
empty `.by`, an empty `.grouping`, and `grouping_set()` with no arguments at
all.

ADR 0008's test strategy asked for "syntactically empty arguments for every
kind" and that coverage was never written; this is it. No test added here
requires a member of `optional_backends()`.

## Documentation

`CONTEXT.md`'s *Nested specification position*, `?grouping_set`, and `NEWS.md`
each gain the rule in the paragraph already stating how a nested argument is
read.

## Acceptance criteria

- [x] A leading or interior empty argument in every nested position raises a
      `marginplyr_error` that names no rlang or marginplyr internal.
- [x] The constructors the assertion covers are derived from
      `grouping_kind_rules()`.
- [x] A trailing empty argument keeps the reading it has today.
- [x] An empty `.by` and an empty `.grouping` keep the readings they have today.
- [x] No test added by this ticket requires a member of `optional_backends()`.

## Local checks

`testthat::test_local()` under `NOT_CRAN=true`, `lintr::lint_package()`,
`spelling::spell_check_package()`, `roxygen2::roxygenise()`, and
`Rscript .github/scripts/verify-context-budget.R` all pass.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -24,12 +24,15 @@ is a column selection. Redundant parentheses are transparent to that reading,
 around the name or around the whole call, because `(` is the identity
 function. A spelling the position does not recognize is never
 evaluated to find out, because a selection such as `starts_with("re")` has no
-meaning outside a selection context. Both readings claim one argument at once:
-a bare name that is a column of the input and is also bound, in the caller's
-environment, to a specification of a kind the position admits — which kinds
-those are varies by constructor. How that name is written settles neither
-reading, so it is refused rather than resolved. ADR 0026 decides that, and
-holds the admitted kinds and what the read costs a caller.
+meaning outside a selection context. An argument the caller left empty gets
+neither reading, and is refused naming the constructor and the position; a
+trailing comma is not one, because R captures no argument for it. Both
+readings claim one argument at once: a bare name that is a column of the
+input and is also bound, in the caller's environment, to a specification of a
+kind the position admits — which kinds those are varies by constructor. How
+that name is written settles neither reading, so it is refused rather than
+resolved. ADR 0026 decides that, and holds the admitted kinds and what the
+read costs a caller.
 _Avoid_: Nested grouping, nested slot
 
 **Grouping plan**:
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,7 +19,12 @@
   parentheses are transparent to every one of these readings, as they are
   elsewhere, because `(` is the identity function: `(s)`, `(!!s)`, and
   `(rollup(region))` are the arguments they wrap, however many pairs deep, and
-  both refusals reach a parenthesized argument as they reach a bare one.
+  both refusals reach a parenthesized argument as they reach a bare one. An
+  argument you left empty gets neither reading, and is refused naming the
+  constructor and the position — `grouping_sets(, region)` reports that its
+  first argument is empty — rather than reporting an internal name you never
+  wrote. A trailing comma is not an empty argument, because R captures no
+  argument for it.
 * Added contextual `grouping_bit()` and `grouping_id()` summary helpers.
 * Added the contextual `share_of_parent()` summary helper, which divides a
   preceding numeric scalar summary by the same measure one `rollup()` level
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -225,6 +225,17 @@ abort_empty_grouping_units <- function(kind) {
   abort_marginplyr("{.fun {kind}} requires at least one dimension.")
 }
 
+# The refusal of an argument the caller left empty. It names the position
+# because that is all there is to name: the argument has no spelling for
+# ADR 0024's rule to quote, so the count of arguments before it is the only
+# thing that points at the comma the caller wrote (#261).
+abort_empty_grouping_arg <- function(constructor, position) {
+  abort_marginplyr(c(
+    "Argument {position} of {.fun {constructor}} is empty.",
+    i = "Remove the comma, or write the columns that position selects."
+  ))
+}
+
 abort_empty_composite <- function() {
   abort_marginplyr(
     "An empty {.fun grouping_set} cannot be a composite dimension."
@@ -356,6 +367,14 @@ preflight_grouping_spec <- function(grouping_spec, data_vars) {
   args <- vector("list", length(grouping_spec$args))
   for (i in seq_along(grouping_spec$args)) {
     arg <- grouping_spec$args[[i]]
+    # Asked of the quosure, which is an object like any other, and before
+    # anything reads the expression out of it -- the rule `R/utils.R` states
+    # for every reader a walk asks first (#168, #174). The refusal is here
+    # rather than in the reader below because this is the frame holding what
+    # the diagnostic names: the constructor and the argument's position (#261).
+    if (rlang::quo_is_missing(arg)) {
+      abort_empty_grouping_arg(rule$constructor, i)
+    }
     nested <- grouping_arg_spec(arg, data_vars)
     if (is.null(nested)) {
       check_ambiguous_nested_name(arg, grouping_spec, rule, data_vars)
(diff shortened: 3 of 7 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, one round, over `git diff a222eeb...ce87a43`
(merge-base with `main`). Two axes, parallel sub-agents. Answered in e50201a.

Five findings.

### Spec — 1

| finding | disposition |
|---|---|
| An empty argument written inside redundant parentheses escaped the refusal and raised `attempt to use zero-length variable name` from `rlang::env_has(env, name, inherit = TRUE)` — an untyped condition naming an rlang internal, which is what AC 1 forbids. | **fixed in code.** `wraps_empty_argument()` in `R/utils.R`, read off `nested_arg_expr()`'s answer. `test-grouping-plan.R`'s refusal test gains one pair and two, over every constructor and both positions, and fails without the fix. |

Reachable only through an injection or a constructed call — the parser rejects
`f((), x)` — but that is the footing every other reading of this position
reaches a parenthesized argument on (#178, #259), and the branch's own
`CONTEXT.md` and `NEWS.md` edits sit inside the paragraph asserting that
transparency.

The axis raised nothing under scope creep: all three *Scope boundaries* items
are untouched. It confirmed the refusal fires through `!!`, `call2`, a user
function's dots, and `summarize_with_margins()`, and that the four unmoved
readings hold.

### Standards — 4

| finding | disposition |
|---|---|
| `nested_arg_expr()`'s header still claimed an argument handed to it may be one a caller left empty. Every caller now runs downstream of the refusal. | **fixed in prose.** The false claim is deleted, not restated; what replaces it is what the refusal actually reads. |
| `abort_empty_grouping_arg()`'s header gave a rationale where its contract belongs, and the rationale was wrong — "the count of arguments before it" describes `i - 1`, the message reports `i`. | **fixed in prose.** It now says what `position` is and what makes it the number the caller typed. |
| `grouping_arg_spec()`'s header cited a comment whose opening sentence reads as the opposite of the property relied on. | **fixed in prose.** It names the property: `is_parenthesized()` answers `FALSE` for the pair that would hold an empty argument. |
| A test name spelled the position "slot", which `CONTEXT.md` lists under `_Avoid_`. | **fixed in prose.** |

The axis reported the diagnostic clean against ADR 0023 and ADR 0015, the tests
clean against the derive-don't-list rule and the one-backend policy, and no
baseline smell worth reporting.

PR #314, opened 2026-08-29
Refuse a nested specification stored as a function

### LINKED ISSUE #265: A specification stored as a function is read by tidyselect as a predicate, so #190's refusal never fires
## Problem

#190 gave a nested position its own diagnostic for a specification that arrives where the spelling gate expects a column selection: the position's contract, plus the binding that makes the call work. It is derived from tidyselect's own refusal — `is_grouping_spec_subscript()` reads the refused subscript out of the condition and recognizes a `margin_grouping_spec`.

A function carrying the class never produces that refusal, because tidyselect does not refuse a function: it calls it, as the predicate form of a selection. The caller receives base R's untyped complaint about the call tidyselect made.

```r
d <- data.frame(a = c("x", "y"), value = c(1, 2))

# recognized, refused in marginplyr's words
my_spec <- function() rollup(a)
inspect_grouping(d, .grouping = grouping_sets(my_spec()))
#> Error: `my_spec()` is a grouping specification, but a nested position
#>   recognizes one only when it is a call to `grouping_set()`, ... [...]
#>   class: marginplyr_error/rlang_error/error/condition

# the same position, the same class, stored as a function
f <- function() structure(function() 1, class = "margin_grouping_spec")
inspect_grouping(d, .grouping = grouping_sets(f()))
#> Error: unused argument (X[[i]])
#>   class: simpleError/error/condition
```

`X[[i]]` names nothing the caller wrote, and the object it complains about carries the class the position exists to recognize — the self-contradicting diagnostic #190 was filed about, in a form that ticket's fix does not reach.

## Expected behavior

Per ADR 0015 and the *Errors and warnings* section of `?marginplyr`, an error a caller can avoid by rewriting their call is a Package condition. A nested specification is exactly that: the rewrite is the one #190's refusal already names. Whatever the object is stored as, a nested position that received a `margin_grouping_spec` should say so in marginplyr's words.

## Reproduction / evidence

Measured on R 4.6.1 at `112c12c`. Pre-existing and unrelated to #262: an object stored as an atomic vector takes #190's refusal in the same position, so what varies is only whether tidyselect refuses the object or calls it.

## Scope boundaries

- The wording of #190's refusal, and the wording for every object that reaches it today.
- The spelling gate itself. `grouping_arg_spec()` decides by how the argument is written (ADR 0019), and a specification stored as a function is refused here for the same reason any other unrecognized spelling is. What is wrong is the diagnostic, not the refusal.
- ADR 0008 fixes how often a caller's quosure is evaluated, and #190's refusal reads the value out of tidyselect's condition rather than evaluating again for exactly that reason. A fix that evaluates the argument to find out whether it is a function would move that count.

## Notes for whoever picks this up

The obstacle is that tidyselect raises nothing to inspect — it succeeds at calling the function and fails inside the call, so the condition carries no refused subscript. Reading the value again is what ADR 0008 forbids, so the answer is likelier to be in what `resolve_grouping_selection()` can learn from the condition it did receive than in a second read.

## Origin

Found by the Spec review of #262's branch, which flagged it as adjacent and pre-existing rather than that ticket's to answer.

sayuks:
Closes #265.

## What was wrong

#190's refusal is written from the condition tidyselect raises about a subscript it will not use. tidyselect raises no such condition about a function: `is.function()` decides its first branch, before any class is read, so it calls the value as the predicate form of a selection. A `margin_grouping_spec` stored as a function therefore reached a nested position carrying the class that position exists to recognize, and the caller received a diagnostic naming nothing they wrote:

| Stored as | Before |
|---|---|
| `function() 1` | `simpleError`: `unused argument (X[[i]])` |
| `function(x) stop("inside")` | whatever the body raised |
| `function(x) "no"` | `Predicate must return TRUE or FALSE, not a string.` |

All three now get #190's refusal, unchanged in wording, with the `marginplyr_error` class ADR 0015 assigns it.

## How

The object is read from the frames tidyselect applied it from, as the error unwinds, and the argument's own expression decides whether the position speaks for it. Neither is a second reading of the caller's quosure — ADR 0008's evaluation-count table is unchanged, and a binding tidyselect has not forced is skipped for that reason. The frames are tidyselect's by the namespace their function closes over, never by a private name, so a release that stops opening them leaves the caller the untyped diagnostic they had rather than raising.

The rule that a position speaks for its own argument and not for a part of one is kept, and drawn differently because it has to be: a condition about a predicate names no subscript to compare a label with. tidyselect walks `c`, `-`, `:`, `!`, `&`, `|`, and `/` in parts, so an argument written with one of those keeps tidyselect's own report, exactly as a specification refused inside `c()` does.

## Not reached

A function that is a valid predicate as well as a specification — one argument, a logical scalar back — is applied without failing, so no condition exists to read the frames from and the position receives a selection. Nothing a constructor builds has that shape. ADR 0008's amendment records it rather than leaving it implied.

## Records

- `design/adr/0008-…`: the amendment for the condition class, its sixth.
- `investigation/reading-a-specification-tidyselect-called.md`: what was measured of tidyselect 1.2.1 — where the value is readable, the four shapes, and why `allow_predicates = FALSE` answers neither half of the question.
- `NEWS.md`: the #190 bullet now covers every way of storing the object.

## Checks

`testthat::test_local()` (632 tests, all pass), `lintr::lint_package()` (no lints), `Rscript .github/scripts/verify-suite-coverage.R`, `Rscript .github/scripts/verify-context-budget.R`, `roxygen2::roxygenise()` (no diff). The new test fails in 8 assertions without the change.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,10 @@
   specification. That refusal is now marginplyr's own diagnostic, naming the
   recognized forms and the binding that works — `s <- my_spec(region)`, then
   `grouping_sets(s, grade)` — rather than tidyselect reporting the
-  specification as an unusable column selection. A name both readings claim —
+  specification as an unusable column selection. It is the diagnostic whatever
+  the specification is stored as, including a function, which tidyselect calls
+  as a column predicate rather than refusing: that case reported an error about
+  a call you never wrote (#265). A name both readings claim —
   a column of your input that is also bound to a nested specification the
   position accepts — is refused rather than resolved by whichever the data
   happens to have, and the refusal names the spelling for each reading:
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,10 +11,10 @@
   specification. That refusal is now marginplyr's own diagnostic, naming the
   recognized forms and the binding that works — `s <- my_spec(region)`, then
   `grouping_sets(s, grade)` — rather than tidyselect reporting the
-  specification as an unusable column selection. It is the diagnostic whatever
-  the specification is stored as, including a function, which tidyselect calls
-  as a column predicate rather than refusing: that case reported an error about
-  a call you never wrote (#265). A name both readings claim —
+  specification as an unusable column selection. It is also the diagnostic for
+  a specification stored as a function, which tidyselect calls as a column
+  predicate rather than refusing, so the call reported an error about a call
+  you never wrote (#265). A name both readings claim —
   a column of your input that is also bound to a nested specification the
   position accepts — is refused rather than resolved by whichever the data
   happens to have, and the refusal names the spelling for each reading:
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -990,21 +990,136 @@ grouping_arg_spec <- function(arg, data_vars) {
 # `abort_share_source_name()` does for the same reason. Every other failure is
 # an External condition and is re-raised as it arrived, with its own class,
 # diagnostic, and cause.
+#
+# A specification stored as a function is refused from the frames instead,
+# because tidyselect refuses no function: it calls one, as the predicate form
+# of a selection, so the caller was left with base R's untyped complaint about
+# the call tidyselect made and no condition named the object (#265). The two
+# readings answer one question and reach one refusal. Which of them is
+# available is decided by how the object is stored and not by what it is, which
+# is why neither is written as the other's fallback.
+#
+# The frames are read as the error unwinds and the condition after it, because
+# a predicate is applied inside `resolve_column_selection()` and the frames
+# that applied it are gone once it returns.
 resolve_grouping_selection <- function(arg, data_proxy) {
+  selection <- sys.nframe()
+  predicate <- FALSE
   tryCatch(
-    resolve_column_selection(
-      arg,
-      data_proxy,
-      on_rename = abort_grouping_rename
+    withCallingHandlers(
+      resolve_column_selection(
+        arg,
+        data_proxy,
+        on_rename = abort_grouping_rename
+      ),
+      error = function(cnd) {
+        if (is_grouping_spec_predicate(selection, nested_arg_expr(arg))) {
+          predicate <<- TRUE
+        }
+      }
     ),
     error = function(cnd) {
       label <- rlang::as_label(nested_arg_expr(arg))
-      if (!is_grouping_spec_subscript(cnd, label)) {
+      if (!predicate && !is_grouping_spec_subscript(cnd, label)) {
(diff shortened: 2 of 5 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` against the merge-base with `main` (`a222eeb`), on `4617dcf`. Standards and Spec ran as parallel sub-agents. Standards sources: `AGENTS.md`, `design/architecture.md`, ADRs 0008/0015/0019/0023/0024/0026, `design/agents/code-review.md`, `investigation/README.md`. Spec: #265.

Answered in `431f1b5`.

### Standards — 4 findings

| Finding | Disposition |
|---|---|
| The refusal speaks for a part of an argument: `is_grouping_spec_predicate()` declines only tidyselect's walk operators, but the frame scan answers for any tidyselect frame, so `all_of(f())` and `starts_with(f())` were refused (hard — breaches `design/architecture.md` and the branch's own ADR amendment) | fixed in code; the test that fails without it covers the five helper shapes |
| Prose false: the header and the ADR amendment assert a narrower scan than the code performed | fixed in prose, in the header, the amendment, and the note |
| Test gap: the part rule was exercised only for `c()` and `-`, both on the operator list | fixed in code |
| Mysterious Name: `selection` holds a frame index | fixed in code — `selection_frame` |
| Latching flag: `predicate <<- TRUE` set by a handler that fires for every signalled error | fixed in code — the handler keeps the condition, and the refusal requires it to be the one that escaped or one wrapping it |

### Spec — 3 findings

| Finding | Disposition |
|---|---|
| (c) Over-refusal beyond the predicate path — the same finding as the first Standards one, measured over `starts_with`, `last_col`, `all_of`, `any_of`, and the unchanged list form | fixed in code |
| (a) Partial: a specification stored as a valid predicate still returns silently, and `NEWS.md` overclaimed | ADR 0008's amendment already records the shape under *What is left*; the `NEWS.md` claim is deleted rather than restated |
| (b) Scope creep: none | — |

Confirmed unaffected by the Spec axis, with an instrumented caller: the quosure evaluation count for every row of ADR 0008's table, #190's wording, `grouping_arg_spec()`, `where()` predicates, `c()`/`-`/`:` parts, injected quosures, `((f()))`, and the empty argument.

### The finding that mattered

`eval_select()` is exported and is the entry, so every frame is under an exported one; excluding exported frames does not reach `all_of()`, which delegates to an internal `as_indices_impl()` that binds the specification. The first exported function reached *after* the entry is a helper the caller wrote, and everything deeper is that helper's — the scan stops there. Measured for all five helper arguments and for the three shapes the ticket is about.

PR #315, opened 2026-08-29
Build the Arrow input shapes once

### LINKED ISSUE #313: The Arrow input shapes are built twice, in two test files
Raised as a judgement call by the Standards round on #310's branch (PR #310),
and dispositioned there as a ticket because its subject is where the suite's
Arrow fixtures live rather than anything that branch asserts.

## What is duplicated

`tests/testthat/test-grouping-backends.R` defines three fixture builders:
`absorbed_summary_data()`, `absorbing_arrow_inputs()`, and
`refusing_arrow_inputs()`. `tests/testthat/test-query-policy.R` reconstructs
all three inline — the same `data.frame(k, v, s)`, the same
`Table$create()` / `record_batch()` / `dplyr::select(..., all_of(names(data)))`
list, and the same `InMemoryDataset$create(Table$create(data))` pair.

The duplication is forced as things stand: those builders are defined inside a
test file, and testthat evaluates each test file in its own environment, so
nothing in `test-query-policy.R` can reach them. Only a `helper-*.R` file is
sourced into the shared environment.

Two of the three copies pre-date #310 — the absorbing half and the data frame
are on `main` — and #310 added the third for the shape it introduced, matching
what the file already did rather than diverging from it.

## Why it is worth a ticket rather than nothing

The copies are not required to agree, so nothing fails when they stop agreeing.
What makes drift visible instead of silent is that `test-query-policy.R`'s
comments argue from what `test-grouping-backends.R` asserts: its Dataset block
says the refusal test there "asserts the condition and passes unchanged through
such a move", and the absorbing block cites the drift block beside it. Those are
claims about another file's coverage, held true today by two fixtures that
happen to build the same shapes.

## What a fix looks like

Move all three builders to a `helper-*.R` file and call them from both test
files. Moving only `refusing_arrow_inputs()` is worse than leaving it: the file
would then share one half of a pair and inline the other, for no reason a reader
could recover.

Worth settling in the same change: whether `test-query-policy.R`'s
`data.frame(k, v, s)` becomes one file-scope fixture, which would remove the
third copy but edits a block the ticket is not otherwise touching.

## Out of scope

The `absorbing` and `refusing` loops themselves. They read alike and assert
zeros that hold for opposite reasons, and the comments above each say which —
that divergence is deliberate.

sayuks:
Closes #313.

`test-grouping-backends.R` defined `absorbed_summary_data()`,
`absorbing_arrow_inputs()`, and `refusing_arrow_inputs()` inside the test file.
testthat evaluates each test file in its own environment, so
`test-query-policy.R` could not reach them and rebuilt all three inline. The
copies were not required to agree, while `test-query-policy.R`'s comments argue
from what `test-grouping-backends.R` asserts — its Dataset block says the
refusal test there "asserts the condition and passes unchanged through such a
move".

## What changed

- `tests/testthat/helper-arrow-shapes.R` is new and holds the three builders.
  Its header states why they are in a helper and what the two suites do with
  them; what each shape *means* stays where it is asserted.
- `test-grouping-backends.R` drops the definitions and keeps the block arguing
  the expression choices, with a pointer to the helper.
- `test-query-policy.R` calls the three builders. Its two
  `data.frame(k, v, s)` literals become `absorbed_summary_data()`, which
  removes the third copy the ticket left open as a question.

The `absorbing` and `refusing` loops are untouched, per the ticket's Out of
scope.

## Checks

`testthat::test_local()` with `NOT_CRAN` set, `lintr::lint_package()` after
`pkgload::load_all()`, and `verify-context-budget.R` all pass. No `R/` change,
so nothing generated needed regenerating.

### DIFF
--- a/tests/testthat/helper-arrow-shapes.R
+++ b/tests/testthat/helper-arrow-shapes.R
@@ -1,23 +1,17 @@
-# The Arrow input shapes two suites take, in a helper file because testthat
-# evaluates each test file in its own environment and only a `helper-*.R` file
-# is sourced into the shared one, so a second suite needing a shape had to
-# rebuild it (#313).
+# The Arrow input shapes `test-grouping-backends.R` and `test-query-policy.R`
+# both take. testthat evaluates each test file in its own environment and only
+# a `helper-*.R` file is sourced into the shared one, so this is where a shape
+# two suites need is built once (#313).
 #
-# `test-grouping-backends.R` asserts which condition each shape raises, and
-# `test-query-policy.R` counts the reads each one causes. The second file's
-# comments argue from what the first asserts -- its Dataset block says the
-# refusal test there "asserts the condition and passes unchanged through such a
-# move" -- and that claim holds only while the two build the same shapes. Two
-# copies were not required to agree, so nothing failed when they stopped.
-#
-# What each shape *means* is stated where it is asserted, not here: this file
-# builds the objects and the division between them is CONTEXT.md's *Absorbing
-# backend*, at the seam ADR 0025 draws it at.
+# The two suites assert different things over the same objects:
+# `test-grouping-backends.R` which condition each shape raises, and
+# `test-query-policy.R` how many reads each one causes. The second file's
+# comments cite the first file's coverage, and that citation holds while both
+# build their shapes here.
 
-# A group, a numeric column, and a string column: the columns the absorbed
-# expressions asserted over need, which is a group collapsed to one string, a
-# subset inside an aggregate, and a statistic over two columns.
-absorbed_summary_data <- function() {
+# A group column, a numeric column, and a string column: what the summary
+# expressions written over this input in both suites read.
+arrow_input_data <- function() {
   data.frame(
     k = c("E", "E", "W"),
     v = c(1, 2, 3),
--- a/tests/testthat/helper-arrow-shapes.R
+++ b/tests/testthat/helper-arrow-shapes.R
@@ -1,0 +1,52 @@
+# The Arrow input shapes two suites take, in a helper file because testthat
+# evaluates each test file in its own environment and only a `helper-*.R` file
+# is sourced into the shared one, so a second suite needing a shape had to
+# rebuild it (#313).
+#
+# `test-grouping-backends.R` asserts which condition each shape raises, and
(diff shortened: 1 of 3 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, both axes, against the merge-base with `main`
(`origin/main` = cf24278; the primary checkout's local `main` was stale at
a222eeb, which would have reported #310's diff as this branch's). Subject:
c35d1b5. Answered in 9776b84.

Standards raised three findings and one judgement call; Spec raised one under
scope creep and one under "implemented but wrong". The first Standards finding
and the Spec scope-creep finding are the same defect from two directions.

| # | Finding | Disposition |
|---|---|---|
| 1 | The claim that both suites build their shapes in one file stood three times — helper header, and a paragraph in each test file. *Answering a review* kind 2: nothing compares the copies. | Fixed in prose. The argument keeps one home, the helper header; each test file carries a bare pointer. |
| 2 | The helper header narrated the defect in the past tense — "had to rebuild", "were not required to agree, so nothing failed when they stopped". `AGENTS.md` *Code comments* puts that in the commit message. Spec found the second clause false as well: `main`'s two copies built identical objects, so no divergence ever happened. | Fixed in prose. Deleted rather than restated. |
| 3 | The header said "What each shape *means* is stated where it is asserted, not here" and then restated the three expression shapes `test-grouping-backends.R` owns. Kind 1 and kind 2. It also quoted a sentence from `test-query-policy.R` verbatim, which nothing keeps in step with its source. | Fixed in prose. The fixture header states its columns and stops; the quote is gone. |
| 4 | Mysterious Name (judgement call): `refusing_arrow_inputs(absorbed_summary_data())` builds the refusing shapes from a builder named for absorption — the objection #302 raised against a container named for absorption holding a `Dataset`. | Fixed in code: renamed to `arrow_input_data()`, 9 call sites. |

Spec found no missing or partial requirement, and confirmed the **Out of scope**
section is untouched: only the list-construction lines were lifted from the
`absorbing` and `refusing` loops, and both divergent comment blocks are
byte-identical to `main`.

Checks after the answer: `testthat::test_local()` with `NOT_CRAN` set (whole
suite, arrow present, no failures), `lintr::lint_package()` after
`pkgload::load_all()` (clean), `verify-context-budget.R` (within baseline).

PR #316, opened 2026-08-29
Classify a grouping kind with its class off

### LINKED ISSUE #289: Strip a grouping kind's class before classifying it, rather than catching what its methods raise
## Problem

`grouping_kind_name()` classifies a kind by asking the value's own `is.na()`
and `length()` methods inside a `tryCatch()`, then classifies the answer a
second time with the class stripped:

```r
name <- tryCatch(
  if (is.character(kind) && length(kind) == 1L && !is.na(kind)) unclass(kind),
  error = function(cnd) NULL
)
if (!is.character(name) || length(name) != 1L || is.na(name)) return(NULL)
name
```

Stripping the class *first* makes the whole question total by construction,
with no catch and no second classification:

```r
if (!is.character(kind)) return(NULL)
name <- unclass(kind)
if (length(name) != 1L || is.na(name)) return(NULL)
name
```

Neither call the shorter form makes can be intercepted. `is.character()` and
`unclass()` are primitives that dispatch on neither S3 nor S4, so nothing a
forged kind can define is reached, and `length()` and `is.na()` are then put to
a plain character vector.

## Evidence

Measured on R 4.6.1 at `457f227`:

```r
setClass("S4hostile", contains = "character")
setMethod("is.character", "S4hostile", function(x, ...) stop("raises"))
#> Error: must supply a function skeleton for 'is.character'
setMethod("unclass", "S4hostile", function(x, ...) stop("raises"))
#> Error: must supply a function skeleton for 'unclass'
```

S3 registration into `base` is ignored the same way: with
`registerS3method("is.character", "ic", function(x, ...) stop("raises"), envir
= asNamespace("base"))` in place, `is.character(structure("a", class = "ic"))`
answers `TRUE`. `unclass()` was checked against every shape `is.character()`
answers `TRUE` for — a classed character, an S4 class extending `character`,
one extending it with additional slots, and a character matrix — and raised for
none.

What that buys is not only brevity. It removes a `tryCatch(error = )`, and with
it the trade ADR 0008 records under *the handler the read is wrapped in*: an
exiting handler unwinds a condition that a calling handler of the caller's
would otherwise have resumed from, and a `signalCondition()` that inherits
`error` without stopping is caught rather than resumed.

## Why it was not done in #280

It reverses a decision, and #280's acceptance does not reach it.

ADR 0008's amendment *for a kind that is no name* (#268) decides that the
printed line asks the value's own methods and reports the empty name when one
raises. Classifying without ever asking them means a kind of
`structure("set", class = "raises")` prints `grouping_set` instead — which is
arguably the better answer, since the kind *is* `set` and the class merely
carries a method that was never relevant. But it is a different answer from the
one recorded.

Substituting the shorter form for `grouping_kind_name()`'s body fails seven
assertions: `test-grouping-interface.R:1255` and `:1282`, which are #268's, and
`test-grouping-plan.R:452`, `:1064`, `:1065`, `:1066`, which are #280's.

## Expected outcome

A decision on whether the printed line should ask a kind's own methods at all,
and if not:

- `grouping_kind_name()` strips before classifying, and loses its catch and its
  second classification;
- ADR 0008 records the reversal, naming what moves — the printed line for a
  kind whose classification would have raised, and the handler trade the
  amendment states;
- the tests above are updated to the decided answer rather than deleted, since
  what they pin is which of two readings the printer takes.

If the answer is that the methods should be asked, this issue is the record of
why, and ADR 0008 gains a *Considered options* entry pointing here.

## Origin

Weighed while answering the Standards axis finding on #280's branch, where the
shorter form was rejected for reaching outside that ticket's acceptance. The
rejection was recorded in `design/review-dispositions.md` §12, which #288
proposes retiring; this issue is where the alternative lives once it does.

sayuks:
Closes #289.

## What changes

`grouping_kind_name()` strips a kind's class before classifying it rather than
after, which makes the question total by construction and removes both the
`tryCatch(error = )` and the second classification:

```r
if (!is.character(kind)) return(NULL)
name <- unclass(kind)
if (length(name) != 1L || is.na(name)) return(NULL)
name
```

Neither call the first pair makes can be intercepted. `is.character()` and
`unclass()` are primitives that dispatch on neither S3 nor S4, so `setMethod()`
refuses a method for either, and a `registerS3method()` into `base` is ignored.
An S4 object extending `character` unclasses to a character whose `class()` is
`"character"`, so an S4 `length()` or `is.na()` method it carried is not found
on the answer either. Measured on R 4.6.1.

`unclass()` takes the class off and leaves every other attribute, so the answer
is one character element with no class rather than a plain string. `%in%` and
`[[` are what the callers put it to, and neither reads an attribute.

## What that reverses

ADR 0008's amendments *for a kind that is no name* (#268) and *for a kind the
guard could not classify* (#280) both decide what a reader does when
classifying raises. This decides that classifying does not raise, and the ADR
gains an amendment recording the reversal and naming what moves:

- the printed line names the constructor of the kind underneath, where it
  printed the empty name;
- the guards in `validate_grouping_spec_early()` admit such a kind, where they
  refused it;
- the nested-name site refuses it as ambiguous, where a raising `is.na()` or
  `length()` left the column reading standing.

`structure("set", class = "raises")` is a kind that spells `set`; the class
carries a method that was never relevant to it.

## Tests

The four tests the reversal moves are updated to the decided answer rather than
deleted. The printed line's warning half is retargeted from a classification
method to the field read, which is where the catch it constrains still is.
`helper-forged-kinds.R`'s headers say what the fixture is for now that no
reader reaches a method it registers. No test added or changed here requires an
optional backend.

## What this branch does not change

`compile_grouping_spec_impl()` records the kind field as it was read. Storing
the classified name there was tried and reverted in the review: it is not a
consequence of this decision, and it does not close the defect it claimed to.
#317 carries that defect and the two decisions it needs.

## Checks

Full `testthat::test_local()` with `NOT_CRAN=true`, `lintr::lint_package()`,
`verify-context-budget.R`, and `verify-suite-coverage.R`, all green. No
roxygen, `README.Rmd`, or vignette source changed, so nothing generated moves.

### DIFF
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -714,11 +714,7 @@ compile_grouping_spec_impl <- function(preflight,
 
   structure(
     list(
-      # The classified name rather than the field, so that what a reader of the
-      # plan compares is the name the guard admitted. `share.R` asks
-      # `identical(plan$kind, "rollup")`, which a kind carrying a class answers
-      # `FALSE` to while compiling as one (#289).
-      kind = grouping_kind_name(preflight$spec$type),
+      kind = preflight$spec$type,
       by = unique(.by),
       dimensions = dimensions,
       sets = normalized,
@@ -903,18 +899,21 @@ grouping_kind_rules <- local({
   }
 })
 
-# The name a kind read off an object is, or `NULL` where it is none: a name is
-# one string, and not the missing one. Every caller holds a value nothing has
-# validated, and every reader of one shares this, as they share the read in
-# `grouping_spec_kind()`. What it decides, why it answers with the name rather
-# than with whether there is one, and why nothing here is asked of the object's
-# own methods are ADR 0008's, in its amendment for a kind classified with its
-# class off.
+# The name a kind read off an object is, or `NULL` where it is none: one
+# character element with the class off, and not the missing one. Every caller
+# holds a value nothing has validated, and every reader of one shares this, as
+# they share the read in `grouping_spec_kind()`. What it decides, why it answers
+# with the name rather than with whether there is one, and why nothing here is
+# asked of the object's own methods are ADR 0008's, in its amendment for a kind
+# classified with its class off.
 grouping_kind_name <- function(kind) {
   # `is.character()` and `unclass()` are primitives that dispatch on neither S3
   # nor S4, so neither reaches a method the value carries: the type test is R's
-  # own, and stripping is what puts the two questions below to a plain character
-  # vector rather than to whatever the class defines for them (#289).
+  # own, and stripping is what puts the two questions below to a character
+  # vector with no class rather than to whatever the class defines for them
+  # (#289). Only the class comes off, so a name and any other attribute the
+  # value carried survive onto the answer; `%in%` and `[[` are what the callers
+  # put it to, and neither reads one.
   if (!is.character(kind)) {
     return(NULL)
   }
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -10,7 +10,7 @@ validate_grouping_spec_early <- function(grouping_spec) {
   }
 
   # Classified where it is read, so that what this guard holds from here on is
-  # a name and not a value that might still refuse to be one (#280).
+  # a name and not the value it was read off (#280).
   kind <- grouping_kind_name(grouping_spec_kind(grouping_spec))
   # A catch of its own, because the two fields are not read together: an object
   # can refuse one while answering the other, and this is the only reader of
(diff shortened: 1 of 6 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel, run against the
merge-base with `main`. The round read `a94bd16`; the branch was rebased onto
`edf3262` mid-round and the reviewed content landed unchanged as `d67eec0`,
which the Standards axis re-pinned itself. Answers are in `5eb6569`.

Standards sources passed: `AGENTS.md` (*Code comments*, *Linting*, *Release
matrix*, *Optional-dependency guards*, *Queries against a lazy input*),
`design/architecture.md` (*Answering a review*, *Structural gates*), and ADRs
0008, 0015, 0026. Spec source: #289.

### Standards — 4 findings

| # | Finding | Disposition |
|---|---|---|
| 1 | `tests/testthat/helper-forged-kinds.R:1-7` and `:34-35`, untouched by the branch and falsified by it: the headers say the fixture exists so that classifying reaches a method of the object's own, and name a refusal, a declined reading, and an empty printed name as the classification's answer — the three outcomes the branch's tests now assert do not happen. Hard: *Code comments* requires a claim found false to be deleted. | Fixed in prose, `5eb6569` |
| 2 | `R/grouping-plan.R:717-720` and the ADR's *The Grouping plan's kind*: the stated motive does not hold, with a counterexample where a plan that compiled as a rollup is still refused by `share_of_parent()`. | Fixed in code — reverted, and #317 filed |
| 3 | "primitives that dispatch on neither S3 nor S4" derived in four places; the two test comments restate the mechanism the ADR owns and can drift from it. | Fixed in prose for `test-grouping-interface.R`; rejected for `test-grouping-plan.R`, whose comment states which three methods that site ever asked, which is the site's fact and not the ADR's |
| 4 | Duplicated Code, minor: the `warns` active-binding fixture is the seventh inline copy of the shape, and `helper-forged-kinds.R` is where it would live. | Rejected — it is the idiom the neighbouring tests in that file already use inline, and hoisting one of seven copies is a change to the other six that this branch's Acceptance does not reach |

The axis verified the branch's empirical claims independently on R 4.6.1 and
found the ADR and commit prose accurate on all of them.

### Spec — 3 findings

| # | Finding | Disposition |
|---|---|---|
| 1 | Partial: the ADR says `length()` and `is.na()` "are put to a plain character vector" and the header promises "a name is one string", but `unclass()` removes only the class — a named character, or an S4 kind with a slot, classifies to an answer carrying attributes. | Fixed in prose, in the header, the inline comment, and the ADR |
| 2 | Scope creep: `compile_grouping_spec_impl()`'s `kind`. Not a consequence of the decision — a benign classed kind already compiled on `main` and already put its class in the plan — with an unreachable path named as its motive and incomplete as a fix. Same category as the rejection #289's *Origin* records. | Fixed in code — reverted; #317 |
| 3 | The `#289` comment and the ADR paragraph assert behaviour the package does not execute. | Same as 2 |

Its check of the branch's central claim holds: at all four readers of an
unvalidated kind, no `is.na()`, `length()`, or `as.character()` reaches the
classed value, and `%in%` and `[[` see the stripped one.

### Evidence the answers landed with

- Finding 1 (both axes' code finding): reverting is what makes `share.R`'s
  question no worse than it was; #317 carries the reproduction, measured on
  this branch and on `main`, and the guard test cites it where it sets the
  plan's `kind` aside rather than comparing it.
- The prose findings are recorded by the diff that changed the prose.
- Full `testthat::test_local()` under `NOT_CRAN=true`,
  `lintr::lint_package()`, `verify-context-budget.R`, and
  `verify-suite-coverage.R` all pass on `5eb6569`.

PR #319, opened 2026-08-29
Fix caused-read tracing blind spots

### LINKED ISSUE #303: The caused-read gate misses a bare entry-point call, and one nested in a subject test's argument

sayuks:
Closes #303

## Summary

- trace both namespace and attached package bindings for execution entry points
- preserve nested entry-point observations while a subject-tested tracer forces its argument
- add positive controls for bare and nested calls and record the decision in ADR 0020

## Verification

- Full testthat suite: 5,566 passed, 0 failed, 0 warned, 0 skipped
- Package lint: no lints
- Always-loaded context budget: 38,753 / 38,810 bytes

### DIFF
--- a/design/adr/0020-ask-before-reading-a-lazy-input.md
+++ b/design/adr/0020-ask-before-reading-a-lazy-input.md
@@ -171,6 +171,13 @@ a non-clustered table, a one-row query that starts a Snowflake warehouse costs
 a minute of credits, every read on Aurora Standard is a billable I/O request,
 and Athena bills a failed query like a successful one. Rows are the wrong unit.
 
+**Require the runtime gate's controls to use qualified calls.** Rejected: the
+catalog identifies an execution entry point by package and name, not by how a
+caller reaches its binding. A qualified-only rule would leave the counter blind
+to the attached `package:*` binding and make that blindness a condition every
+later control has to remember. The gate traces both the namespace binding and
+any distinct attached binding instead (#303).
+
 ## Documentation consequences
 
 The reference carries a *When marginplyr queries your data* section stating
@@ -194,6 +201,11 @@ entry point, and a second snapshot records the set of entry points scanned for,
 so a scan that stopped covering `compute()` fails rather than reporting a clean
 result. A third records which backend kinds hold `collect_selection_proxy`, so
 extending exemption 1 is visible in a diff.
+
+The runtime counter traces every binding a qualified or bare call can reach.
+When a subject-tested entry forces its caller's promise, tracing is re-enabled
+for that force alone, so an execution entry point nested in the argument remains
+visible. Positive controls cover both shapes (#303).
 
 Snapshots run only where `NOT_CRAN` is set, which is the `structure` job and
 the `backend` jobs; a plain CRAN check does not execute this gate.
--- a/tests/testthat/test-query-policy.R
+++ b/tests/testthat/test-query-policy.R
@@ -237,7 +237,7 @@ is_lazy_backend_input <- function(x) {
   !grouping_backend(x)$kind %in% c("local", "other")
 }
 
-# One installation per entry, in a frame of its own.
+# One installation per entry binding, in a frame of its own.
 #
 # The tracer has to be a function *literal* written at the `trace()` call, and a
 # variable holding the same closure is not the same thing. `trace()` substitutes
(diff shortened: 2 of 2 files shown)


sayuks:
## Code review — `0350a82`

## Standards

One finding: the two new below-public-seam controls lacked the adjacent rationale required by `design/architecture.md`’s Test seams rule. Fixed in `434db95` by recording the false-clean result each assertion prevents. No baseline smells found.

## Spec

One finding: attached re-export bindings remained invisible—for example, bare `as_tibble()` from `package:dplyr` while the catalog owner is `tibble`. Fixed in `434db95` by tracing every attached package binding that holds the catalog target and strengthening the control to exercise the re-export route. No scope creep or nested-tracing defect found.

Summary: Standards 1 finding (fixed; worst: missing below-seam rationale); Spec 1 finding (fixed; worst: attached re-export blind spot).

PR #321, opened 2026-08-29
Assert every verifier is still reached by something (#292)

### LINKED ISSUE #292: Six verifier scripts stop running if their workflow step is deleted, and nothing fails
## Problem

Each `.github/scripts/verify-*.R` is invoked as a workflow step. Deleting that
step deletes the assertion, and nothing reports it: the job still runs, still
passes, and the diff that removed it is a few lines of YAML that read like
cleanup. A gate that stops asserting reports nothing, which is the shape every
one of these scripts exists to refuse.

One is already closed against this. `check-tarball.R:25` sources
`verify-library-isolation.R` rather than the workflow calling it, and says why:
so the assertion "cannot be dropped from a job while the job goes on claiming
to check a tarball." Six are not.

## Evidence

```sh
for s in .github/scripts/verify-*.R; do
  b=$(basename "$s")
  grep -lq "source(.*$b" .github/scripts/*.R || echo "$b"
done
```

returns `verify-backend.R`, `verify-context-budget.R`,
`verify-depends-only.R`, `verify-must-error.R`, `verify-site.R`, and
`verify-suite-coverage.R`. `verify-library-isolation.R` is absent from that
list because of the `source()` above.

## What this fixes

One assertion covering all six, derived rather than listed: every
`verify-*.R` under `.github/scripts/` is either named as a step in a workflow
or sourced by a script that is. It is not itself a step — it is sourced from
`check-tarball.R`, for the reason that file already gives for the assertion it
carries.

## Where the regress stops

Not at a count of levels. It stops where deleting the assertion stops being
silent: removing this `source()` means editing a script whose job is to check a
tarball, and removing that job's step means no tarball is checked, which no
diff can make look like cleanup.

## Acceptance

- The set of verifiers is derived from `.github/scripts/`, so a new one is
  covered without an edit here;
- the assertion fails when a workflow step naming a verifier is deleted,
  demonstrated against a copy under `/tmp`;
- it asserts its own mechanism before its verdict — a derivation returning an
  empty set fails naming itself, not the repository;
- its failure message names what to do, as `verify-suite-coverage.R:222-244`
  does;
- it is reached by `source()` and not by a step of its own, and
  `check-tarball.R` says why beside the call.

## Not in scope

Guarding `check-tarball.R`'s own step. That is where the regress is left, and
the reason is above.

sayuks:
Closes #292.

## What this adds

`.github/scripts/verify-verifier-invocation.R`. It derives the set of
`verify-*.R` under `.github/scripts/` from the directory and fails unless each
member is named by a workflow step or sourced by a script that is. Reaching is
transitive: the walk starts at the scripts workflow steps name and follows
`source()` calls out from there, so `verify-library-isolation.R` is reached
through `check-tarball.R`.

`check-tarball.R` sources it, as it already sources `verify-library-isolation.R`
and for the same reason one level out. Guarding `check-tarball.R`'s own step is
out of scope, per the issue: removing the `source()` means editing a script
whose job is to check a tarball, and removing that job's step means no tarball
is checked.

`AGENTS.md` gains the rule, and `verify-context-budget.R`'s baseline moves in
the same commit that spends it.

## What counts as an invocation

`Rscript <path>` on a workflow line, or `source("<path>")` in a script, with
full-line comments discarded first. Matching a bare path would count a step's
`name:` or an artifact value, and a workflow comment quoting a gate's command
for a local run is not a step — `release-matrix.yaml:56` is one.

## Mechanism assertions

Each fails naming the script rather than the repository:

- no `verify-*.R` found — the loop iterates over nothing;
- no workflow found — nothing could be found invoking anything;
- the reader applied to a probe pair (one `run:` line, one commented line) does
  not return exactly the uncommented one.

## Demonstrated against a copy under `/tmp`

`git archive HEAD` into `/tmp`, then, one at a time:

| change | result |
|---|---|
| baseline copy | passes, 8 verifiers, each with its route |
| a `run: Rscript .github/scripts/verify-*.R` step deleted | fails naming that verifier |
| the `source()` deleted from `check-tarball.R` | fails naming `verify-library-isolation.R` |
| `verify-site.R` renamed with the step left as it was | fails twice: the new name is run by nothing, and the path the step names answers to no file |
| every `verify-*.R` moved away | fails naming its own derivation |
| `.github/workflows/` moved away | fails naming its own reading |
| comment stripping disabled | fails naming its own reader |

## Review

One round, recorded in a comment below.

### DIFF
--- a/.github/scripts/check-tarball.R
+++ b/.github/scripts/check-tarball.R
@@ -25,11 +25,9 @@ source(".github/scripts/ci-helpers.R")
 source(".github/scripts/verify-library-isolation.R", local = new.env())
 
 # Asserts that every other verifier in that directory is still reached by a
-# workflow step or by a script that is. It runs here for the same reason the
-# assertion above does, one level out: a verifier reached only by a step is
-# silenced by deleting that step, and deleting this `source()` means editing a
-# script whose job is to check a tarball. That is where the regress stops, and
-# `verify-verifier-invocation.R`'s header says why it stops there (#292).
+# workflow step or by a script that is. It runs here rather than as its own
+# step for the same reason the assertion above does, one level out; the
+# script's own header says where that regress stops (#292).
 source(".github/scripts/verify-verifier-invocation.R", local = new.env())
 
 tarball_dir <- Sys.getenv("MARGINPLYR_TARBALL_DIR", "tarball")
--- a/.github/scripts/check-tarball.R
+++ b/.github/scripts/check-tarball.R
@@ -24,6 +24,14 @@ source(".github/scripts/ci-helpers.R")
 # separate environment keeps its working names off this one.
 source(".github/scripts/verify-library-isolation.R", local = new.env())
 
+# Asserts that every other verifier in that directory is still reached by a
+# workflow step or by a script that is. It runs here for the same reason the
+# assertion above does, one level out: a verifier reached only by a step is
+# silenced by deleting that step, and deleting this `source()` means editing a
+# script whose job is to check a tarball. That is where the regress stops, and
+# `verify-verifier-invocation.R`'s header says why it stops there (#292).
+source(".github/scripts/verify-verifier-invocation.R", local = new.env())
+
 tarball_dir <- Sys.getenv("MARGINPLYR_TARBALL_DIR", "tarball")
 check_dir <- check_directory()
 label <- check_label()
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -24,8 +24,11 @@ source(".github/scripts/ci-helpers.R")
 # What the closure measured when it was last set, and the commit that set it.
 # Moving this means saying in the same commit what was added and what paid for
 # it.
-baseline_bytes <- 38810L
-baseline_note <- "#298 moving three review instructions into the closure"
+baseline_bytes <- 39223L
+baseline_note <- paste(
+  "#292 adding the verifier-invocation rule, paid for by nothing:",
+  "the gate it names is what makes every other one here undeletable"
+)
 
 entry <- "CLAUDE.md"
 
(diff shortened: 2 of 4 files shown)


sayuks:
## Review record

One round of `mattpocock-skills:code-review` against the merge-base with `main`, read at 547329b, answered in 7e50c62.

### Standards — six findings

| finding | disposition |
|---|---|
| `check-tarball.R`'s comment re-derives the regress argument the script header owns, against *Code comments* | fixed in prose: the site now says which way it falls and cites |
| `AGENTS.md` names all seven existing verifiers and not this one, which `design/agents/code-review.md` states it does | fixed in prose: rule added, context baseline moved in the same commit with what paid for it |
| "every verifier here is named in prose somewhere" is false as scoped | fixed in prose: the claim is deleted, the comment states what stripping is for |
| the `call. = FALSE` comment is byte-identical to `verify-library-isolation.R`'s | fixed in prose: reduced to a citation |
| `basename()` over a `list.files()` that already returns bare names; `sort()` over an already-sorted set | fixed in code |
| `scripts_dir` interpolated raw, so `.github`'s `.` is a wildcard | fixed in code: escaped with `gsub(..., fixed = TRUE)` |

### Spec — two findings

| finding | disposition |
|---|---|
| the header claims the `Rscript`/`source(` prefix rules out a mention; it does not for a workflow value outside `run:` that spells the whole command, demonstrated by a `name:` carrying it | fixed in prose: the false claim is deleted, and what the match does and does not reach is stated, with the reason narrowing to `run:` would misread a block scalar |
| the `absent` report is a second assertion no spec line asks for | rejected: the walk must test `file.exists()` before reading, and skipping a name silently is the shape this gate refuses. The comment at the site carries it |

Both axes confirmed the five Acceptance bullets by execution against a `/tmp` copy, including that `release-matrix.yaml:56`'s comment quoting `Rscript .github/scripts/verify-suite-coverage.R` does not rescue that verifier when its step is deleted.

### Checks after the answers

- `Rscript .github/scripts/verify-verifier-invocation.R` — passes, 8 verifiers
- `Rscript .github/scripts/verify-context-budget.R` — 39223 of a 39223-byte baseline
- `lintr::lint_dir(".github", ...)` — no lints
- `devtools::test()` — `[ FAIL 0 | WARN 0 | SKIP 0 | PASS 5567 ]`

PR #322, opened 2026-08-29
Check pluralization coverage locations

### LINKED ISSUE #311: The pluralization coverage table's line references into test-grouping-plan.R point at unrelated lines
## Problem

`pluralizing_coverage()` in `tests/testthat/test-diagnostic-pluralization.R`
says, per namespace binding that raises a pluralizing diagnostic, where both
arms of that diagnostic are pinned. Six of its references are `file:line` pairs
into `tests/testthat/test-grouping-plan.R`, and all six land on unrelated code.

They were exact when written. `14a6ded` ("Define the pluralization baseline by
coverage, not provenance", 2026-08-21) introduced all three entries, and at that
commit every one of the six numbers sat on the `expect_identical()` text it
names. `test-grouping-plan.R` has gone from 928 lines to 1940 since, and the
references went with it.

The table's own header concedes that nothing checks them:

> What it cannot check is that a line reference still points at the assertion it
> names -- that is review surface, as a marker's text is.

So this rots silently and reads, to anyone who does not follow a reference, like
a table that is maintained. That is the same failure mode the header argues
against one paragraph earlier, where it explains why the *set* of covered
bindings is derived rather than listed: "a table that fell behind fails rather
than shrinking the question". The set is derived; the locations are not.

The references into the other three files still land, because those files did
not grow above the lines named. So this is a `test-grouping-plan.R` problem
today and a general one in principle.

## Reproduction / evidence

Against `origin/main` at `a222eeb`. The table's order is singular arm then
plural arm, not source order — worth preserving in any fix.

| entry | the table says | what is at that line now | where the arm actually is |
| --- | --- | --- | --- |
| `abort_grouping_rename` | `:493 and :543` | `rlang::call2("(", ...)` ; `data_vars <- c("s", "grade", "value")` | `:1505` `"Can't rename grouping dimension:\n"` ; `:1555` `"Can't rename grouping dimensions:\n"` |
| `abort_by_rename` | `:604 and :645` | a comment about nested ambiguity ; the string `"nested_raising"` | `:1616` `"Can't rename \`.by\` column:\n"` ; `:1657` `"Can't rename \`.by\` columns:\n"` |
| `compile_grouping_spec_impl` (first element) | `:915 and :768` | a comment about compilation passes ; a `compile(eval(quote(grouping_sets(rollup(value)))...))` call | `:1927` `"Duplicate grouping sets were produced at positions:\n"` ; `:1780` `"...at position groups:\n"` |

That every one was correct at `14a6ded`:

```
$ git show 14a6ded:tests/testthat/test-grouping-plan.R > /tmp/at14.R
$ for n in 493 543 604 645 768 915; do printf "%4s  " "$n"; sed -n "${n}p" /tmp/at14.R; done
 493      "Can't rename grouping dimension:\n",
 543        "Can't rename grouping dimensions:\n",
 604      "Can't rename `.by` column:\n",
 645        "Can't rename `.by` columns:\n",
 768        "Duplicate grouping sets were produced at position groups:\n",
 915        "Duplicate grouping sets were produced at positions:\n",
```

The three entries that still resolve, for contrast — `abort_absorbed_summary`
(`test-grouping-backends.R:197 and :540`), `check_summary_context_helpers`
(`test-contextual-helpers.R:696 and :728`), and `report_branch_warnings`
(`test-execution-conditions.R:332 and :291`) — each land on their assertion.

Not in scope: `.github/scripts/verify-backend.R:85-86` also spells
`test-factor.R:124` and `test-share-backends.R:53`, but those are sample
testthat output inside a comment describing that tool's format. They are not
references to assertions and are not required to resolve.

## Decision required

Whether a location in this table should stay a line number is a maintainer's
call, because the cheap disposition is the one that has already failed once.

- **Refresh the six numbers.** Cheapest, changes no structure, and will rot
  again on the same mechanism — `test-grouping-plan.R` grew a thousand lines in
  eight days, and nothing fails when it does.
- **Reference something the file can be searched for, and assert it resolves.**
  Name the `test_that()` block holding each arm and have the existing test fail
  when the named file holds no block by that name. That is derived rather than
  listed, in the idiom `AGENTS.md` applies to every other gate here, and it
  turns "review surface" into something a run can answer. One wrinkle to
  settle: `abort_by_rename`'s singular expected text sits in the
  `by_rename_message()` helper above the block, not inside it, so the reference
  would name the block whose assertion consumes the helper.
- **Drop the location column.** Keep the derived key set and stop claiming to
  say where each went. Honest, and it loses the only thing that routes a reader
  from a binding to its two arms.

## Acceptance criteria

- [ ] Every location in `pluralizing_coverage()` resolves to the arm it names,
      for all three files, not only the three entries that drifted.
- [ ] The singular-then-plural order of each pair is preserved, or the header
      says the order changed and to what.
- [ ] Whichever disposition is chosen, the header no longer describes as
      unchecked something a run now checks — or, if the reference stays
      unchecked, says why refreshing it is expected to hold this time.

## Scope boundaries

Deliberately unchanged:

- The diagnostics themselves, their wording, and their condition classes.
- The assertions the table points at.
- Which bindings the table has entries for, and how that set is derived — the
  derivation is not what failed here.
- The `verify-backend.R` comment above, which is illustrative output.

## Testing seam

`tests/testthat/test-diagnostic-pluralization.R`, in or beside the block that
already asserts the table names every derived binding. A check that each
location resolves belongs with the check that the key set is complete, since
they are the two halves of the same claim.

## Review origin

Found while answering the code review of #260's branch (PR #309), by following
a reference in this table and landing on an unrelated comment. Verified
independently of that branch: the drift is present at the merge-base `85d3735`
and at `a222eeb`, and `git log -S` attributes all three entries to `14a6ded`,
where they were correct.

sayuks:
## Summary

- replace drifting `file:line` references with named `test_that()` blocks
- preserve singular-then-plural ordering in structured locations
- parse referenced test files and require every named block to resolve exactly once

## Tests

- `Rscript -e 'pkgload::load_all(".", quiet = TRUE); lintr::lint_package()'`
- `Rscript -e 'pkgload::load_all(".", quiet = TRUE); testthat::test_local(".")'` (5586 passed)
- `Rscript .github/scripts/verify-context-budget.R`

Closes #311

### DIFF
--- a/tests/testthat/test-diagnostic-pluralization.R
+++ b/tests/testthat/test-diagnostic-pluralization.R
@@ -86,7 +86,11 @@
 # `test-grouping-plan.R` fails any test file whose source spells it as a call
 # -- the scan behind "the only caller of the plan compiler" is deliberately
 # blunt, and a key in a table is not a call.
-pluralizing_test_location <- function(file, ..., .additional = character()) {
+pluralizing_test_location <- function(
+  file,
+  ...,
+  .local_locations = character()
+) {
   blocks <- c(...)
   label <- paste0(
     file,
@@ -94,7 +98,7 @@ pluralizing_test_location <- function(file, ..., .additional = character()) {
     paste0("`", blocks, "`", collapse = " then ")
   )
   structure(
-    c(label, .additional),
+    c(label, .local_locations),
     class = c("pluralizing_test_location", "character"),
     file = file,
     blocks = blocks
@@ -132,7 +136,7 @@ pluralizing_coverage <- function() {
       "test-grouping-plan.R",
       "compile_grouping_spec() reads a narrowed duplicates vocabulary",
       "duplicated grouping sets in more than one group name their groups",
-      .additional = "this file, the unknown `.by` invariant block"
+      .local_locations = "this file, the unknown `.by` invariant block"
     ),
     grouping_helper_vars = "this file, the grouping-helper block",
     proxy_columns = "this file, the selection proxy invariant block",
--- a/tests/testthat/test-diagnostic-pluralization.R
+++ b/tests/testthat/test-diagnostic-pluralization.R
@@ -71,8 +71,10 @@
 # verify-site.R` draws about its markers and `AGENTS.md` explains there: the
 # set of things that must be covered is derived, and the table only says where
 # each one went, so a table that fell behind fails rather than shrinking the
-# question. What it cannot check is that a line reference still points at the
-# assertion it names -- that is review surface, as a marker's text is.
+# question. An external location names the `test_that()` block holding each
+# arm, in singular-then-plural order; one name means that one block holds both.
+# The coverage gate parses each named file and requires each block name to
+# resolve exactly once, so a moved or renamed pin makes this table fail.
 #
 # A binding holding more than one is why the value is a vector. The plan
 # compiler's implementation, named as a key below, raises the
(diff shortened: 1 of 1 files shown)


sayuks:
Code review — fixed point `main@687328c`, reviewed snapshot `8b6caea`.

## Standards

1 judgement-call finding: possible Mysterious Name in `.additional`, which did not reveal that it held local/non-parsed coverage locations. Disposition: fixed in `76ad6ad` by renaming it to `.local_locations`. No hard documented-standard violations.

## Spec

No findings. The chosen #311 disposition, acceptance criteria, testing seam, and scope boundaries are all satisfied.

Summary: Standards 1 finding (fixed; worst: possible Mysterious Name); Spec 0 findings.

PR #323, opened 2026-08-29
Record a Grouping plan's kind as the name it spells (#317)

### LINKED ISSUE #317: Decide what a Grouping plan records as its kind, and what asks it
## Problem

A compiled Grouping plan records its kind as the specification's `type` field
was read, class and attributes included:

```r
# R/grouping-plan.R, compile_grouping_spec_impl()
kind = preflight$spec$type,
```

`R/share.R` then asks that field an identity question:

```r
check_parent_grouping_kind <- function(plan) {
  if (!identical(plan$kind, "rollup")) {
    abort_ambiguous_parent()
  }
  invisible(NULL)
}
```

A kind that spells `rollup` while carrying any attribute compiles as a rollup
and is refused by `share_of_parent()` as though it were not one. Measured on
this branch and on `main`:

```r
d <- data.frame(g = c("a", "a", "b"), h = c("x", "y", "x"), v = c(1, 2, 3))
s <- marginplyr:::new_grouping_spec(
  c(nm = "rollup"),
  list(rlang::quo(g), rlang::quo(h))
)
summarize_with_margins(d, n = sum(v), share = share_of_parent(n), .grouping = s)
#> Error: `share_of_parent()` requires `.grouping` to be one pure `rollup()`.
```

The plan itself is a rollup plan — the same call without the Parent share
compiles and returns the rollup's sets. The names-only shape above predates
#289; the classed shape reaches the plan too, because #289 stopped a class on
a kind being a reason to refuse the specification.

## Scope

Nothing a constructor builds has this shape: `rollup()` stores the string
`"rollup"`. It is reachable only from a specification object assembled by hand,
which is the same population every other guard in this lineage (#262, #264,
#268, #280, #289) is written for.

## What has to be decided

Two questions, and they are not the same one.

1. **What a plan records.** The classified name, or the field as read.
   Recording `grouping_kind_name(preflight$spec$type)` was tried on #289's
   branch and reverted: it is not a consequence of that ticket's decision, and
   it does not close the defect on its own, because `unclass()` removes the
   class and nothing else. A plan holding the name would need the name to be
   free of every attribute, which is a separate decision about what
   `grouping_kind_name()` answers with — it currently promises "one character
   element with the class off", and `%in%` and `[[`, its only consumers today,
   read no other attribute.
2. **What `R/share.R` should ask.** `identical()` on the raw field is the
   comparison that fails. Asking `grouping_kind_name(plan$kind)` there, or
   holding the plan to a normalized kind at construction, are different places
   to put the same answer, and ADR 0008's compatibility constraint on "the
   Grouping plan representation or its consumers" is what the second one
   reaches.

`check_parent_grouping_spec()` reads `grouping_spec$type` raw before the plan
is compiled, so a decision here has to say whether that reader moves too.

## Expected outcome

A decision recorded in ADR 0008, and a test that fails without the fix, over
both shapes — a kind carrying names and a kind carrying a class — through
`share_of_parent()` rather than through the plan field alone, since the field
is internal and the refusal is what a caller sees.

## Origin

Raised by both axes of the review of #289's branch (PR #316), against a change
made there and reverted in the same review.

sayuks:
Closes #317.

## The defect

A compiled Grouping plan recorded its kind as the specification's `type` field
was read, attributes and all:

```r
kind = preflight$spec$type,
```

`R/share.R` then compared that field with `identical()`, which reads every
attribute. A kind spelling `rollup` under a name or a class compiled as a rollup
plan and was refused by `share_of_parent()` as though it were not one — the same
call without the Parent share returns the rollup's sets.

## The decision

Recorded in ADR 0008, as the amendment *a kind read as a name and compared as a
value*.

- **What `grouping_kind_name()` answers with.** A bare name: every attribute
  comes off, not only the class #289 took. `unclass()` takes the class and
  `attributes<-` takes what is left; the second is a primitive like the first
  two, so `setMethod()` refuses a method for it, and by the time it is called
  there is no class left for anything to dispatch on in any case.
- **What a plan records.** That answer. A plan is compiled from a preflighted
  specification, so the field is one name by then and the classification never
  answers `NULL` there — what it removes is the attributes a caller's kind
  carried onto the plan. Putting the question here rather than at each consumer
  makes a plan's kind one string by construction, for every consumer the plan
  reaches and every one written later.
- **What the share readers ask.** `check_parent_grouping_kind()` compares
  `plan$kind` directly, because a plan holds a name.
  `check_parent_grouping_spec()` runs before a plan exists, so it classifies the
  field it read. The two agree because both compare a name, not because both
  were written to remember to.

## What moves

One population, at one guard: a specification whose kind spells `rollup` under
an attribute goes from refused to accepted. No refusal changes its condition
class, message, call context, or position in the detection order, and no
specification that was accepted is refused. Nothing a constructor builds is
affected — `rollup()` stores the string `"rollup"`, so stripping is the identity
for every kind marginplyr writes. The other consumers of a classified kind read
no attribute either: `%in%` compares values, `[[` indexes a list by one, and the
printed line reaches `cat()`, which renders values and not labels.

## Tests

- `test-share.R` — through `share_of_parent()` rather than the plan field, since
  the field is internal and the refusal is what a caller sees. Both shapes the
  ticket names, compared against the plain `rollup()` result rather than merely
  asserted to return; and the refused pair as well, because a comparison that
  stopped discriminating passes the accepted half alone.
- `test-grouping-plan.R` — what the plan records, one attribute per case, since
  `identical()` fails on any one of them.
- `test-grouping-plan.R`'s class-off test stops setting the kind field aside; it
  now compares it, and its comment says why.

Both new tests fail without the change. No test added or changed here requires
an optional backend.

## Local checks

`lintr::lint_package()` clean; full `testthat::test_local()` green;
`verify-suite-coverage.R` verified all 640 tests execute in at least one
single-backend configuration. No generated file changes (no roxygen, README, or
`man/` edits).

### DIFF
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -714,10 +714,9 @@ compile_grouping_spec_impl <- function(preflight,
 
   structure(
     list(
-      # The name the field spells, not the field as read, so that a consumer
-      # comparing this with `identical()` compares what the specification said
-      # and no attribute it carried (#317, ADR 0008). The preflight above ran
-      # the guard, so the field is one name and this is never `NULL`.
+      # The name the field spells, not the field as read (#317, ADR 0008). The
+      # preflight above ran the guard, so the field is one name and this is
+      # never `NULL`.
       kind = grouping_kind_name(preflight$spec$type),
       by = unique(.by),
       dimensions = dimensions,
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -714,7 +714,11 @@ compile_grouping_spec_impl <- function(preflight,
 
   structure(
     list(
-      kind = preflight$spec$type,
+      # The name the field spells, not the field as read, so that a consumer
+      # comparing this with `identical()` compares what the specification said
+      # and no attribute it carried (#317, ADR 0008). The preflight above ran
+      # the guard, so the field is one name and this is never `NULL`.
+      kind = grouping_kind_name(preflight$spec$type),
       by = unique(.by),
       dimensions = dimensions,
       sets = normalized,
@@ -900,24 +904,26 @@ grouping_kind_rules <- local({
 })
 
 # The name a kind read off an object is, or `NULL` where it is none: one
-# character element with the class off, and not the missing one. Every caller
+# character element carrying no attribute, and not the missing one. Every caller
 # holds a value nothing has validated, and every reader of one shares this, as
 # they share the read in `grouping_spec_kind()`. What it decides, why it answers
-# with the name rather than with whether there is one, and why nothing here is
-# asked of the object's own methods are ADR 0008's, in its amendment for a kind
-# classified with its class off.
+# with the name rather than with whether there is one, and why the answer
+# carries no attribute are ADR 0008's, in its amendments for a kind classified
+# with its class off and for a kind read as a name and compared as a value.
 grouping_kind_name <- function(kind) {
   # `is.character()` and `unclass()` are primitives that dispatch on neither S3
   # nor S4, so neither reaches a method the value carries: the type test is R's
   # own, and stripping is what puts the two questions below to a character
   # vector with no class rather than to whatever the class defines for them
-  # (#289). Only the class comes off, so a name and any other attribute the
-  # value carried survive onto the answer; `%in%` and `[[` are what the callers
-  # put it to, and neither reads one.
+  # (#289). Every remaining attribute comes off with it, because `identical()`
+  # reads them all and that is what a consumer compares this answer with (#317).
+  # `attributes<-` is a primitive too, and by then there is no class left for
+  # anything to dispatch on.
   if (!is.character(kind)) {
     return(NULL)
   }
   name <- unclass(kind)
+  attributes(name) <- NULL
   if (length(name) != 1L || is.na(name)) {
     return(NULL)
   }
--- a/R/share.R
+++ b/R/share.R
@@ -710,8 +710,14 @@ share_grouping_spec_validator <- function(kinds) {
 }
 
 check_parent_grouping_spec <- function(grouping_spec) {
+  # The field is read bare rather than through `grouping_spec_kind()`, because
+  # `validate_grouping_spec_early()` has run by the time this validator is
+  # called and is what establishes there is a kind to read.
   kind <- if (is.null(grouping_spec)) NULL else grouping_spec$type
-  if (!identical(kind, "rollup")) {
+  # Classified rather than compared as read: `identical()` reads every
+  # attribute, so a kind spelling `rollup` under a name or a class was refused
+  # here while compiling as a rollup plan (#317).
+  if (!identical(grouping_kind_name(kind), "rollup")) {
     abort_ambiguous_parent()
   }
   invisible(NULL)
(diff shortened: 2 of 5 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes, against the merge-base with `main`
(`git diff main...HEAD`, merge-base `b95ea30`) over `a42976c`. Answers landed in
`21d2827`.

### Standards

| Finding | Disposition |
|---|---|
| `test-grouping-plan.R`'s new header claims one attribute per case is what makes a partial strip detectable, and that a combined case would pass a class-only strip. False: `unclass()` on a value carrying a class and a name leaves the name, so `identical()` fails either way. | **Fixed in prose** (`21d2827`). Corrected to what one per case actually buys — naming which attribute survived — rather than restated, per *Code comments* in `AGENTS.md`. |
| The comment on the plan's `kind` field re-derives ADR 0008's *What a plan records* beside a citation to that ADR — the second copy *Code comments* names. Same shape, weaker, at `check_parent_grouping_spec()`. | **Fixed in prose** (`21d2827`). The clause goes at both; the citation and each site's own fact stay. The share-side comment now says which way its site falls: it holds the caller's specification and not a plan. |
| Duplicated Code (judgement): the decorated-kind fixture and the string `"decorated_grouping_kind"` are built in both test files. | **Rejected.** `helper-forged-kinds.R`'s own header states why it is shared: "the hazard is in the registration and not in the fixture" — two sites spelling one class name each get whichever method was registered last. Neither new fixture registers a method, so the hazard does not arise, and the two differ in shape (one varies the attribute, one varies the kind). |
| Duplicated Code (judgement): `identical(…, "rollup")` + `abort_ambiguous_parent()` at two sites, with nothing structural holding them in step. | **Rejected.** The shared extraction already exists and is `abort_ambiguous_parent()`, whose comment owns the argument for why there are two raising sites. The comparison cannot be shared: one reads a specification and one reads a plan, which is the asymmetry the ADR amendment records. Pre-existing in shape. |

Verified with no finding: every `grouping_kind_name()` call site (`is.null`,
`%in%`, `[[`, `cat`) reads no attribute, so no accepted specification becomes
refused; the "never `NULL`" claim holds on every path, including `.grouping =
NULL`; `validate_grouping_spec_early()` does run before the validator; the
class-off test was strengthened, not weakened; neither new test requires a
member of `optional_backends()`. The ADR's factual claims were re-measured —
`attributes<-` is a primitive `setMethod()` refuses, an S4 class extending
`character` unclasses to `class() == "character"`, the printed line is unchanged
under `names`/`class`/`dim`, and the "tried on #289's branch and reverted"
claim is true (`d67eec0` added it, `5eb6569` reverted it).

### Spec

No missing requirement, no scope creep, no wrong implementation. Both of the
issue's separated questions and the `check_parent_grouping_spec()` question are
answered in the ADR amendment. The test fails without the fix over both shapes
the issue names, through `share_of_parent()`; each half of the fix is
load-bearing — with only `R/share.R` patched, all four rollup shapes are still
refused, so the test reaches `check_parent_grouping_kind()` too. `cube` under
every shape is still refused, so the comparison did not merely stop
discriminating. The issue's reproduction now returns the rollup with correct
shares.

| Finding | Disposition |
|---|---|
| `validate_nested_grouping_units()` still compares raw: `if (!identical(nested$type, "set"))`. A nested `grouping_set` spelling `set` under an attribute is refused inside `rollup()`, where the bare string compiles. `expand_rollup()`'s `stopifnot()` is the same shape one pass later, unreachable behind it. | **Ticket** — #324. #317 names neither site and scoped its amendment to "the one field a plan records a kind in", so this needs its own decision about the comparison and about the diagnostic that interpolates `parent$type`. Per *Answering a review* in `design/architecture.md`: a finding this branch's Acceptance does not reach becomes a ticket. |

**Summary.** Standards: 4 findings — worst, one false claim in a test comment,
fixed. Spec: 1 finding — worst and only, a site outside this branch's
Acceptance, ticketed as #324.

PR #325, opened 2026-08-29
Compare a nested kind as a name (#324)

### LINKED ISSUE #324: Decide what a nested-specification guard compares a kind with
## Problem

`validate_nested_grouping_units()` compares a nested specification's kind with
`identical()` on the field as read:

```r
# R/grouping-plan.R
validate_nested_grouping_units <- function(parent, nested) {
  if (!identical(nested$type, "set")) {
    abort_marginplyr(paste0(
      "{.fun {parent$type}} only accepts columns or {.fun grouping_set} ",
      "composite dimensions."
    ))
  }
```

A nested `grouping_set` whose kind spells `set` while carrying any attribute is
therefore refused inside `rollup()` and `cube()`, where the same specification
with a bare `"set"` compiles. Measured on `main` and on #317's branch:

```r
d <- data.frame(a = 1, b = 2, v = 3)
inner <- marginplyr:::new_grouping_spec(
  c(nm = "set"),
  list(rlang::quo(a), rlang::quo(b))
)
summarize_with_margins(d, n = sum(v), .grouping = rollup(!!inner))
#> Error: `rollup()` only accepts columns or `grouping_set()` composite dimensions.
```

The nested specification is a `set`: the preflight classified it as one — that
is how its own rule was found — and the same call with the bare string returns
the composite dimension.

`expand_rollup()`'s `stopifnot(identical(arg$nested$spec$type, "set"))` is the
same comparison one pass later. It is unreachable behind the guard above today,
so a decision about the guard has to say whether that assert moves too.

## Scope

The same population as #262, #264, #268, #280, #289, and #317: nothing a
constructor builds has this shape — `grouping_set()` stores the string `"set"` —
so it is reachable only from a specification object assembled by hand.

## What has to be decided

#317 decided that `grouping_kind_name()` answers with a bare name and that a
Grouping plan records that answer, and recorded it in ADR 0008. It scoped its
amendment to "the one field a plan records a kind in", and deliberately did not
reach this guard, which reads a nested specification rather than a plan.

Two things follow, and neither is settled by that decision.

1. **What the guard compares.** Whether it classifies the field it read, as
   `check_parent_grouping_spec()` now does, or is handed a kind something else
   normalized. The parent's own rule was already found through
   `grouping_kind_name()`, so a third answer is that the preflight passes the
   name it classified to the rule rather than the rule re-reading the field.
2. **What the diagnostic renders.** It interpolates `parent$type` into a glue
   template, so the same question is asked of the *parent's* kind by the message
   rather than by a comparison. A decision about the comparison should say
   whether the message moves with it.

## Expected outcome

A decision recorded in ADR 0008 — the constraint it amends is the one holding
error condition classes, messages, and detection order fixed, since a refusal
stops firing for a population — and a test that fails without the fix, through a
Margin verb rather than through the guard alone, since the guard is internal and
the refusal is what a caller sees.

## Origin

Raised by the Spec axis of the review of #317's branch (PR #323), as a site
#317's decision does not reach.

sayuks:
Closes #324.

## The defect

`validate_nested_grouping_units()` compared a nested specification's kind with
`identical()` on the field as read:

```r
if (!identical(nested$type, "set")) {
```

A nested `grouping_set` whose kind spells `set` while carrying any attribute was
therefore refused inside `rollup()` and `cube()`, where the same specification
with a bare `"set"` compiles. #317 scoped its answer to the one field a plan
records a kind in and deliberately did not reach this guard, which reads a
caller's nested specification rather than a plan.

## The decision

Recorded in ADR 0008, as the amendment *a nested kind read as a name*.

- **What the guard compares.** The name, classified where it reads — as
  `check_parent_grouping_spec()` does, since what this reader holds is the
  caller's own specification and not a plan.
- **What moves with it.** `resolve_grouping_units()`'s assert on the same field
  one pass later. It was unreachable while the guard refused this population;
  admitting the population makes it reachable, and left as read it would refuse
  exactly that population again with an untyped `stopifnot()` message.
- **What the diagnostic renders.** Unchanged. `abort_marginplyr()` expands a
  message through `cli::format_inline()`, which renders a vector's values: it
  drops names and `dim` and reaches neither an `as.character()` nor a `format()`
  method a class on the kind defines, so `{.fun {parent$type}}` already spells
  the name. Measured, not assumed — the test asserts it over a parent whose own
  kind carries each attribute.
- **Rejected.** Having the preflight hand the rule the name it already
  classified, and a `grouping_kind_is()` predicate over the three sites now
  spelling `identical(grouping_kind_name(x), <name>)`. The ADR says why for
  each.

## What moves

A nested specification whose kind spells `set` under any attribute reads inside
`rollup()` and `cube()` as the same specification with a bare `set` does, at one
guard.

- A non-empty one is accepted where it was refused.
- An empty one is refused still, and by the refusal the bare spelling already
  receives — `abort_empty_composite()`, the line after the comparison — rather
  than by the nesting refusal that used to stand in front of it. That
  sub-population is the one place this change moves a refusal instead of
  removing one, so its message and its position in the detection order both
  change; its condition class does not.

No other refusal changes in any respect, and no specification that was accepted
is refused. Nothing a constructor builds is affected — `grouping_set()` stores
the string `"set"`. The remaining reads of a kind field do not move:
`abort_empty_grouping_units()` takes one for a message alone,
`admitted_nested_kinds()` keys its memo on one and `[[` indexes a list by a
value, and `find_grouping_kind_rule()` classifies already.

## Tests

`test-grouping-plan.R` — through `summarize_with_margins()` rather than the
guard, since the guard is internal and the refusal is what a caller sees, and
because that route is also what reaches the assert one pass later. Both parents,
three attribute shapes, and:

- the accepted half compared against the plain `grouping_set()` result rather
  than merely asserted to return;
- the empty composite beside each accepted shape, asserted against the message
  the bare `set` receives rather than against that sentence;
- the refused half over the same shapes, since a comparison that stopped
  discriminating passes the accepted half alone;
- a parent whose own kind carries each attribute, which is what pins the
  diagnostic.

The test fails without either half of the change: without the guard's
classification it fails on the guard's refusal, and without the assert's it
fails on `identical(arg$nested$spec$type, "set") is not TRUE`. No test added
here requires an optional backend.

## Local checks

`lintr::lint_package()` clean; full `testthat::test_local()` green;
`verify-context-budget.R` unchanged at 39223 bytes;
`verify-suite-coverage.R` verified all 641 tests execute in at least one
single-backend configuration. No generated file changes (no roxygen, README, or
`man/` edits).

Review record: the round is in the comment below.

### DIFF
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -276,7 +276,11 @@ allow_nested_grouping <- function(parent, nested) {
 }
 
 validate_nested_grouping_units <- function(parent, nested) {
-  if (!identical(nested$type, "set")) {
+  # Classified rather than compared as read (#324, ADR 0008): what this reader
+  # holds is a nested specification the caller wrote and not a plan. The
+  # diagnostic keeps the parent's field as read: `cli::format_inline()` renders
+  # a vector's values and reaches no method an attribute could carry.
+  if (!identical(grouping_kind_name(nested$type), "set")) {
     abort_marginplyr(paste0(
       "{.fun {parent$type}} only accepts columns or {.fun grouping_set} ",
       "composite dimensions."
@@ -790,7 +794,10 @@ resolve_grouping_units <- function(preflight, data_proxy) {
           cols <- resolve_grouping_selection(arg$quo, data_proxy)
           return(lapply(cols, function(col) col))
         }
-        stopifnot(identical(arg$nested$spec$type, "set"))
+        # Classified as the guard that admitted this argument classifies
+        # (#324): a comparison on the field as read fires here, untyped, for
+        # exactly the population that guard now admits.
+        stopifnot(identical(grouping_kind_name(arg$nested$spec$type), "set"))
         cols <- resolve_grouping_set(arg$nested, data_proxy)
         if (length(cols) == 0L) {
           abort_empty_composite()
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -278,8 +278,8 @@ allow_nested_grouping <- function(parent, nested) {
 validate_nested_grouping_units <- function(parent, nested) {
   # Classified rather than compared as read (#324, ADR 0008): what this reader
   # holds is a nested specification the caller wrote and not a plan. The
-  # diagnostic keeps the parent's field as read: `cli::format_inline()` renders
-  # a vector's values and reaches no method an attribute could carry.
+  # diagnostic below keeps the parent's field as read, which the same amendment
+  # decides.
   if (!identical(grouping_kind_name(nested$type), "set")) {
     abort_marginplyr(paste0(
       "{.fun {parent$type}} only accepts columns or {.fun grouping_set} ",
--- a/design/adr/0008-centralize-grouping-specification-kind-rules.md
+++ b/design/adr/0008-centralize-grouping-specification-kind-rules.md
@@ -675,11 +675,16 @@ a `stopifnot()` message, which is the untyped condition the guard exists to
 replace.
 
 **What moves.** A nested specification whose kind spells `set` under any
-attribute goes from refused to accepted inside `rollup()` and `cube()`, at one
-guard. No refusal changes its condition class, its message, its call context, or
-its position in the detection order, and no specification that was accepted is
-refused. Nothing a constructor builds is affected: `grouping_set()` stores the
-string `"set"`.
+attribute reads inside `rollup()` and `cube()` as the same specification with a
+bare `set` does, at one guard. A non-empty one is accepted where it was refused.
+An empty one is refused still, and by the refusal the bare spelling already
+receives — `abort_empty_composite()`, the line after the comparison — rather
+than by the nesting refusal that used to stand in front of it. That
+sub-population is the one place this amendment moves a refusal instead of
+removing one, so its message and its position in the detection order both
+change; its condition class does not. No other refusal changes in any respect,
+and no specification that was accepted is refused. Nothing a constructor builds
+is affected: `grouping_set()` stores the string `"set"`.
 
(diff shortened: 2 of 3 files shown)


sayuks:
## Code review round 1

`mattpocock-skills:code-review` over `git diff origin/main...HEAD` at `e3ec9d3`, two axes in parallel. Standards sources: `AGENTS.md` (*Linting*, *Code comments*, *Release matrix*, *Optional-dependency guards*, *Queries against a lazy input*), `design/architecture.md`'s *Answering a review*, and ADR 0008. Spec source: #324.

Three findings. All answered in `6ec3a5c`.

**Spec — the amendment's *What moves* is false.** A nested `set` whose kind carries an attribute and whose args are empty does not go from refused to accepted: on `origin/main` it gets the nesting refusal, on the branch it gets `An empty grouping_set() cannot be a composite dimension.` — the one the bare spelling already received. Message and detection point both move, which is what the amended constraint exists to record. Fixed in the ADR, and the empty shape joins the test beside each accepted one.

**Standards — a second copy of the ADR's argument.** The guard's comment restated *What the diagnostic renders* beside its citation, and had drifted wider than it: "no method an attribute could carry" where the ADR names `as.character()` and `format()` on a class. The clause is deleted, per *Code comments* in `AGENTS.md`.

**Standards — a `grouping_kind_is()` predicate** over the three sites spelling `identical(grouping_kind_name(x), <name>)`. Rejected; recorded in the amendment's *Considered options*.

Both axes independently verified that each half of the change is load-bearing, and that the ADR's `cli::format_inline()` and `[[` claims hold under adversarially registered `as.character()`/`format()` methods. The Spec axis confirmed no site #324 names was missed and found no scope creep.

PR #326, opened 2026-08-29
Say what `## Module responsibilities` holds (#291)

### LINKED ISSUE #291: `## Module responsibilities` reads as a map of `R/` and names thirteen of its twenty-three files
## Problem

`design/architecture.md:54` opens a section of twelve `### Name (R/file.R)`
subsections. Nothing in it says what set it covers, so a reader takes it for
the map of `R/` and finds ten files missing from it.

It is not that map, and it should not become one. Each subsection is an
ownership contract — `:78` says the Grouping plan module "resolves selections
against the prepared metadata but does not read a backend or execute a Margin
operation" — and what the ten omitted files hold is not a contract of that
kind. Four are public-verb wrappers whose lifecycle `R/margin-operation.R`
already owns and documents, two are roxygen with one line of code, and four are
modules whose ownership nobody has needed to write down.

Reading it as an inventory is also the failure this repository has ruled on
twice. `verify-matrix-coverage.R` and the `coverage` job existed to keep a
hand-written matrix in step and were retired for it (#71, #93), and the test
inventory two sections below was removed for the same reason (#284): a list
names what it named when it was written.

## Evidence

```sh
awk '/^## Module responsibilities/{f=1} f&&/^## Opaque/{f=0} f' \
  design/architecture.md | grep -oE 'R/[a-z_-]+\.R' | sort -u | wc -l   # 13
ls R/*.R | wc -l                                                        # 23
```

Omitted: `expand_with_margins.R`, `grouping-context.R`, `grouping-spec.R`,
`inspect-grouping.R`, `internal-names.R`, `marginplyr-package.R`,
`nest_by_with_margins.R`, `nest_with_margins.R`, `retail_sales.R`,
`summarize_with_margins.R`.

## What this fixes

One sentence under the heading, saying what the section holds: the modules
whose ownership is not readable from the file they sit in. `R/` is what says
which files exist.

## Acceptance

- The section states its own scope, so a reader does not measure it against
  `ls R/*.R`;
- no subsection is added or removed to make a count come out;
- adding an `R/*.R` file needs no edit here.

## Not in scope

Writing the four missing ownership contracts. That is a judgement about which
modules have one worth stating, and this ticket is about the section not
claiming to be something it is not.

sayuks:
Closes #291.

`design/architecture.md:54` opens twelve `### Name (R/file.R)` subsections and says nothing about what set they are, so a reader takes the section for the map of `R/` and finds ten of its twenty-three files missing. It is not that map, and #291 rules out making it one: each subsection is an ownership contract, and a file whose own contents make its boundaries readable has none to write there.

One sentence under the heading states the scope and names `R/` as what says which files exist, in the idiom *Test seams* already uses for `tests/testthat/`.

Against the acceptance conditions:

- the section states its own scope, so a reader does not measure it against `ls R/*.R`;
- no subsection is added or removed;
- the sentence names no file, so adding an `R/*.R` needs no edit here.

Writing the four missing ownership contracts is out of scope per the ticket.

No R code changed, and `design/` is `.Rbuildignore`d, so no gate reads the edited section — `verify-context-budget.R` still passes at 39223 bytes (this file is not in the always-loaded closure).

### DIFF
--- a/design/architecture.md
+++ b/design/architecture.md
@@ -53,6 +53,10 @@ on a verb kind.
 
 ## Module responsibilities
 
+Each subsection below is one module's ownership contract, written where the
+file it names does not make its own boundaries readable; `R/` is what says
+which files exist.
+
 ### Margin operation (`R/margin-operation.R`)
 
 Owns the shared prepare and finalize lifecycle, common option normalization,
--- a/design/architecture.md
+++ b/design/architecture.md
@@ -53,9 +53,10 @@ on a verb kind.
 
 ## Module responsibilities
 
-Each subsection below is one module's ownership contract, written where the
-file it names does not make its own boundaries readable; `R/` is what says
-which files exist.
+Each subsection here writes down what one module owns and what it leaves to
+another, which its file does not show on its own; the section covers the
+modules where that has been needed, and not the contents of `R/`, which is
+what says which files exist.
 
 ### Margin operation (`R/margin-operation.R`)
 


sayuks:
One review round, `mattpocock-skills:code-review`, both axes, against `origin/main` at `467f358` — the local `main` ref was stale by two merged PRs, and `design/agents/code-review.md:15`'s warning about a base the branch has fallen behind cuts the same way when the base is the one behind. Snapshot read: `cb1f486`. Answered in `f667c40`.

Three findings, all fixed in prose.

|---|---|---|---|
| 1 | Standards and Spec, independently | The sentence read as a membership criterion: "written where the file it names does not make its own boundaries readable" says every module whose file does not show its boundaries has a subsection here. #291 says four omitted files are exactly that, and puts writing their contracts out of scope — so the criterion reported the section four short, reproducing the "ten files missing" reading at smaller scale. | Fixed. It now covers "the modules where that has been needed", which claims nothing about a module with no subsection. |
| 2 | Standards | "the file it names" was false of `R/utils.R`, a 1144-line file holding two modules, whose subsection scopes itself to one of them in its first line. | Fixed. The sentence says "one module" and names no file, so a heading's parenthesis is not read as a per-file contract. |
| 3 | Spec | "below" was unbounded: `### Structural gates` and `### Release gates` are also subsections below, and are neither modules nor ownership contracts. | Fixed. "here" and "the section" are bounded by the heading. |

The acceptance conditions were re-checked against the answered sentence rather than the reviewed one: the section states its own scope; no subsection was added or removed; and the sentence names no file and imposes no completeness obligation, so adding an `R/*.R` still needs no edit here.

No second round. `design/agents/code-review.md:46` sends a second run only over executable behaviour the first did not see, and this diff has none — the findings are verified against the diff that answers them, which is what records a disposition of this kind.

PR #327, opened 2026-08-29
Delete the argument `inst/suggests/guard.R` already holds (#294)

### LINKED ISSUE #294: *Optional-dependency guards* restates 79 always-loaded lines of an argument `inst/suggests/guard.R` already holds
## Problem

`inst/suggests/guard.R:1-30` argues why the guard exists and why it is shaped
as it is. `AGENTS.md`'s *Optional-dependency guards* argues the same thing
again in 79 lines, and both files are read — the guard's header by anyone
changing it, this section by every task whether or not it touches a guard.

The two have already agreed which is which. `guard.R:29-30`:

> `AGENTS.md` is authoritative for **which guards must go through here** and
> for **what asserts that they do**.

So the guard's header owns the argument and this section owns the rule. The
section does not act like it.

## Evidence

Each of these is in `guard.R` and again in *Optional-dependency guards*:

| the argument | `guard.R` |
|---|---|
| `requireNamespace()` answers "installed", not "new enough" | `:4-6` |
| `duckdb (>= 1.5.5)`, `shared_home = FALSE` a hard error under 1.5.4.x, #123 | `:7-10` |
| `R CMD check` stops at "unsuitable version" before any test runs, so the exposure is everything that is not `R CMD check` | `:12-17` |
| DESCRIPTION states each constraint once and this is the only reader, so nothing is written twice | `:19-21` |
| it lives under `inst/` for the reason `must-error.R` does | `:21-24` |

## What this fixes

Those paragraphs are deleted here, not moved: the destination already holds
them. What stays is the rule `guard.R` says this file owns — guard through
`marginplyr_suggest_available()`, and what asserts that a guard does — stated
positively, without the list of calls to avoid.

Two paragraphs are not in `guard.R` and need deciding rather than deleting: the
skip wording `verify-backend.R` matches on, and what the two scans in
`test-documentation.R` read. Each has a file it is about.

## Acceptance

- Every deleted paragraph is accounted for: the commit names the line in
  `guard.R` that already held it;
- the two paragraphs `guard.R` does not hold are moved to the file they are
  about and cited, or kept with a stated reason;
- the always-loaded text states the positive rule and not the calls to avoid;
- `wc -c CLAUDE.md AGENTS.md` falls, and `baseline_bytes` in
  `verify-context-budget.R` is lowered **in the same commit**;
- the `depends-only` job and `verify-backend.R` still fail on what they failed
  on.

## Not in scope

*Chunks that must fail*, which was in this ticket and is now out of it. Its
target declares the opposite division: `inst/vignette-hooks/must-error.R:18-21`
says `AGENTS.md` is authoritative "for the three properties of this
implementation that are load-bearing", so those 76 lines are not a second copy
— they are where that file put them on purpose. Moving them means changing a
declared authority, which is a decision and not a cleanup, and it is the kind
of decision #282 records this repository failing to derive a rule for.

sayuks:
Closes #294.

`AGENTS.md`'s *Optional-dependency guards* restated in 79 always-loaded lines the argument `inst/suggests/guard.R:1-30` makes in its own header. `guard.R:29-30` had already settled the split — the header owns the argument, `AGENTS.md` owns the rule and what asserts it — so the section is reduced to that.

## What was deleted, and where it already was

| the argument | already held at |
|---|---|
| `requireNamespace()` answers "installed", not "new enough" | `guard.R:4-6` |
| `duckdb (>= 1.5.5)`, `shared_home = FALSE` a hard error under 1.5.4.x (#123) | `guard.R:7-10` |
| `R CMD check` stops before any test runs, so the exposure is everything else | `guard.R:12-17` |
| DESCRIPTION states each constraint once, and the guard is its only reader | `guard.R:19-21` |
| it lives under `inst/` for the reason `must-error.R` does | `guard.R:21-24` |
| guarding on a package DESCRIPTION does not suggest is an error | `guard.R:94-100` |
| the skip wording `verify-backend.R` matches on | `tests/testthat/helper-optional-backends.R:300-315`, `:228-236`, `:288-296` |
| what the two scans read, and why each source is added where it is reachable | `tests/testthat/test-documentation.R:126-141`, `:143-153`, `:167-194`, `:400-414` |

## What moved rather than being deleted

The scans paragraph held two sentences `test-documentation.R` did not: why the README is in the scanned set, now in the `readme_sources()` header, and that the scan is deliberately blunt, now in the `version_blind_guards` header — which the `cran_claims` comment below it was already citing as though it were stated there.

## Consequences

The *Suggest*/*backend* terminology note scoped a section that no longer uses either word, so it goes too; *Release matrix* is where the distinction is stated. Two references in that section followed it: it no longer claims both words are in use in *Optional-dependency guards*, and the `requireNamespace()` reason it cites now names `inst/suggests/guard.R`.

`wc -c CLAUDE.md AGENTS.md` falls 39223 → 34839, and `verify-context-budget.R`'s baseline falls with it in the same commit.

## Verification

- `Rscript .github/scripts/verify-context-budget.R` — passes at the new baseline
- full `testthat::test_local()` — passes

### DIFF
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -24,7 +24,7 @@ source(".github/scripts/ci-helpers.R")
 # What the closure measured when it was last set, and the commit that set it.
 # Moving this means saying in the same commit what was added and what paid for
 # it.
-baseline_bytes <- 34839L
+baseline_bytes <- 34826L
 baseline_note <- paste(
   "#294 deleting the argument *Optional-dependency guards* restated:",
   "`inst/suggests/guard.R` already held it, so nothing paid for this"
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -24,10 +24,10 @@ source(".github/scripts/ci-helpers.R")
 # What the closure measured when it was last set, and the commit that set it.
 # Moving this means saying in the same commit what was added and what paid for
 # it.
-baseline_bytes <- 39223L
+baseline_bytes <- 34839L
 baseline_note <- paste(
-  "#292 adding the verifier-invocation rule, paid for by nothing:",
-  "the gate it names is what makes every other one here undeletable"
+  "#294 deleting the argument *Optional-dependency guards* restated:",
+  "`inst/suggests/guard.R` already held it, so nothing paid for this"
 )
 
 entry <- "CLAUDE.md"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -317,82 +317,16 @@ as it stood.
 ### Optional-dependency guards
 
 Every guard deciding whether optional code runs goes through
-`marginplyr_suggest_available()` in `inst/suggests/guard.R`. Never
-`requireNamespace()` and never `rlang::is_installed()`, which answer "is it
-installed" and not "is it new enough" — the question DESCRIPTION actually asks,
-since several Suggests carry a version constraint. An installed-but-too-old
-package passes an installation check, the guarded code runs, and it fails at
-the feature instead of skipping (#123). `duckdb (>= 1.5.5)` is the case that
-showed it: under 1.5.4.x `duckdb::duckdb(shared_home = FALSE)` is a hard error
-rather than a degraded result.
+`marginplyr_suggest_available()` in `inst/suggests/guard.R`. That file's header
+is authoritative for why the guard exists and why it is reached by `source()`
+rather than exported.
 
-`R CMD check` is not the exposure. It stops at `checking package dependencies`
-with "required and available but unsuitable version" before any test, example,
-or vignette runs, so CRAN, the release matrix, and any local check were always
-safe. What the guard covers is everything that is not `R CMD check`:
-`devtools::test()`, `pkgload::load_all()` plus `testthat::test_local()`, an
-example run interactively, and the site build — whose `Config/Needs/website`
-entries carry no versions, so a site job can legitimately install a version the
-package's own Suggests entry rejects.
-
-*Suggest* and *backend* are used from here on in the senses *Release matrix*
-below separates them into, since the helpers this section names are shared with
-the jobs that section describes.
-
-DESCRIPTION states each constraint and the guard is the only thing that reads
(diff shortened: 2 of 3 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, both axes, against the merge-base with `main` (`467f358...b39ff58`). Six findings; each is answered in `febc0b4`.

|---|---|---|
| *Optional-dependency guards* said "Two files assert it" while describing the second file as a call site — `guard.R:29-30` makes this section authoritative for what asserts the rule, so the entry was false | Standards | fixed in prose |
| *Installation instructions* compared its bluntness to "the version-blind guard scan above", whose statement `b39ff58` moved out of `AGENTS.md` | Standards, Spec | fixed in prose |
| `b39ff58` deleted a paragraph forbidding `rlang::is_installed()` as well as `requireNamespace()`, and cited only `guard.R:4-6`, which names the latter | Spec | accounted for in `febc0b4`, which names `test-documentation.R:138-144` |
| the deleted "a version is not a third place" sentence had no entry in the accounting | Spec | accounted for in `febc0b4`, which names *Release matrix* for the rule and `guard.R:19-21` for the reason |
| `b39ff58` cited `test-documentation.R` at pre-insertion line numbers, so two of the four ranges pointed at the wrong code | Spec | corrected in `febc0b4` |
| one line left at 109 columns by the rewrap | Spec | fixed in prose; the Standards axis did not raise it, no repository document stating a width |

Verified clean by both axes: every other deleted paragraph is held at the line the commit names (all of `guard.R:4-6`, `:7-10`, `:12-17`, `:19-21`, `:21-24`, `:94-100`, and `helper-optional-backends.R:228-236`, `:288-296`, `:300-315`); `baseline_bytes` matches `wc -c CLAUDE.md AGENTS.md` in the same commit; the two moved sentences sit with their subject and are not second copies; nothing in the diff changes what the `depends-only` job or `verify-backend.R` fails on — `documentation_sources()` never reads the test file. No scope creep: both *Release matrix* edits follow from deletions that would otherwise dangle, and *Chunks that must fail* is untouched.

Not run a second time, per `design/agents/code-review.md`: a second round reads the first round's answers, and this branch has no executable behaviour the first round did not see.

PR #329, opened 2026-08-30
Move three sections' arguments out of the always-loaded closure (#295)

### LINKED ISSUE #295: *Site verification*, *Dependency metadata*, and *Installation instructions* are 171 always-loaded lines, and only one has a target that already argues them
## Problem

Three sections, 171 of `AGENTS.md`'s 642 lines, each about a branch an ordinary
commit does not enter. The most extreme is *Installation instructions*: 62
lines of release runbook for a field no commit has ever moved.

```sh
git log --all --oneline -S'cran-status: published' -- DESCRIPTION   # empty
```

They are grouped because they are three different cases, and telling them apart
is the work.

## Evidence, per section

**Installation instructions (62 lines) — mostly a second copy.**
`tests/testthat/test-documentation.R` already argues five of its six points:

| the argument | `test-documentation.R` |
|---|---|
| the markers all name marginplyr, because the README's comparison table links to another package's CRAN page | `:262-263` |
| prose needing to name the CRAN call has to spell it some other way | `:264-265` |
| matched case-insensitively, because the two spellings are one host | `:272-274` |
| the fixtures execute the `published` direction, since a branch first evaluated on release day has never run | `:287-290` |
| `published` requires both halves — the instruction and the badge | `:292-297` |

The sixth is here alone: why the cranlogs badge is a claim of availability
rather than a report of zero interest. The four-step release procedure is the
rule and stays; where it stays is the open question, since a runbook has no
home in this repository yet.

**Site verification (44 lines) — a target exists.**
`.github/scripts/verify-site.R` derives what it checks and argues its own
derivation; `altdoc.yaml` holds the rest. Whether it already argues each
paragraph is what this ticket establishes.

**Dependency metadata (65 lines) — no target.**
The case against `attachment::att_amend_desc()`, the `DBI = FALSE` closure, and
the `Config/Needs/website` constraint check are about `DESCRIPTION`, which
takes no comments. `verify-depends-only.R` enforces one part of it; the manual
audit has no gate and no owner. **This ticket decides where it goes, or that it
stays.**

## What this fixes

Paragraph by paragraph: delete where the target already argues it and name the
line; move where a target exists and does not; keep, with a stated reason,
where no target does. "It stays in `AGENTS.md`" is an admissible answer — the
reason has to be about what a task needs loaded before it knows the branch
applies, not about where the text happens to sit.

No new catch-all: `investigation/README.md`'s *Graduation* admits `AGENTS.md`
or an ADR, and `design/checks/` was refused on that ground.

## Acceptance

- Every deleted paragraph names the file and line that already held it;
- every kept paragraph states why no file it is about could hold it;
- `test-documentation.R` still asserts both directions of the `cran-status`
  field, and `verify-site.R` still fails on what it failed on;
- `wc -c CLAUDE.md AGENTS.md` falls, and `baseline_bytes` in
  `verify-context-budget.R` is lowered **in the same commit**.

## Not in scope

Flipping `Config/marginplyr/cran-status`. The procedure is being placed, not
run.

sayuks:
Closes #295.

`AGENTS.md` falls from 34815 to 28858 bytes; `baseline_bytes` in
`.github/scripts/verify-context-budget.R` falls from 34826 to 28869. Almost all
of the reduction is unpaid for: every byte removed was an argument another file
already made, or one moved to the file that owns the decision. The 165 bytes
that are paid for are the rule the review found deleted without a home, restored
in `04b4363`; the note on the baseline says so.

## Installation instructions — 62 lines to 16

Kept: the rule an ordinary commit can break — the field, its two values, the
route the documentation gives while it reads `unpublished`, and what no page may
name.

Deleted, with the line that already held it:

| the argument | already held by |
|---|---|
| the field records a fact no page can be read to check | `tests/testthat/test-documentation.R:233-238` |
| both directions are load-bearing, and `published` needs both halves | `tests/testthat/test-documentation.R:328-333` |
| the fixtures execute the `published` direction, since a branch first evaluated on release day has never run | `tests/testthat/test-documentation.R:322-326` |
| the markers all name marginplyr, and prose must spell the call otherwise | `tests/testthat/test-documentation.R:292-296` |
| matched case-insensitively, the two spellings being one host | `tests/testthat/test-documentation.R:307-311` |

Moved:

- why the cranlogs badge is a claim of availability rather than a report of zero
  interest — the one argument this section held alone — to the `cran_claims`
  vector it explains, `test-documentation.R:298-300`.
- the publication-day runbook: the milestone, the four steps, and how steps 1 to
  3 hold each other up, to `test-documentation.R:241-262`. A release knows the
  branch applies before it reads anything, so the steps do not have to be loaded
  before it; the file they move to is the one that fails when 1 is done without
  2, or 2 without 1.

## Site verification — 44 lines to 18

`verify-site.R` argues all four of its argument paragraphs, so nothing was added
there:

| the argument | already held by |
|---|---|
| the derivation and what it replaced (#114) | `verify-site.R:1-14, 29-32, 64-69, 117-122` |
| the two assertions over the derivation | `verify-site.R:70-75, 124-128` |
| markers sit on top of the baseline, a key naming no page is an error, a rendered diagnostic outlives prose | `verify-site.R:16-21, 329-333` |
| the scans, the search index, and the step summary the workflow no longer shows (#99) | `verify-site.R:140-142, 189-191, 215-223, 484-487` |

Kept: how to run it locally. `verify-site.R`'s header says what it checks and not
how to produce what it reads, and `altdoc.yaml` shows a CI sequence whose
`local::.` install is not the local ordering, so the render command and the
install-first caveat have no file to move to.

## Dependency metadata — 65 lines to 36, and it stays

The branch is an edit to `DESCRIPTION`, which is DCF and carries neither a
comment nor a citation; `att_amend_desc()` is invoked by no file here, so nothing
on the path to that edit is read on the way; and the gates that would catch the
mistake — `depends-only` and `verify-depends-only.R` — run after the commit is
made. A task has to hold the refusal and the audit procedure before it starts,
which is the test *Always-loaded context* sets.

Two parts of it did have owners:

- the `DBI = FALSE` closure argument →
  `tests/testthat/helper-optional-backends.R:154-161`. One sentence stays,
  because *Chunks that must fail* and *Release matrix* both cite this section
  for that case.
- `Config/Needs/website` is not a weaker home, how altdoc came to be in Suggests
  (#113), and the pak constraint re-check → the field's own consumer,
  `.github/workflows/altdoc.yaml:34-53`.

`Config/marginplyr/cran-status` is untouched.

A review round ran against the merge-base and is recorded in a comment below;
`04b4363` answers it.

## Verification

- `Rscript .github/scripts/verify-context-budget.R` — 28869 of 28869 bytes.
- `Rscript .github/scripts/verify-verifier-invocation.R` — 8 verifiers.
- `NOT_CRAN=true` full testthat suite — passing, `test-documentation.R`
  included.
- `lintr` on both changed R files — no lints.

### DIFF
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -24,11 +24,12 @@ source(".github/scripts/ci-helpers.R")
 # What the closure measured when it was last set, and the commit that set it.
 # Moving this means saying in the same commit what was added and what paid for
 # it.
-baseline_bytes <- 28704L
+baseline_bytes <- 28869L
 baseline_note <- paste(
-  "#295 deleting the arguments *Site verification* and two others restated,",
-  "and moving the rest to `test-documentation.R` and `altdoc.yaml`, so",
-  "nothing paid for this either"
+  "#295 deleting the arguments three sections restated and moving the rest to",
+  "`test-documentation.R` and `altdoc.yaml`; the 5957 bytes that removed paid",
+  "for restoring the rule no file held, that no package is promoted to",
+  "Imports merely to make a check pass"
 )
 
 entry <- "CLAUDE.md"
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -24,10 +24,11 @@ source(".github/scripts/ci-helpers.R")
 # What the closure measured when it was last set, and the commit that set it.
 # Moving this means saying in the same commit what was added and what paid for
 # it.
-baseline_bytes <- 34826L
+baseline_bytes <- 28704L
 baseline_note <- paste(
-  "#294 deleting the argument *Optional-dependency guards* restated:",
-  "`inst/suggests/guard.R` already held it, so nothing paid for this"
+  "#295 deleting the arguments *Site verification* and two others restated,",
+  "and moving the rest to `test-documentation.R` and `altdoc.yaml`, so",
+  "nothing paid for this either"
 )
 
 entry <- "CLAUDE.md"
--- a/.github/workflows/altdoc.yaml
+++ b/.github/workflows/altdoc.yaml
@@ -36,10 +36,8 @@ jobs:
           extra-packages: local::.
           # altdoc is declared in DESCRIPTION's `Config/Needs/website` rather
           # than in Suggests, because nothing outside `.Rbuildignore`d paths
-          # uses it -- this workflow, `altdoc/`, and
-          # `.github/scripts/verify-site.R` are the whole of it -- so a
-          # Suggests entry installed a dependency closure for a package the
-          # tarball never mentions (#113).
+          # uses it, so a Suggests entry installed a dependency closure for a
+          # package the tarball never mentions (#113).
           #
           # That is not a weaker home: pak resolves this field, and it parses
           # and enforces a version constraint written here, so a floor moved
--- a/.github/workflows/altdoc.yaml
+++ b/.github/workflows/altdoc.yaml
@@ -34,6 +34,23 @@ jobs:
       - uses: r-lib/actions/setup-r-dependencies@v2
         with:
           extra-packages: local::.
+          # altdoc is declared in DESCRIPTION's `Config/Needs/website` rather
+          # than in Suggests, because nothing outside `.Rbuildignore`d paths
+          # uses it -- this workflow, `altdoc/`, and
+          # `.github/scripts/verify-site.R` are the whole of it -- so a
+          # Suggests entry installed a dependency closure for a package the
+          # tarball never mentions (#113).
+          #
+          # That is not a weaker home: pak resolves this field, and it parses
+          # and enforces a version constraint written here, so a floor moved
+          # across loses nothing. Nothing in this repository fails if that
+          # stops holding -- a dropped constraint silently installs an older
+          # altdoc -- so re-check it rather than assume it. Put an
+          # unsatisfiable constraint in the field of a throwaway package and
+          # resolve it with `pak::pkg_deps("local::<pkg>", dependencies =
+          # list(direct = "Config/Needs/website", indirect =
+          # "Config/Needs/website"))`; it must fail naming the constraint, not
+          # resolve.
           needs: website
 
       # Before the render rather than after it. A `must_error` option that
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,66 +68,20 @@ file. Moving the pin means regenerating `README.md` in the same commit.
 
 ### Installation instructions
 
-An installation instruction is a claim about the outside world, and no file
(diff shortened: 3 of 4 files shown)


sayuks:
## Review round

`d40bfd0` (`git diff d40bfd0...HEAD`), reading `1dd42ce`. Standards sources:
`AGENTS.md`'s *Always-loaded context*, *Code comments*, *Documentation*, the
three sections this branch edits, *Chunks that must fail*, and *Release
matrix*; `design/architecture.md`'s *Answering a review* and *Repository
placement*; `design/agents/code-review.md`; `investigation/README.md`'s
*Graduation*. Spec source: #295.

Five findings. Three answered in `04b4363`, two rejected.

|---|---|---|
| *Dependency metadata* said the grep "leaves two questions to something else" and then answered the second with the grep | Standards, type 1 | fixed in prose, `04b4363` |
| `altdoc.yaml`'s moved comment enumerated altdoc's three users, and the enumeration was incomplete — `AGENTS.md`'s local render command is a fourth | Standards, type 1 | fixed in prose, `04b4363`: the enumeration is deleted, and what (#113) rests on is that nothing outside `.Rbuildignore`d paths uses it |
| the rule "no package is promoted to Imports merely to make a check pass" was deleted and `git grep` finds it in no other file | Standards type 4 / Spec | fixed in prose, `04b4363`. It was a rule and not an argument, so it is restored rather than moved; `baseline_bytes` moves to 28869 with a note, paid for by the 5957 bytes this branch removes |
| the publication-day runbook floated between two definitions in `test-documentation.R` | Standards, judgement | fixed in prose, `04b4363`: it is now inside the comment block above `cran_status_field`, the constant naming the field it flips |
| kept paragraphs state their reason in the commit message rather than in `AGENTS.md` | Spec, judgement | rejected. The reason a paragraph stays is an argument about the closure, and *Always-loaded context*'s own remedy sends an argument to the file that owns the decision or to the ticket — writing it into the closure is the byte the budget exists to refuse. *Dependency metadata* keeps the one clause a task acts on (`DESCRIPTION` is DCF and carries no comment or citation to reach the audit from); the rest is above and in `1dd42ce` |
| the new pointers ("Its *Publication day* comment", "`altdoc.yaml`'s comment on the field") are checked by nothing | Spec, soft | rejected. No citation in `AGENTS.md` is gated — `inst/suggests/guard.R`, `verify-suite-coverage.R`, and the ADR references are all unchecked — and #295 asked for none. A gate over them is a separate ticket, not this branch's |

Verified clean by both axes: all twelve `verify-site.R` citations, both
`helper-optional-backends.R` ranges, and all seven `test-documentation.R`
ranges open to text that holds the cited argument; `verify-site.R` is
unmodified; `DESCRIPTION` untouched and `cran-status` still `unpublished`;
every surviving cross-reference inside `AGENTS.md` resolves.

Issue #264, opened 2026-08-26
`print.margin_grouping_spec()` reads the kind off an object that may have none


sayuks:
## Problem

`print.margin_grouping_spec()` reads `x$type` as its first act:

```r
print.margin_grouping_spec <- function(x, ...) {
  rule <- find_grouping_kind_rule(x$type)
  name <- if (is.null(rule)) x$type else rule$constructor
  cat("<marginplyr grouping specification: ", name, ">\n", sep = "")
  invisible(x)
}
```

`$` is not a question every object a class can sit on answers, so an object carrying the class over an atomic vector or a closure raises base R's own error instead of printing:

```r
print(structure(1:3, class = "margin_grouping_spec"))
#> Error: $ operator is invalid for atomic vectors
#>   class: simpleError/error/condition

print(structure(function() 1, class = "margin_grouping_spec"))
#> Error: object of type 'closure' is not subsettable
#>   class: notSubsettableError/error/condition
```

A list carrying the class with no `type` prints, having nothing to name:

```r
print(structure(list(args = list()), class = "margin_grouping_spec"))
#> <marginplyr grouping specification: >
```

## Relationship to #262

This is the same too-early read #262 fixed in `validate_grouping_spec_early()`, at the other site that reads the kind. It was deliberately left out of that ticket's scope, which was the guard. #262 added `grouping_spec_kind()` in `R/grouping-plan.R`, which answers `NULL` where an object cannot be asked for its kind; this site does not use it.

It is not the same defect, though, and the answer is not the same. The guard has a refusal to reach. `print()` has none: it is a print method, and a print method that raises is a different problem from a validator that raises the wrong class. Whether the right answer is `grouping_spec_kind()` plus the existing empty-name fallback, or something that says the object is unreadable, is what this ticket has to decide.

## Reproduction / evidence

Measured on R 4.6.1 at `112c12c`. Pre-existing and unchanged by #262: `git show eb7ca62:R/grouping-spec.R` reads the field the same way.

## Scope

`R/grouping-spec.R`, `print.margin_grouping_spec()` only. The wording of the printed line for every object that prints today is deliberately unchanged, including the empty name above.

## Origin

Found by the Standards review of #262's branch, which noted the sibling site while judging the guard's fix.

PR #330, opened 2026-08-30
Cut the argument #268 owns from ADR 0008's amendment (#296)

### LINKED ISSUE #296: ADR 0008's #268 amendment argues alternatives the ticket rejected, which `Repository placement` sends to the ticket
## Problem

`design/architecture.md:684` says an amendment "records what the decision now is
and what moved with it. The argument for changing it is the ticket's, and the
ticket is what the amendment cites: an amendment that makes the case again is a
second copy of the ticket's, and it goes stale the way a comment re-deriving an
ADR does."

`## Amendment: the printed line for a kind that is no name` in
`design/adr/0008-centralize-grouping-specification-kind-rules.md:233` runs 93
lines — a fifth of the 473-line ADR — against #268's 21 lines of production R.
Several of its paragraphs are the ticket's argument rather than what moved.

## Evidence

Three sites, all in that amendment:

- `:271` — "Widening either to `condition` is the rewrite #268 rejects, since
  it would take a warning signalled on the way to answering for a failure to
  answer, and a test holds that direction." An alternative weighed and
  rejected. `design/agents/code-review.md` routes that to an ADR's
  *Considered options*, and this is not one; the decision it would have changed
  is #268's, whose argument the ticket owns.
- `:312` — "Rendering the field rather than dropping it is the alternative, and
  it was rejected for what it would have to promise…" Same shape, four more
  lines of it.
- `:320` and `:146` — what the change does *not* affect: what a constructor
  builds, and the constraint on evaluation count and timing. Neither moved.

## What this fixes

The amendment keeps what the decision now is and what moved with it — the
condition classes that changed, the printed line that changed, and the
superseded clauses — and cites #268 for why. The rejected alternatives and the
non-changes go, or are reduced to the sentence that says a reader will find
them in the ticket.

**Superseded text stays.** ADR 0001 carries `## Amendment: the finalizer does
not order rows` and `## Amendment: the finalizer orders rows again, on request`
side by side; deleting the first would destroy the record of why `.sort` left
and came back. This ticket removes argument the ticket owns, never a clause the
ADR superseded.

## Acceptance

- No paragraph of the amendment argues an alternative #268 rejected;
- every clause the amendment supersedes is still readable in it;
- what the decision now is, and what moved, is unchanged in substance;
- `test-grouping-plan.R` and `test-grouping-interface.R` still pass, and the
  behaviour #268 landed is still asserted.

## Not in scope

The other 22 amendments across the ADRs. Whether this is one case or a pattern
is a question this ticket's outcome is evidence for, not one it settles.

sayuks:
Closes #296.

`design/architecture.md`'s amendment rule holds that an amendment records what
the decision now is and what moved with it, and cites the ticket for why.
Three passages of ADR 0008's *the printed line for a kind that is no name*
were the ticket's argument instead.

## What changed

- The rejected alternatives — widening either catch to `condition`, and
  rendering the field rather than dropping it — reduce to one sentence naming
  #268 as where they are read. `design/agents/code-review.md` routes an
  alternative weighed and rejected to *Considered options* in the ADR holding
  the decision it would have changed; that decision is #268's, so the ticket
  is the home and not this ADR.
- The paragraph on what the change does not affect goes. What a constructor
  builds is already stated by the amendment above it, and the evaluations
  position by that amendment and again by the one below.

## What stayed

Every behaviour claim, including the clauses a later amendment supersedes:
the `tryCatch(error = )` qualification, the printed line the #264 pins now
render, and the sentence leaving the guards alone that the #280 amendment
cites. The qualification keeps what it says about a warning signalled on the
way to answering — `test-grouping-interface.R` holds that direction — with
only the rejected rewrite dropped from the sentence.

The amendment runs 80 lines where it ran 93.

## Verification

`test-grouping-plan.R` and `test-grouping-interface.R` pass, and the tests
asserting the behaviour #268 landed are untouched. Documentation only: no
file the package build reaches is changed.

### DIFF
--- a/design/adr/0008-centralize-grouping-specification-kind-rules.md
+++ b/design/adr/0008-centralize-grouping-specification-kind-rules.md
@@ -308,9 +308,6 @@ environment among them — `cat()` rendering anything of no length as nothing.
 Which of them a given object is goes unreported, for the reason the amendment
 above gives.
 
-The alternatives weighed against dropping the field, and what each would have
-had to promise, are #268's.
-
 ## Amendment: the condition class of a kind the guard could not classify
 
 The amendment above says the guards are left alone and why, and #280 is where
@@ -384,12 +381,12 @@ avoid.
 
 **Evaluations.** The constraint holding the number and timing of evaluations is
 not reached. The field is read once, as the amendments above left it. How often
-classifying asks the value's own methods is fixed by no decision, which the
-amendment above already records, and it moves in both directions: it falls at
-the guards and at the printed line, where the registry lookup's own guard asked
-them a second time and is now handed a plain string; and it is unchanged in
-number at the nested-name site, which asks `is.na()` where it did not and
-reaches `as.character()` where it no longer does.
+classifying asks the value's own methods is fixed by no decision, and it moves
+in both directions: it falls at the guards and at the printed line, where the
+registry lookup's own guard asked them a second time and is now handed a plain
+string; and it is unchanged in number at the nested-name site, which asks
+`is.na()` where it did not and reaches `as.character()` where it no longer
+does.
 
 **The handler the classification is wrapped in.** `tryCatch(error = )` a third
 time, so the trade the amendment for a specification the printer could not read
--- a/design/adr/0008-centralize-grouping-specification-kind-rules.md
+++ b/design/adr/0008-centralize-grouping-specification-kind-rules.md
@@ -268,10 +268,9 @@ constructor name the registry holds, or the kind itself, which
 The qualification is the read's as much as this one's. `tryCatch(error = )`
 catches a condition for what it says it is, so a condition raised with `stop()`
 whose class does not inherit `error` is caught by neither and reaches the
-caller. Widening either to `condition` is the rewrite #268 rejects, since it
-would take a warning signalled on the way to answering for a failure to answer,
-and a test holds that direction. Under `options(warn = 2)` a warning is an
-error and is caught, which is R's semantics for the option rather than
+caller. A warning signalled on the way to answering is not taken for a failure
+to answer, and a test holds that direction. Under `options(warn = 2)` a warning
+is an error and is caught, which is R's semantics for the option rather than
 something decided here.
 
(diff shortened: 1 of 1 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel, against
`git diff origin/main...HEAD` at `f8f2918` (merge-base `d9b5714`). Standards
sources passed: `AGENTS.md` *Code comments* and *Investigation notes*;
`design/architecture.md` *Repository placement* and *Answering a review*;
`design/agents/code-review.md`. Spec: #296, with #268 as the ticket that owns
the argument being removed.

**Spec axis — no findings.** All four acceptance criteria met. The ticket's
`:146` reference resolves into `## Amendment: the condition class of a
specification the guard could not read`, a different amendment that *Not in
scope* excludes; the sentence the ticket describes there is the one at old
`:321`–`:324`, inside the amendment, which the diff deleted. Leaving `:146`
alone is correct rather than a miss.

**Standards axis — three findings, all answered in `8c96a6c`.**

| finding | disposition |
|---|---|
| the #280 amendment's **Evaluations.** paragraph cited "the amendment above" for a paragraph `f8f2918` deleted | fixed in prose — the clause goes |
| "…are #268's" sent the reader to a ticket whose body holds no such argument | fixed in prose — the sentence goes |
| `f8f2918`'s message said what a constructor builds "is already stated by the amendment above" | fixed in prose — corrected in `8c96a6c`'s message |

The first two are category 1 of *Answering a review*: the prose is false. Each
is recorded by the diff that changed it, per `design/agents/code-review.md`.

The Fowler baseline has no site in a prose-only diff.

**Not run a second time.** `design/agents/code-review.md` says to run the skill
again only over executable behaviour the first run did not see. This branch
changes no executable behaviour, and both answers are prose deletions the diff
records.

**Verification.** Full suite green at `8c96a6c` — no failures, no skips,
including `test-grouping-plan.R` and `test-grouping-interface.R`. No test file
is in the diff, so the behaviour #268 landed is asserted by the same tests as
before.

Issue #273, opened 2026-08-27
37 comment paragraphs in `R/` name an ADR and run past six lines, where the argument is the ADR's


sayuks:
## Problem

`AGENTS.md`'s *Code comments* section says a citation names where the decision lives and the comment says which way this site falls under it. The corpus predates the rule. A paragraph that names an ADR and runs long is where the two are most likely to have merged, and there are 37 of them, holding 328 lines.

This is a candidate list, not a defect list. The section states that no scan reaches this defect, and this scan is the demonstration: it over-fires. Three of the longest are correct as written —

- `R/conditions.R:353` (13 lines) — how marginplyr's argument label diverges from dplyr's in two places and why neither divergence is observable. ADR 0022 is cited for the convention; the divergence analysis is this site's alone.
- `R/utils.R:158` (11 lines) — that `rlang::call_name()` and `rlang::call_args()` unwrap a one-sided formula, what that did to the direct-share path, and where it was measured. External behaviour the code cannot show.
- `R/summary-selections.R:612` (6 lines) — two families rather than one, because the owner differs; ADR 0019 cited and applied.

— and at least one is the defect in its plainest form:

- `R/summary-selections.R:156` (13 lines) — a description of what four assertions in three *test* files were re-pinned to, naming `test-summarize-operation.R`, `test-static-expression-analysis.R`, and `test-contextual-helpers.R`. That is a commit message that settled in a source comment, and *Code comments* names it: a comment describes the code it sits beside, in the present tense.

## Reproduce the list

Run from the repository root:

```bash
python3 - <<'PY'
import re, glob
FUNC = re.compile(r"^\s*[A-Za-z._][\w.]*\s*(<-|=)\s*function\b")
ADR = re.compile(r"ADR[- ]?\d")
blocks = []
for path in sorted(glob.glob("R/*.R")):
    lines = open(path).read().split("\n")
    cur, start = [], None
    for i, line in enumerate(lines, 1):
        if re.match(r"^\s*#([^']|$)", line):
            if start is None: start = i
            cur.append(re.sub(r"^\s*#\s?", "", line.rstrip()))
        else:
            if cur: blocks.append((path, start, bool(FUNC.match(line)), cur))
            cur, start = [], None
    if cur: blocks.append((path, start, False, cur))
paras = []
for p, s, h, c in blocks:
    cur, pstart = [], s
    for j, x in enumerate(c):
        if x.strip() == "":
[...]

PR #331, opened 2026-08-30
Cut the argument the ADRs own from R/conditions.R's comments (#273)

sayuks:
First file of #273. `R/conditions.R`'s eight scan hits are read and decided; the
other 23 paragraphs are untouched.

## What each was

| site (at `origin/main`) | category | where the deleted argument lives |
|---|---|---|
| `:12` | 1 | ADR 0023's opening paragraph |
| `:21` | 1 | ADR 0024, *What retrieval-time formatting was doing* and *What this costs* |
| `:284` | 1 (part) | ADR 0021, *No rendering decision takes part in the identity* |
| `:300` | 1 + 3 | ADR 0023, *cli becomes an Import, with its own floor* |
| `:325` | 1 | ADR 0023's paragraph putting `report_branch_warnings()` into scope |
| `:353` | 1 | ADR 0022, *dplyr's rendering is reproduced only as far as it is observable* |
| `:387` | 1 (part) | ADR 0022, two *Considered options* |
| `:420` | 1 (part) | ADR 0022, *The identification is a test against marginplyr's own rendering* |

None was category 2. Every deleted sentence was checked against the ADR that owns
it before deletion and each was already there, so no ADR gains a sentence.

`:300` is the one the ADR asked for by name: ADR 0023's *cli becomes an Import*
closes with "The availability paragraph above `written_message_lines()` stops
being true when this lands ... and is rewritten in the same commit". What it
carried beyond that was the Suggests-floor-never-binds argument, which is that
section's own two sentences.

## Two corrections to #273

**The split says `R/conditions.R` (7).** At `origin/main` the scan reports eight
hits for this file; `:12` is the extra one.

**`R/conditions.R:353` is named as correct under category 2**, "the divergence
analysis is this site's alone". It is not: twelve of its thirteen lines are in
ADR 0022's *dplyr's rendering is reproduced only as far as it is observable*,
sentence for sentence -- `error_label_named()`'s `paste0()`, `expr_as_label()`'s
two branches, the undocumented option, the pronoun deparse, "disagree in those
two places", both unobservability arguments, and the dplyr's-direction-only
argument. Only the first sentence, what `summary_argument_labels()` answers, is
this site's own. It is cut as category 1. Recorded on the issue.

## What the sites keep

The contract, the citation, and which way the site falls under it -- that
`ansi_strip()` learned `link` in cli 3.3.0 so DESCRIPTION's floor admits this
reader; that `format_error_bullets()` is needed because the count line is
appended to a rendered message rather than raised; that the `stopifnot()` in
`branch_argument_map()` guards an invariant and not a Package condition.

## The count

The scan #273 reproduces moves from **37 paragraphs / 324 lines** at
`origin/main` to **31 / 265**. `R/conditions.R` goes from eight hits to two, both
seven lines with the duplicated argument gone -- the over-fire the issue says the
scan has by construction.

## Checks

Comments only, and no roxygen among them, so `man/`, `NAMESPACE`, and `README.md`
are unaffected and no `AGENTS.md` byte budget moves.

Refs #273

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -12,20 +12,11 @@
 # It is also the interpolating entry point: `message` is a cli template a call
 # site passes unexpanded, and `env` is what lets cli evaluate the template's
 # `{}` expressions in that site's own frame. ADR 0023 is authoritative for the
-# idiom every template is authored in -- the short refusal plus `i` bullets, the
-# inline style each subject takes, `{?}` for every plural, and the rule that
-# caller-derived text is interpolated as a value and never concatenated into the
-# template. ADR 0015's boundary is untouched by that: this still owns the class
-# and the blamed call.
+# idiom every template is authored in.
 #
-# The template is expanded here, as the condition is raised, rather than by
-# `cli::cli_abort()` when the condition is read. That is what keeps a subject
-# spelled the way the caller spelled it: `cli_abort()` sets `use_cli_format`,
-# and the retrieval-time formatting collapses a run of whitespace inside an
-# interpolated value as readily as inside the template, so a column named
-# `a  b` was named `a b` in the refusal. ADR 0024 records the decision and what
-# it costs -- there is no wrapping to a reader's width any more, and styling is
-# fixed by the raising session rather than the reading one.
+# The template is expanded here, as the condition is raised, rather than when
+# the condition is read, which is what keeps a subject spelled the way the
+# caller spelled it. ADR 0024 is authoritative for that decision and its costs.
 #
 # `format_inline()` is the inline half of cli and does not consult the width, so
 # every element comes back as the one line it was written as. Names carry the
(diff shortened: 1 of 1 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` against the merge-base with `main`, `4dab04d...b6201c9`. Two axes, two findings, both rejected with evidence.

**Standards, `R/conditions.R:387` (judgement call).** That the retained ADR 0015 clause keeps the side ("an invariant rather than a Package condition") but gives a consequence where the deleted clause gave the reason -- that the constructor checked the pair, so a disagreement here means a rewrite changed the arity.

Rejected. ADR 0022's *Considered options*, under **Carry the caller's label as an attribute on each quosure**, holds it and attributes it: "so a pair whose lengths disagree cannot be built, and one that stops agreeing later is an invariant that stops the operation (ADR 0015)". The paragraph already cites ADR 0022 two sentences earlier, for the position claim that comes from the same option. Restating the reason beside the citation is the second copy #273 exists to remove.

**Spec, `R/conditions.R:289` (nit).** That `(ADR 0023)` sits after "expanding its template through `cli::format_inline()`", which is ADR 0024's decision, where ADR 0023 owns the Import-and-floor claim the sentence is about.

Rejected. The citation is true of both clauses: ADR 0023's amendment *the template is expanded when the condition is raised* states "`abort_marginplyr()` expands its template with `cli::format_inline()` and raises through `rlang::abort()`", and a reader following it there is routed to ADR 0024 by that amendment's opening line. What remains is that the sentence could be read as attaching the citation to the nearer clause, and how prose reads is not one of the four kinds *Answering a review* admits.

**Clean on everything the round was pointed at.** Both axes independently traced all eight deletions to the ADR that owns them, hunk by hunk, and neither found a sentence no ADR holds. Both confirmed the `:353` override of #273's category-2 call. The Spec axis re-ran the issue's scan on both trees and reproduced 37/324 -> 31/265 and 8 -> 2. No non-comment line changed, no `# nolint` touched, no line over 80 columns, no `#'` roxygen, and the Fowler baseline has no site in a diff with no code in it.

PR #333, opened 2026-08-30
Cut ADR 0019's own amendments from R/utils.R's comments (#273)

sayuks:
Third file of #273, after #331 (`R/conditions.R`) and #332 (`R/summary-selections.R`).

## What each was

| site (at `origin/main`) | category | where the deleted text lives |
|---|---|---|
| `:238` | 1 | ADR 0019 *Amendment: redundant parentheses are read through, in one place* |
| `:436` | 1 | ADR 0019 *Amendment: an injected name is read for the name it carries* |
| `:158` | 2, one clause 3 | nothing -- rlang behaviour no ADR holds; the clause is about #163's text |
| `:538` | 2 | nothing -- left alone |

## `:238`

Both of its sentences are the amendment's:

> `(` is the identity function, so all three writings of that call evaluate
> identically and R's own parser records the difference in the tree rather than
> in what runs.

> **One place, not one place per family.** The reading lives in the shared
> readers of `R/utils.R` -- the name, the namespace, the head, and the arguments
> -- so every analysis that reads an expression inherits it.

The site keeps that the three readers below read through the pairs, and cites
the amendment for the reading and for where it stops.

## `:436`

The same ADR's other amendment, restated down to its bolded rule:

> **The environment is discarded, and that follows from the decision above
> rather than sitting beside it.** ... so there is no lookup for a quosure's
> environment to answer.

> That is the opposite of the answer #165 reached one layer out ... **the
> environment is honoured wherever evaluation happens, and these arguments are
> never evaluated.**

The ADR is thicker than the comment was -- it also holds the Grouping
specification constructors as the case showing the rule is about the criterion
and not about ownership -- so nothing moves to it.

## `:158` -- #273 was right about this one

*Problem* names it as correct as written: "that `rlang::call_name()` and
`rlang::call_args()` unwrap a one-sided formula, what that did to the
direct-share path, and where it was measured. External behaviour the code cannot
show." That holds. No ADR carries rlang's unwrapping, and `ADR-0015` is cited
only for which side the untyped condition falls on, which is the one job a
citation has.

Unlike `R/conditions.R:353` in #331, this one survives the check. One clause
goes: "which is what the ticket's severity note does not cover" is a remark
about #163's text rather than about this code.

## `:538`, left alone

`An invariant, not a Package condition (ADR-0015)` is the form *Code comments*
gives verbatim. ADR 0015's *Consequences* asks each `stopifnot()` site to
establish that it is unreachable from the public interface, and why this one is
-- the two walks that hand a quosure here map `static_call_args()`'s arguments
one to one, and the ones that change an argument count are named `across()` and
`pick()` calls -- is what only this site can say.

## One sub-threshold paragraph, cut with its block

`:246` is in `:238`'s block and under the scan's seven-line threshold, so #273
has not measured it. "Identity here stays syntactic" is the amendment's
"identity is still syntactic", and the `get()` example is *Boundary for callable
identity*. Cut for the reason #332 cut `:152` and `:170`: a verbatim ADR copy
beside a paragraph trimmed for holding one is the defect left standing. Cutting
it moves nothing the scan reports.

Two are left, and each for a stated reason:

- `:266` (5 lines) -- "A head is unwrapped only down to a name ... `(\"sum\")(1)`
  is not a call to `sum`" is the amendment's *Where the head unwrapping stops*.
  It heads a different function, so it is outside the block this branch rewrote.
- `:419` (8 lines) -- cites no ADR, which is #273's own scope boundary.

## The count

**28 paragraphs / 237 lines** at `origin/main` -> **26 / 222**. `R/utils.R` goes
from four hits to two, both category 2.

## Checks

- `lintr::lint_package()` after `pkgload::load_all(\".\")`: no lints.

Refs #273

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -257,11 +257,9 @@ unparenthesized_value <- function(expr) {
   expr
 }
 
-# The name a head or a function reference spells. A head is unwrapped only down
-# to a name, because that is the only thing a head can be unwrapped to without
-# changing what R does with it: `("sum")(1)` is not a call to `sum` but the
-# error R raises for applying a non-function, and `(function(x) x)(1)` names
-# nothing either way. Both keep the answer they have always had.
+# The name a head or a function reference spells, unwrapped only down to a
+# name. ADR 0019's *redundant parentheses are read through, in one place*
+# names that boundary; this is the reader it stops at.
 #
 # The early return is the other half of the missing-marker rule above. This is
 # asked of an argument a caller may have left empty -- an `across()` `.fns` is
--- a/R/utils.R
+++ b/R/utils.R
@@ -164,8 +164,8 @@ assert_lazy_table <- function(x) {
 # for another shape, carrying the formula as the expression that branch then
 # reads `[[2L]]` of (#163). `rlang::call_args()` unwraps the same way, which is
 # how a formula reached the direct-share path and was computed as the share it
-# is not -- measured on 5f078ea, and the one shape of this whose misread raised
-# nothing, which is what the ticket's severity note does not cover.
+# is not -- measured on 5f078ea, and the one shape of this whose misread
+# raises nothing.
 #
 # Every analysis that reads a name recognizes a call by that name, none of them
 # recognizes `~`, and each already has an answer for a name it does not know:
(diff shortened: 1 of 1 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` against the merge-base with `main`, `6a101c8...09165d0`. Two axes, one finding answered in `89285e3`, one correction to that commit's message, and one thing worth recording because it is the ticket verifying itself.

**Standards, `:266` -- answered in code.** That the citation `09165d0` added declares ADR 0019's *redundant parentheses are read through, in one place* authoritative "for that reading and for where it stops", while the `unparenthesized_name()` header twenty lines below re-derives that amendment's *Where the head unwrapping stops* uncited, with both of the ADR's examples and its reason. The axis noted that "a different function's header" is a fact about where the scan draws a paragraph rather than about who owns the argument, which is right.

Answered: cut to the header's own contract, with the boundary cited rather than restated. The second paragraph stays -- the early return being the other half of the missing-marker rule is what only this site knows. Neither axis asked for it to stay; one called it answerable by a ticket, the other the closest verbatim ADR copy left in the file. Cutting moves nothing the scan reports, the paragraph being under seven lines.

**Correction to `09165d0`'s message.** It writes `:527` for the `rebuild_static_call()` invariant, which is that paragraph's line at HEAD among line numbers that are otherwise `origin/main`'s. It is `:538` there, as the table above has it.

**`R/utils.R:158` is category 2, and this is #273's own example verifying.** The Spec axis grepped all 26 ADRs for every claim the paragraph makes -- `call_name`/`call_ns`/`call_args`, `defused`, "not a symbol", the one-sided formula, the right-hand-side unwrapping, the direct-share misread, `5f078ea` -- and found none of it. The only ADR touching formulas is 0010, which rejects them as a `.fns` interface decision, a different fact. So of the three paragraphs *Problem* names as correct as written, `R/conditions.R:353` was wrong (corrected in #331) and this one is right. `R/summary-selections.R:612` is the third and is under the scan's threshold, so it is not on the list either way.

**What else the round confirmed.** Both axes traced every deletion to ADR 0019 and found none absent, and none where the ADR was thinner -- it is the thicker document in both cases, enumerating the five inheriting sites for `:238` and holding the Grouping specification constructors case for `:436`. Both agreed `:538` stays: ADR 0015's *Consequences* asks each `stopifnot()` site to establish its own unreachability. The Standards axis checked the tense change in `:158` and found "misread raises nothing" true of rlang today. No non-comment line changed, no `# nolint` touched, no added line over 80 columns, no roxygen, and the Fowler baseline has no site in a diff with no code in it.

PR #334, opened 2026-08-30
Cut ADR 0019, 0020, and 0023's arguments from R/share.R (#273)

sayuks:
Fourth file of #273, after #331 (`R/conditions.R`), #332 (`R/summary-selections.R`),
and #333 (`R/utils.R`).

**#273's split says `R/share.R` (4). At `origin/main` the scan reports five**:
`:1564`, `:2141`, `:2162`, `:2181`, `:2363`. `:2141` is the one the count is
missing -- the second undercount in the ticket's split, after `R/conditions.R`'s.

## What each was

| site (at `origin/main`) | category | where the deleted text lives |
|---|---|---|
| `:1564` | 1 | ADR 0019 *an injected name is read for the name it carries*, its un-injected change |
| `:2141` | 1 | ADR 0020 exemption 2 |
| `:2181` | 1 | ADR 0020 *Amendment: only an answer is remembered, so the bound is per request* |
| `:2363` | 1 (part) | ADR 0023's second and third amendments |
| `:2162` | 2 | nothing -- left alone |

## `:1564`

ADR 0019 carries it, and carries more:

> the empty argument is a symbol whose name is `""`, and `share_of_total(x = )`
> is *one* argument and empty, so it was admitted and refused one layer on as
> ... which names a summary nobody wrote -- #181's defect, in the family that
> ticket did not reach. ... Restoring `rlang::is_symbol()` on that branch was the
> alternative and was rejected

What stays is the arity contrast, which the ADR does not have and which is what
makes the empty argument reachable here at all: `share_of_total(, )` is two
arguments to the parser and is refused for the count.

## `:2141` and `:2181`

The header restated ADR 0020's second exemption down to the probe's own SQL
(`SELECT SUM('x') FROM (SELECT 1 AS z)`). The inline restated that ADR's
amendment, including the Oracle and SAP HANA cost argument and that whether an
unanswered attempt was transient cannot be read from a raised query.

## `:2363`

> prose belongs in the template, where the idiom ADR 0023 states applies to it
> and a reviewer reads it beside the refusal it completes

is ADR 0023's second amendment: "It is still written beside the call that raises
it, which is the property *Two rules are gated* asks a template for -- a constant
bound elsewhere stays refused."

The paragraph above it pointed at `abort_share_helper_position()` in this file
for the two-call shape. ADR 0023's **third amendment** holds that decision and
names that refusal as its own example, so this is the same pointer #332 removed
from `R/summary-selections.R:92`. What stays is the contrast with
`abort_share_source_type()` twenty lines up, which binds a *value* where this
would bind prose -- a site-local fact no ADR reaches.

## `:2162`, left alone

`An invariant, not a Package condition (ADR 0015)` is the form *Code comments*
gives. What four sites do with this string, that none has a default, and that
the cache write is the one that does and fails closed, is what ADR 0015's
*Consequences* asks each `stopifnot()` site to establish about itself.

## One deletion crosses #273's stated boundary

`:2172` is deleted, and #273 excludes it in as many words:

> Comment paragraphs that cite no ADR are out of scope. The rule reaches them,
> but nothing has measured them and a ticket without a list is not this one.

It cites no ADR. It is deleted anyway, and the reason is worth reviewing rather
than accepting: it is ADR 0020's amendment restated -- the dropped socket, the
permissions blip, the resuming warehouse, and `.check_share_source = FALSE` as
the only way back are that amendment's own sentences -- and it sits in the same
comment run as `:2181`, two lines above it. Leaving it would have left the
measured deletion holding nothing: one paragraph cut for restating the
amendment, the other kept restating it immediately above.

This is the shape both review axes raised in #332 (`:152`, `:170`) and #333
(`:266`), reached here through the *citation* half of the scan's filter rather
than the *length* half. The earlier three were under seven lines; this one is
eight lines and simply does not name the ADR it copies.

If that is judged wrong, the remedy is to restore `:2172` and file the
no-citation paragraphs as their own ticket, which is what #273's boundary
anticipates.

## The count

**26 paragraphs / 222 lines** at `origin/main` -> **22 / 188**. `R/share.R` goes
from five hits to one.

## Checks

- `lintr::lint_package()` after `pkgload::load_all(".")`: no lints. The two
  over-80 lines in this file are pre-existing roxygen table rows carrying
  `# nolint: line_length_linter`, which is the exception *Linting* documents.

Refs #273

### DIFF
--- a/R/share.R
+++ b/R/share.R
@@ -2340,10 +2340,10 @@ abort_share_source_dialect <- function(kinds, verdict, call) {
   # arms share is the opening of the refusal and the remedy that follows it.
   #
   # Binding the clause in R and interpolating it would remove that repetition,
-  # and it is what `abort_share_source_type()` does twenty lines up -- but what
-  # it binds there is a *value*, the class the data turned out to hold, where
-  # this clause is prose marginplyr wrote. ADR 0023's *Two rules are gated* is
-  # what refuses a template bound outside the call it raises from.
+  # and it is what `abort_share_source_type()` does -- but what it binds there
+  # is a *value*, the class the data turned out to hold, where this clause is
+  # prose marginplyr wrote. ADR 0023's *Two rules are gated* is what refuses a
+  # template bound outside the call it raises from.
   if (identical(verdict, "converts")) {
     abort_marginplyr(
       c(
--- a/R/share.R
+++ b/R/share.R
@@ -1564,11 +1564,9 @@ validate_share_direct_syntax <- function(expr, output_name) {
   # The second half of that question is not inert here, and the arity is what
   # makes it reachable: `share_of_total(, )` is two arguments to the parser and
   # is refused for the count, but `share_of_total(x = )` is *one* argument and
-  # it is empty. `rlang::is_symbol()` answered `TRUE` for it, because the empty
-  # argument is a symbol whose name is `""`, so the call was admitted and
-  # refused one layer on for an unknown preceding summary named `` -- a summary
-  # nobody wrote, which is #181's defect reached in the share family. That is a
-  # change to an un-injected spelling and is recorded as one in ADR 0019.
+  # it is empty. ADR 0019's *an injected name is read for the name it carries*
+  # records what asking it with `is_name_part()` changed here, and why the
+  # alternative was rejected.
   carried <- unwrap_injected_args(args)
   if (length(carried) != 1L || !is_name_part(carried[[1L]])) {
     # `injected_quosure_clause()` is a whole sentence assembled around a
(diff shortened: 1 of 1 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` against the merge-base with `main`, `9444d38...0131145`. Two axes, one finding answered in `d371c08`, and one adjudication this branch was opened asking for.

**Standards, `R/share.R:2343` -- answered in code.** That `abort_share_source_type()` does not sit twenty lines above the comment saying so: it is defined at `:1343` and called at `:1324` and `:2801`, all about a thousand lines away. The claim was already false at `origin/main`; `0131145` rewrote the sentence carrying it, which is what makes this the place to correct it. *Answering a review* asks a claim found false to be deleted rather than restated, so the locator goes and the contrast stays.

**Spec, the boundary crossing -- upheld.** `:2172` cites no ADR, which #273's *Scope boundaries* excludes, and the axis was asked to adjudicate the deletion on evidence rather than on the argument the body made. Its verdict is that the deletion stands, on two grounds the commit message did not have:

- The boundary's own reason is "nothing has measured them". That is defeated for this paragraph: it was measured, in the same reading that adjudicated `:2181`, and traced to ADR 0020's *Amendment: only an answer is remembered* sentence for sentence.
- The one fact in it that is this code's rather than the ADR's -- that `"refuses"` and `"converts"` are verdicts and `"unknown"` is not -- is held twice more in this file, by the headers of `share_dialect_measured_names()` and `share_dialect_verdict_names()`, neither of which this branch touches. What was deleted was a third copy.

It also confirmed the block reading the body rested on: `origin/main` `:2162-2189` is one uninterrupted comment run between the `stopifnot()` and the cache write, with `:2180`'s blank `#` the only separator.

**What else the round confirmed.** Both axes traced every deletion to the ADR that owns it and found none absent, and none where the ADR was thinner -- ADR 0019 also holds the rejected `rlang::is_symbol()` restoration, ADR 0020 the probe's SQL and the whole amendment, ADR 0023's third amendment naming this file's own helper-position refusal. `FROM DUAL` and `FROM DUMMY` survive at `probe_share_dialect()`'s header, where the scaffolding they describe is, and are held by no document outside this file, so that check mattered. The retained citations were checked against the sections they name, including that "ADR 0020's second exemption ... for the bound its amendment sets" does not point a reader at superseded text, exemption 2 itself ending "The bound stated here is amended below". Both agreed `:2162` stays as category 2. The scan undercount verified: five hits here, not the four #273's split names.

No non-comment line changed, no `# nolint` added or disturbed -- the `# nolint end` above the last hunk is untouched context -- no added line over 80 columns, no roxygen, and the Fowler baseline has no site in a diff with no code in it.

Issue #272, opened 2026-08-27
The 96-line header above `check_ambiguous_nested_name()` re-derives about half of ADR 0026's argument


sayuks:
## Problem

`R/grouping-plan.R:364-459` is a 96-line comment block above a function body of about 30 lines — the longest comment in the package. It is a header, so under `AGENTS.md`'s *Code comments* section it is `check_ambiguous_nested_name()`'s documentation, and part of it is exactly that. The rest is ADR 0026's argument written out again in different words.

Read as paragraphs:

| Paragraph | Subject | Where the meaning lives |
|---|---|---|
| 1–3 | why the refusal exists; "Available, not disagreeing"; what reading the binding costs | ADR 0026 |
| 4 | why the preflight and not the gate | ADR 0026, *Where the decision is made* — near verbatim |
| 5–7 | "More than the count moves"; ADR 0008's compatibility list; "What that costs a caller" | ADR 0026 |
| 8–12 | `is_name_part()` and the `#261` empty-argument path; the `#262` unreadable-object catch; why `rule` is the parent's own; why the memo key is sound | **here, and nowhere else** |

Paragraph 4 is the clearest case. The comment:

> It sits in the preflight rather than in the gate because the preflight runs once for a whole operation and is handed to the compilation passes, where the gate runs again on each.

ADR 0026, *Where the decision is made*:

> The preflight runs once for a whole operation and is handed to the compilation passes, where the gate runs again on each.

Verbatim overlap across the whole block is 17%, so no copy-detector reaches this. What makes it a defect is that an amendment to ADR 0026 leaves the paragraphs standing and wrong, and nothing compares the two.

## What must not be lost

Paragraphs 8–12 are not in ADR 0026 and must stay:

```bash
grep -n '#26[12]' design/adr/0026-*.md   # no matches
```

The `#261` path (`grouping_arg_spec()` binds an empty argument a line earlier and raises there, so following the rule here keeps this from becoming a second site of it), the `#262` catch and what it decides, `rule` being the parent's own and what that buys the memo — each is a fact about this call site, established after ADR 0026 was written or below the level it records.

## What to build

Split the block. Roughly 41 lines stay; roughly 55 return to ADR 0026, which already holds their content, so most of the move is a deletion rather than a transfer — check each paragraph against the ADR before deleting it, and where the ADR is thinner than the comment, the ADR is what gains the sentence.

[...]

PR #335, opened 2026-08-30
Cut what ADR 0015 and 0023 own from R/grouping-plan.R (#273)

sayuks:
Fifth file of #273, after #331, #332, #333, and #334.

**#273's split says `R/grouping-plan.R` (6). The scan reports eight** at
`origin/main`: `:180`, `:352`, `:453`, `:614`, `:912`, `:1034`, `:1185`, `:1335`.
Third undercount in the split, after `R/conditions.R`'s and `R/share.R`'s.

## The exclusion still holds, for a different reason

#273 says:

> `R/grouping-plan.R:364-459` is #272 and is excluded here; two of its
> paragraphs appear in the scan.

**#272 is CLOSED / COMPLETED**, by #277 and #278. The ticket's "two" is wrong in
the other direction from the undercounts above: at `04f093f` the scan hit
**three** paragraphs inside `364-459` -- `:364`, `:407`, and `:438`. #272 removed
the first two and kept the third, which is `:453` today, so one is in the scan
now. It is exactly what #272's *What must not be lost* named:

> The `#261` path ..., the `#262` catch and what it decides, `rule` being the
> parent's own and what that buys the memo -- each is a fact about this call
> site, established after ADR 0026 was written or below the level it records.

So `:453` stays, and the exclusion holds because the question was answered
rather than because it is open.

## What each was

| site | category | disposition |
|---|---|---|
| `:180` | 2 | left |
| `:352` | 2 | left |
| `:453` | 2 | left -- #272's settled residue |
| `:614` | 2, one clause 3 | clause cut |
| `:912` | 2 | left |
| `:1034` | 1 (part) | cut to the citation |
| `:1185` | 2 | left |
| `:1335` | 1 | cut to the citation |

Five untouched is the highest proportion in this ticket so far, which is what a
file #272 has already been through should look like.

## `:1034`

> Every other failure is an External condition and is re-raised as it arrived,
> with its own class, diagnostic, and cause.

against ADR 0015:

> External conditions propagate unchanged, and rethrows preserve the original
> condition rather than reclassifying it.

The contrast the site needs -- this report is marginplyr's own and stays
parentless, where everything else propagates -- now carries the citation. The
pointer to `abort_share_source_name()` goes with it, being what the citation
covers.

## `:1335`

Two of ADR 0023's rules. "The branch ADR 0023's `{?}` rule dissolves rather than
relocates" is that ADR's own sentence:

> A rule permitting an `if` where `{?}` is awkward relocates that idiom rather
> than dissolving it

And the structural gate reading `abort_marginplyr()`'s own argument, with its
pointer to `abort_share_helper_position()` in `R/share.R`, is the third instance
of a shape already removed from `R/summary-selections.R:92` (#332) and
`R/share.R:2358` (#334). ADR 0023's third amendment holds that decision and
names that refusal itself.

`:1329` goes with it -- same block, under the threshold, and "That difference
used to be a labels list the refusal read a singular or a plural noun out of" is
a record of what the code was.

## `:614`, one clause

What it establishes -- that each of the three `.by` sources binds every value to
`data_vars`, so no documented call reaches this `stop()` -- is what ADR 0015's
*Consequences* asks each such site to supply about itself, and is category 2.
"Before #134 one did, and reported a column the caller never wrote" is a past
change, and the sentence before it already states the rule for a future one.

## The five left alone

- `:180` -- the ADR-0005 citation says which way the site falls; unresolved fixed
  keys standing in as none, and the two checks made again against the resolved
  keys, are in no ADR.
- `:352` and `:912` -- contract plus citation. `:912` names *which two
  amendments* of ADR 0008 answer it, which is the form this ticket has been
  converging on across five files.
- `:1185` -- what is not forced and why: an unforced binding, an active binding,
  and `...`. Only this site knows it.
- `:453` -- above.

## A finding this branch does not answer: #336

#335's Spec axis was pointed at `:453` to test the exclusion rather than accept
it, and found that the paragraph re-derives **ADR 0008** -- an ADR neither ticket
checked it against. `:458-460` against ADR 0008:340-342, and `:453-454` against
ADR 0008's *The nested-name site*, a section written about this exact call site.

It is not answered here. `:453` is inside the range #273 excludes by name, and
*Answering a review* asks a finding a branch's Acceptance does not reach to
become a ticket rather than prose. Filed as #336, with what the review found and
why two passes missed it: #272 compared the block against ADR 0026 only, and
#273 excludes it because #272 owned it -- which stopped being true when #272
closed.

## The count

**22 paragraphs / 188 lines** at `origin/main` -> **20 / 171**.
`R/grouping-plan.R` goes from eight hits to six, all six deliberate.

## Checks

- `lintr::lint_package()` after `pkgload::load_all(".")`: no lints. The two
  over-80 lines in this file are pre-existing and outside every hunk.

Refs #273

### DIFF
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -619,10 +619,9 @@ compile_grouping_spec_impl <- function(preflight,
   # `grouping_name_proxy(data_vars)`, this same set; and a selection column
   # names alone cannot settle, resolved against the typed snapshot, whose
   # columns are that set on every backend. A `.by` source that could invent a
-  # name is the thing that has to justify itself here — before #134 one did,
-  # and reported a column the caller never wrote. The `.by`/`.grouping` overlap
-  # check below is the opposite case and stays a Package condition: an ordinary
-  # call reaches it.
+  # name is the thing that has to justify itself here. The `.by`/`.grouping`
+  # overlap check below is the opposite case and stays a Package condition: an
+  # ordinary call reaches it.
   unknown_by <- setdiff(.by, data_vars)
   if (length(unknown_by) > 0L) {
     stop(
@@ -1034,10 +1033,8 @@ grouping_arg_spec <- function(arg, data_vars) {
 # The refused value is read from the condition rather than by evaluating the
 # quosure again to identify it: tidyselect has already evaluated it once, and
 # ADR 0008 fixes how often a caller's quosure runs. This is marginplyr's own
-# report about a position of its own, so it stays parentless, as
-# `abort_share_source_name()` does for the same reason. Every other failure is
-# an External condition and is re-raised as it arrived, with its own class,
-# diagnostic, and cause.
+# report about a position of its own, so it stays parentless, where every
+# other failure is an External condition and propagates unchanged (ADR 0015).
 #
 # A specification stored as a function is refused from the frames instead,
 # because tidyselect refuses no function: it calls one, as the predicate form
@@ -1328,19 +1325,14 @@ resolve_column_selection <- function(arg, data_proxy, on_rename) {
 
 # One caller mistake, one diagnostic: `.grouping` and `.by` refuse a renaming
 # selection for the same reason, and only the noun and the rule it states
-# differ. That difference used to be a labels list the refusal read a singular
-# or a plural noun out of; it is now two refusals, and the resolution above is
-# handed whichever one speaks for its selection.
+# differ. Two refusals rather than one, with the resolution above handed
+# whichever speaks for its selection.
 #
 # Written out rather than kept as one template with the noun interpolated,
-# because the noun is what pluralizes. `{?s}` reads the quantity beside it, so
-# a shared template would have to choose between two noun pairs, which is the
-# branch ADR 0023's `{?}` rule dissolves rather than relocates. Writing them
-# out also keeps both inside the structural gate, which reads
-# `abort_marginplyr()`'s own argument and refuses a template bound elsewhere --
-# the shape `abort_share_helper_position()` records in `R/share.R`, where what
-# two calls share is written out in each for that reason. ADR 0023's third
-# amendment names that refusal and this pair together.
+# because the noun is what pluralizes and `{?s}` reads the quantity beside it,
+# so a shared template would have to branch on it. ADR 0023's `{?}` rule is
+# what refuses that branch, and its third amendment names this pair beside
+# `R/share.R`'s helper-position refusal as what an R branch may choose instead.
 #
 # The pairs arrive alone in an `i` bullet, per ADR 0023's condition 2: how many
 # of them there are is the caller's decision.


sayuks:
## Review record

`mattpocock-skills:code-review` against the merge-base with `main`, `962181e...0fa0111`. Two axes. One finding routed to a ticket, one correction to this body, and a verdict on the five paragraphs this branch left alone -- which is what the round was pointed at, since a paragraph wrongly left shows up in no diff.

**Spec, `R/grouping-plan.R:453` -- routed to #336, not answered here.** The axis was asked to test the exclusion rather than accept it, and found the paragraph re-derives **ADR 0008**, an ADR neither ticket checked it against:

- `:458-460`, "the two cannot disagree about what a kind is", against ADR 0008:340-342, "neither site can disagree with the other about what a kind is".
- `:453-454`, "a binding that raises when it is read is not a specification, so the selection reading stands where it raises", against ADR 0008's *The nested-name site* -- a section written about this exact call site.

It is not answered in this branch. `:453` is inside the range #273 excludes by name, and *Answering a review* asks a finding a branch's Acceptance does not reach to become a ticket rather than prose added here. Filed as #336. The reason two passes missed it is worth the ticket on its own: #272 compared the block against ADR 0026, which is what its own *Problem* is about, and #273 excludes the range because #272 owned it -- which stopped being true when #272 closed.

**Correction to this body.** It said `:453` is "one paragraph in the scan, not two", correcting #273's "two of its paragraphs appear in the scan". The correction was itself wrong: at `04f093f` the scan hit **three** inside `364-459` -- `:364`, `:407`, `:438`. #272 removed the first two and kept the third. One is in the scan now because #272 worked, not because the ticket miscounted in the direction I said. The body is corrected.

**The five leaves -- both axes agree four are correct, and the fifth is #336.** `:180` (ADR 0005 is fifteen lines and holds none of the fixed-key content), `:352`, `:912`, and `:1185` are category 2. The Standards axis singled out `:912` as the strongest form in the file: it names *which two amendments* of ADR 0008 answer it and keeps only the contract. `:453` is the one, and it is #336.

**What else the round confirmed.** Both cuts trace to the ADR that owns them, and no ADR was thinner -- ADR 0015 holds the propagation rule with "with its own class, diagnostic, and cause" being an enumeration of "unchanged" rather than new content, and ADR 0023 holds both the `{?}`-relocation sentence and the third amendment. The riskiest sentence added here checks out: the third amendment's "`R/grouping-plan.R`'s two renaming refusals" is `abort_grouping_rename()` and `abort_by_rename()`, the pair this header sits above. Both category-3 deletions are carried verbatim in the commit message. The undercount verified: eight hits at both `04f093f` and `origin/main`, against the split's six.

Two judgement calls were raised and neither is answered in code. Dropping the `abort_share_source_name()` cross-reference leaves "parentless" stated at two sites with no shared home, and "parentless" appears in no ADR, so there is nothing to cite. `R/grouping-plan.R:919`'s inline restates ADR 0008's class-off amendment, but cites #289/#317 rather than an ADR and is what *Code comments* admits as an inline -- an R behaviour the code cannot show -- so it is not a second copy beside a citation.

No non-comment line changed, no `# nolint` added or disturbed, no added line over 80 columns, no roxygen, and the Fowler baseline has no site in a diff with no code in it.

Issue #336, opened 2026-08-30
check_ambiguous_nested_name()'s surviving header re-derives ADR 0008, which neither #272 nor #273 checked it against


sayuks:
## Problem

`R/grouping-plan.R:453` is the 17-line paragraph #272 left of the ninety-six-line header above `check_ambiguous_nested_name()`. #272 checked that block against **ADR 0026** and kept this paragraph as one of the facts "established after ADR 0026 was written or below the level it records". That check was right about ADR 0026. It was never run against **ADR 0008**, and the paragraph re-derives it in at least two places.

Found by the Spec axis of #335's review round, which was pointed at this paragraph because #273 excludes it and the exclusion needed testing.

## The overlaps

`R/grouping-plan.R:458-460`:

> Both that reading and the classification below are the guard's own, so the two cannot disagree about what a kind is or about what one is readable off.

ADR 0008:340-342, *Where the classification is made*:

> One function reads the field, one classifies what the read returned, and neither site can disagree with the other about what a kind is.

`R/grouping-plan.R:453-454`:

> A binding that raises when it is read is not a specification, so the selection reading stands where it raises

ADR 0008:368-371, *The nested-name site* -- a section written about this exact call site:

> That site is not a guard and its answer is not a guard's: a kind it cannot classify is not a kind the position admits, so the column reading stands, which is the answer a binding that raises when it is read already gets.

The paragraph's `#262` clause cites the issue rather than ADR 0008's *Amendment: the condition class of a specification the guard could not read*, which is where that decision lives. Its `ADR-0015` citation is correct and does the job a citation does.

I have not read the remaining sentences against ADR 0008 -- the two above are what the review surfaced, and the point of this ticket is that nobody has.

## Why this was not caught

Two checks passed over this paragraph and neither could see it:

- **#272** compared the block against ADR 0026, which is the ADR its own *Problem* section is about. ADR 0008 was not in its frame.
- **#273** excludes `R/grouping-plan.R:364-459` by line range, on the ground that #272 owns it. #272 is now CLOSED / COMPLETED, so nothing owns it.

The scan #273 reproduces does hit this paragraph -- it is one of the eight in `R/grouping-plan.R` -- so it is on that ticket's list and off its scope at the same time.

## What to build

[...]

PR #337, opened 2026-08-30
Read the last eight of #273's list, and cut four (#273)

### LINKED ISSUE #273: 37 comment paragraphs in `R/` name an ADR and run past six lines, where the argument is the ADR's

sayuks:
The six files with one or two hits each. **This finishes #273's list**, after
#331, #332, #333, #334, and #335.

## What each was

| site (at `origin/main`) | category | disposition |
|---|---|---|
| `R/contextual-helpers.R:1` | 1 | cut to the citation |
| `R/contextual-helpers.R:36` | 1 | cut to the citation |
| `R/factor.R:41` | 2 | left |
| `R/grouping-adapter-union.R:273` | 1 + self-copy | defers to the guard |
| `R/grouping-adapter-union.R:302` | 1 (part) | cites |
| `R/grouping-spec.R:252` | 2 | left |
| `R/margin-label.R:268` | 1 | cut to the citation |
| `R/margin-operation.R:389` | 1 | cut to the citation |

## `R/contextual-helpers.R`

Both headers are ADR 0019's. `:1` restated *The recognized spellings live in one
registry*:

> Before this decision every reader carried a spelling and a test of its own,
> and one of them -- `where`, in `contains_selection_predicate()` -- had no
> namespace test at all; `investigation/contextual-helper-execution-mechanism.md`
> tabulates them.

`:12`, the language-capture paragraph beside it, is the rest of that same
section ("Putting them under one reader would put two rules under it") and goes
with it -- same block, cites no ADR, the `R/share.R:2172` case from #334.

`:36` restated *The rule* and *What the criterion decides*, including that "a
caller who binds `tibble` gets their own function" and that the constructor gate
"decides *whether the argument is evaluated at all*".

## `R/margin-label.R:268`

ADR 0023's third amendment was written **about this refusal**, and the paragraph
restated it:

> The two arms here name different subjects -- the one colliding label, which is
> interpolated, against the words `Margin labels` standing in for several
> distinct ones, which is not -- so no single template can write both.

`:278`'s structural-gate sentence is the same amendment's and goes with it. What
stays is what the shape costs at this site: the columns bullet four times, the
noun four times, each remedy twice.

## `R/margin-operation.R:389`

ADR 0018's ***`arrange()` writes the key into two places***, both paragraphs,
down to `window_order()` accepting "a bare column name or `desc()` of one and
rejects everything else".

## `R/grouping-adapter-union.R` -- a copy of itself, not of an ADR

`:273` and `:302` are the same argument written twice, thirty lines apart in one
file, and `:273` said so: "the comment on the guard itself is where that sits".
Both also re-derived ADR 0015's predicate --

> The predicate is avoidability, not the origin of the cause.

-- as "ADR 0015 sorts a condition by what the caller can do about it rather than
by what happened". The header now defers to the guard for which way each arm
falls, and the guard cites.

This is the one shape in the ticket that is not ADR duplication. *Code comments*
reaches it the same way: the second copy reads as support for the first while
nothing compares the two.

## The two left alone

- `R/factor.R:41` -- ADR 0015's form over a call chain only this site knows: every
  Margin verb calls `validate_margin_operation()` before the
  `finalize_margin_operation()` that reaches here, so the `stopifnot()` is
  unreachable, which is what that ADR's *Consequences* asks each such site to
  establish. The ADR 0020 citation beside it says which way the declared-level
  collision falls.
- `R/grouping-spec.R:252` -- names which ADR 0008 amendment answers it and how it
  falls here (the empty string). The form the whole ticket converged on, and the
  same as `R/grouping-plan.R:912`.

## The count, corrected

**37 paragraphs / 328 lines** at `04f093f` -> **15 / 126** here.

328 is the figure at `04f093f`, which is the commit the list was taken at. Every
pull request in this ticket, this one included, wrote **324** against it. 324 is
the count at `4dab04d`, where these six branches started; the set of 37 is the
same at both. Corrected in `1a0f4a7`.

All thirty-seven have been read and decided. **What survives is not "the ones
left alone"** -- an earlier draft of this body said so and it is false. Of the
fifteen:

| | sites |
|---|---|
| byte-identical to a paragraph on the original list | `R/factor.R:41`, `R/grouping-plan.R:180`, `R/share.R:2155`, `R/summary-selections.R:718`, `R/utils.R:525` |
| never on the list -- written after `04f093f` | `R/grouping-plan.R:352`, `:911`, `:1182` |
| cut by one of the six branches and still over seven lines | the rest |

The filter is a length heuristic, and that is the honest reading of the residue:
a paragraph it still reports may be one nobody touched, one this ticket
rewrote, or one written after the list was taken, and the number alone does not
say which.

## One of the 37 recorded nowhere until now

`R/grouping-plan.R:506` at `04f093f` is the `admitted_nested_kinds()` header, and
it is on the list. `526b2c6` removed it -- **#272's** work, one commit before
these branches started -- so it is decided, and no #273 pull request said so.
Recorded here.

## The one it did not settle

`R/grouping-plan.R:453` is on the list and outside the scope, because #273
excludes its range as #272's and #272 has closed. #335's review found it
re-derives ADR 0008, which neither ticket checked it against. Filed as **#336**,
open.

## Checks

Closes #273

### DIFF
--- a/R/contextual-helpers.R
+++ b/R/contextual-helpers.R
@@ -23,9 +23,9 @@
 # `test-contextual-helpers.R` asserts the laziness rather than trusting it.
 #
 # `contextual` is what ADR 0019's criterion decides, and it is not a synonym
-# for "read statically": three of the four families here are read statically
+# for "read statically": the families carrying it `FALSE` are read statically
 # and are still ordinary names. That ADR's *What the criterion decides* is
-# authoritative for which family the field is true of and why.
+# authoritative for which and why.
 static_spelling_rules <- function() {
   list(
     grouping = function() {
--- a/R/contextual-helpers.R
+++ b/R/contextual-helpers.R
@@ -2,20 +2,9 @@
 # the three families read alongside them, keyed by the family that decides what
 # recognition does. It is the one place those names and the namespaces they may
 # carry are written down, and the one place their namespace test is
-# implemented. Before #172 each reader carried a spelling and a test of its
-# own, two derived from a rules table and the rest written out, and one --
-# `where`, in `contains_selection_predicate()` -- had no namespace test at all,
-# so a spelling could be recognized by the analysis and missed by the rewrite
-# beside it. `investigation/contextual-helper-execution-mechanism.md` counts
-# and tabulates them (#172, ADR 0019).
-#
-# The language-capture primitives are the one set of names read statically and
-# deliberately kept out. `language_capture_formal()` and
-# `captured_call_parts()` (`R/utils.R`) carry `quote`, `substitute`, and
-# `expression` with a `base` namespace test of their own, and that test is not
-# this one: a capture is refused where the head names a binding the analysis
-# can see, so it answers to the calling environment where every family here
-# refuses to. Registering them would put two rules under one reader.
+# implemented. ADR 0019's *The recognized spellings live in one registry* is
+# authoritative for why one place, for what it replaced, and for the
+# language-capture primitives it keeps out.
 #
 # A family names its own spellings only where no other table already owns them.
 # `share_kind_rules()` is the authoritative description of a contextual share
@@ -34,14 +23,9 @@
 # `test-contextual-helpers.R` asserts the laziness rather than trusting it.
 #
 # `contextual` is what ADR 0019's criterion decides, and it is not a synonym
-# for "read statically". A Contextual helper's meaning arises only through
-# static rewriting, so no caller binding can change what the verb does with it.
-# The other three families are read statically and are still ordinary names: a
-# Grouping specification constructor's spelling decides only whether a nested
-# argument is evaluated at all -- which is what separates a nested
-# specification from a tidyselect selection -- and a data-frame constructor's
-# spelling is read only to predict the output names of a data-frame-valued
-# summary. Both really run whatever the name is bound to.
+# for "read statically": three of the four families here are read statically
+# and are still ordinary names. That ADR's *What the criterion decides* is
+# authoritative for which family the field is true of and why.
 static_spelling_rules <- function() {
   list(
     grouping = function() {
--- a/R/grouping-adapter-union.R
+++ b/R/grouping-adapter-union.R
@@ -297,12 +297,9 @@ summarize_margin_branch <- function(.data,
   )
 
   # A lazy input whose branch came back local was read, and both arms of this
-  # are that one fact. Which arm depends on who did the reading (ADR 0015): an
-  # Arrow input absorbing is something the caller can rewrite, and the refusal
-  # names the rewrites; any other backend answering a lazy input with a local
-  # frame is a defect here or there, which no rewrite of their call avoids.
-  # Attributing the second to Arrow would be worse than not reporting it -- a
-  # diagnostic that misdirects, over a defect.
+  # are that one fact. Which arm depends on who did the reading, and ADR 0025's
+  # *How the refusal is raised* is authoritative for why each falls where it
+  # does.
   #
   # The second arm is also what catches a branch list that is part local and
   # part lazy, which `combine_margin_branches()` does not: a lazy-first mix is
--- a/R/grouping-adapter-union.R
+++ b/R/grouping-adapter-union.R
@@ -271,13 +271,10 @@ abort_absorbed_summary <- function(labels) {
 # expression once per grouping set and this stops at the first.
 #
 # Over an Arrow input the guard raises the refusal rather than an internal
-# invariant (ADR 0015), although what it reports is a defect. A caller reaching
-# it can act on it, and the action is the one the refusal already names; making
-# them read a bug report instead would move the cost of marginplyr's drift onto
-# them. It has a second arm, for every other lazy backend, where that reasoning
-# inverts and an invariant is what the caller should get; the comment on the
-# guard itself is where that sits, and ADR 0025 records both. The maintainer's
-# signal is `test-query-policy.R`, which fails when a read happens at all.
+# invariant, and its second arm inverts that for every other lazy backend.
+# Which way each falls is on the guard itself, and ADR 0025 records both. The
+# maintainer's signal is `test-query-policy.R`, which fails when a read
+# happens at all.
 summarize_margin_branch <- function(.data,
                                     ...,
                                     .by,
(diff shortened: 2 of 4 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` against the merge-base with `main`, `408a0fc...21bbe08`. Two axes. One hard violation, two judgement calls, all answered in `1a0f4a7`, plus three corrections to what this ticket has been recording.

**Standards, `R/contextual-helpers.R:36` -- hard violation, answered.** The rewrite made a true sentence false. The original said "The other three families are read statically and are still ordinary names", which is true: `static_spelling_rules()` has **eight** families, five carrying `contextual = TRUE` and three not. Summarising it as "three of the four families here" invented a count the table does not have, and contradicted this same header's own opening line. It now names the families carrying `FALSE`, which is checkable from the table and does not drift when a family is added.

This is a different failure from the five rounds before it. Those were cuts that left an ADR's argument standing, or citations pointing at the wrong section. This one is a *summary that introduced a claim the deleted text did not make* -- and nothing had checked it against the implementation.

**Standards, `R/grouping-adapter-union.R:302` -- judgement call, answered.** Deduplicating the header against the guard left the guard re-deriving **ADR 0025**: "A diagnostic that misdirects is worse than the bug report ADR 0015 asks for" is that ADR's own sentence at `0025:127`, and the guard cited only ADR 0015 while the header said ADR 0025 records both arms. The guard now cites it.

**Standards, `R/margin-operation.R` -- judgement call, answered.** The new citation named ADR 0018's bolded lead-in *`arrange()` writes the key into two places*, which holds which of the two is a Margin order and nothing else; why clearing it takes nothing away is the next bolded paragraph and the `ORDER BY` is a consequence below both. It now cites the amendment holding all three.

**Spec: `Closes #273` is justified.** All five *Acceptance* bullets verified independently, including the first end to end -- all 37 sites at `04f093f` enumerated and each traced to a branch that changed it or a body that says it was left. Three corrections came out of that:

- **`04f093f` reports 328 lines, not 324.** 324 is `4dab04d`, where these branches started. Every pull request in this ticket wrote 324 against `04f093f`. The set of 37 is the same at both.
- **`R/grouping-plan.R:506` is one of the 37 and no #273 pull request named it.** It is the `admitted_nested_kinds()` header, removed by `526b2c6` -- #272's work, one commit before these branches. Decided, never recorded as decided here.
- **The survivors are not "the ones left alone".** Of the fifteen the scan now reports, five are byte-identical to a paragraph on the list, three were written after `04f093f` and were never on it, and the rest were cut by one of the six branches and still measure over seven lines. This body's earlier "every one of the sixteen was left alone deliberately" was false, and its 43% was derived from it.

All three are corrected in `1a0f4a7`'s message and in the body above.

**#336 satisfies the first bullet rather than blocking it.** `R/grouping-plan.R:453` was read, decided, and kept; the ADR 0008 overlap #335's review found sits outside #273's declared `364-459` scope, and *Answering a review* asks such a finding to become a ticket. #273's boundary is respected by not editing it here.

No non-comment line changed anywhere in the nine commits this ticket contributed, `tests/` was never touched, and `R/grouping-plan.R:364-459` was not edited in any of the six branches.

PR #344, opened 2026-08-31
Order a one-occurrence Grouping plan on a lazy backend

### LINKED ISSUE #339: Emit the Margin order's ORDER BY for a one-occurrence Grouping plan on a lazy backend
## What to build

A caller who asks for a Margin order on a lazy input gets the order the reference
promises, whatever shape the Grouping plan has.

Today the order is dropped, silently, whenever the compiled plan holds exactly one
grouping-set occurrence. The result carries no outermost `ORDER BY`, so the rows arrive
in whatever order the backend chose. Two verbs reach it:

- `expand_with_margins()` on every dbplyr backend, since expansion always takes the
  portable path. Confirmed on DuckDB, SQLite, and the PostgreSQL simulator.
- `summarize_with_margins()` on a backend without native `GROUPING SETS` — SQLite and
  the generic SQL kind. A plan with the native path is unaffected.

Plans reaching it include the everyday `.grouping = NULL` with `.by`, and
`grouping_spec(x)`, `grouping_sets(grouping_set(x))`, and `.duplicates = "drop"`
collapsing two identical sets into one.

What the caller sees instead of a diagnostic is dbplyr's own
`ORDER BY is ignored in subqueries without LIMIT`, raised at collection, naming
`arrange()` — a verb they never wrote — and not naming `.sort`.

Two observations bound the cause. A plan with two or more occurrences composes its
branches with `union_all()`, which breaks the query into levels, and its order survives.
And when `.id` names the same column the ordering key reads, the order survives too,
because the finalization then has no column to drop after ordering. The difference is
where the ordering key's column is materialized relative to the ordering itself.

## Acceptance criteria

- [ ] `summarize_with_margins()` and `expand_with_margins()` with `.sort` other than
      `"none"` produce an outermost `ORDER BY` on a lazy backend for a plan holding one
      grouping-set occurrence, for every spelling that produces one.
- [ ] Collected rows match the local result for those calls, including the case where a
      missing value must sort last against the dialect's own opposite default.
- [ ] No dbplyr `ORDER BY is ignored in subqueries` warning reaches a caller from a
      Margin verb.
- [ ] `expect_margin_order_agrees()` covers a one-occurrence plan, so the gap that let
      this pass every backend job is closed rather than fixed case by case. The helper's
      existing scenarios are all `rollup()` plans, which is why nothing failed.
- [ ] Plans holding two or more occurrences keep the order they have today, and a plan
      that ran on the native `GROUPING SETS` path still runs there — asking for a Margin
      order must still cost no native plan.
- [ ] `.sort = "none"` still records no order and clears none.

## Blocked by

None — can start immediately.

sayuks:
Closes #339.

## What changed

A Grouping plan holding exactly one grouping-set occurrence lost its outermost `ORDER BY` on a lazy backend. The staged Grouping set identifier is a column the same query computes there, so the projection that dropped it after the ordering wrapped that query, and dbplyr discarded the ordering — surfacing as `ORDER BY is ignored in subqueries without LIMIT`, naming an `arrange()` the caller never wrote.

Such a plan reads no identifier at all: every Grouping bit is constant over one occurrence, and the tiebreak orders by a column with one value. `margin_order_terms()` now omits the tiebreak on those plans, and `order_margin_result()` runs the projection before the ordering whenever the key does not read the identifier, so no query level is added after the `ORDER BY`.

Plans holding two or more occurrences are unchanged, and which adapter runs is still decided before the identifier is allocated — a one-occurrence plan that ran on the native `GROUPING SETS` path still runs there.

ADR 0018 gains an amendment recording the refinement to its "resolvable in the `FROM` clause" rule.

## Tests

- `expect_margin_order_agrees()` gains three one-occurrence scenarios — a `grouping_sets(grouping_set(x))` summary, a `.by`-only summary, and an expansion — and now asserts that collecting raises no warning. Its existing scenarios are all `rollup()` plans, which is why every backend job passed.
- Rendered-SQL tests cover every spelling that compiles to one occurrence (`.by` alone, `grouping_spec()`, `grouping_sets(grouping_set())`, `.duplicates = "drop"`) for both verbs on the portable path, the native plan on the PostgreSQL simulator, and `.sort = "none"` recording no order.

Local: `testthat::test_local()` 5750 pass, 0 fail, 0 skip; `lintr::lint_package()` clean; `verify-suite-coverage.R` exits 0.

### DIFF
--- a/R/margin-operation.R
+++ b/R/margin-operation.R
@@ -369,6 +369,21 @@ order_margin_result <- function(operation, result, execution) {
     length(operation$plan$dimensions) == 0L || !is.null(sort_id)
   )
 
+  staged_id <- if (identical(sort_id, operation$set_id_name)) {
+    NULL
+  } else {
+    sort_id
+  }
+
+  # A staged identifier the key does not read is dropped before the ordering
+  # rather than after it, so that the projection adds no query level for dbplyr
+  # to discard the `ORDER BY` from (ADR 0018).
+  if (!is.null(staged_id) && !margin_order_reads_set_id(operation$plan)) {
+    result <- dplyr::select(result, -dplyr::all_of(staged_id))
+    sort_id <- NULL
+    staged_id <- NULL
+  }
+
   terms <- margin_order_terms(
     plan = operation$plan,
     sort = operation$sort,
@@ -377,11 +392,8 @@ order_margin_result <- function(operation, result, execution) {
   if (length(terms) > 0L) {
     result <- dplyr::arrange(result, !!!terms)
   }
-  if (
-    !is.null(sort_id) &&
-      !identical(sort_id, operation$set_id_name)
-  ) {
-    result <- dplyr::select(result, -dplyr::all_of(sort_id))
+  if (!is.null(staged_id)) {
+    result <- dplyr::select(result, -dplyr::all_of(staged_id))
   }
   forget_margin_window_order(result, backend = operation$backend)
 }
@@ -442,10 +454,17 @@ margin_order_terms <- function(plan, sort, sort_id) {
     )
   }
 
-  if (!is.null(sort_id)) {
+  if (!is.null(sort_id) && margin_order_reads_set_id(plan)) {
     terms <- c(terms, list(margin_column_pronoun(sort_id)))
   }
   terms
+}
+
+# Whether the key reads the Grouping set identifier. A plan holding one
+# occurrence gives every row the same one, so neither a Grouping bit nor the
+# tiebreak has anything to order by (ADR 0018).
+margin_order_reads_set_id <- function(plan) {
+  length(plan$set_ids) > 1L
 }
 
 # One column's missingness term. Written as a comparison rather than as the
--- a/R/margin-operation.R
+++ b/R/margin-operation.R
@@ -369,33 +369,37 @@ order_margin_result <- function(operation, result, execution) {
     length(operation$plan$dimensions) == 0L || !is.null(sort_id)
   )
 
-  staged_id <- if (identical(sort_id, operation$set_id_name)) {
-    NULL
-  } else {
-    sort_id
-  }
+  # The identifier the key reads, and `NULL` where it reads none (ADR 0018).
+  # Both the terms and the position of the projection below derive from this
+  # one variable, so no term can name a column the projection has taken away.
+  key_id <- if (length(operation$plan$set_ids) > 1L) sort_id else NULL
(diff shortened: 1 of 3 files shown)


sayuks:
## Code review record

Round run with `mattpocock-skills:code-review` against `git diff main...HEAD` at `75dafae`, both axes. Answered in `db0743b`.

### Standards — 4 findings, all answered in code

| finding | disposition |
|---|---|
| `margin_order_reads_set_id()`'s header re-derived ADR 0018's own argument (`AGENTS.md`, *Code comments*) | fixed in code — the predicate is gone; `key_id` in `order_margin_result()` is the one place the fact is written |
| the site comment appended the ADR's reason after stating which way the site falls | fixed in prose — trailing clause deleted |
| one fact written three times, including a `sort_id <- NULL` reset a reader had to prove was in step | fixed in code — both the ordering terms and the projection's position derive from `key_id`, so no term can name a column the projection took away; `margin_order_terms()` is back to what `main` has |
| `staged_id` and `sort_id` hold one value under two meanings, with the `.id` fact uncommented | fixed in code — extracted `margin_staged_sort_identifier()`, whose header carries the contract |

### Spec — 4 findings

| finding | disposition |
|---|---|
| *"for every spelling that produces one"* — only single-dimension one-occurrence plans were covered; a two-dimension one reproduced the bug pre-fix, and `.by` alongside `.grouping` was untried | fixed in code — `two_dimensions` and `by_and_grouping` join `margin_order_single_set_specs()`; `expect_margin_order_agrees()` gains `single_set_dimensions` and its expansion scenario becomes two-dimensional |
| *"`.sort = "none"` still records no order and clears none"* — only `summarize_with_margins()`, and only the first half | fixed in code — the render loop covers both verbs, and a second test holds the other half: a window ordering the input carried survives a call that asks for no order |
| scope: the tiebreak is now omitted for `.id` on a one-occurrence plan too, untested | fixed in code — a test compares the rendered `ORDER BY` with and without `.id` and finds them equal |
| `margin_order_reads_set_id()` was named for the key and implemented as the occurrence count, so a future key change could desynchronise the gate from the term it guards | fixed in code — same answer as the third Standards finding |

The Spec axis verified criteria 2, 3 and 4 by running the branch's test file against `main`'s `R/margin-operation.R`: 28+ failures pre-fix, so they are asserted by tests rather than by construction. Re-checked after the answers: 38 failures pre-fix, 0 post-fix.

Local after `db0743b`: `testthat::test_local()` 5803 pass / 0 fail / 0 skip; `lintr::lint_package()` clean.

PR #345, opened 2026-08-31
Refuse a selection the names settle before the typed snapshot

### LINKED ISSUE #342: Answer the name-only selection reader from what tidyselect settles from names alone
## What to build

`is_name_only_expr()` answers one question, and that question is the whole
specification:

> Is tidyselect's evaluation of this expression — success or failure alike —
> settled by the column names alone?

Every construct is classified by that sentence, and anything the reader cannot
answer for stays `FALSE`, so the typed snapshot resolves it. Today three
classes of expression answer `FALSE` that the sentence settles as `TRUE`, and
each of them makes preparation send the zero-row selection-proxy query before
reporting an error the names already decided. That is the ordering ADR 0005
forbids.

1. **`/`**, tidyselect's set-difference operator. Its walk descends into `/`
   exactly as it descends into `c()`, `:`, `!`, `-`, `|`, and `&`, so a `/`
   whose operands are name-only is name-only. Measured against a DuckDB
   connection, an unknown column named through `all_of()` reports locally with
   no query sent, while the same unknown column reached through
   `everything() / zzz` sends one first. Both `.grouping` and `.by` have this.

2. **The spellings tidyselect refuses outright** — `*`, `^`, `&&`, `||`. Their
   failure depends on nothing at all, so a fortiori not on column types.
   `grouping_set(g * v)` sends the query before reporting the arithmetic
   refusal. These must answer `TRUE` *without descending*: tidyselect refuses
   them before walking anything under them, so `where(is.numeric) * a` fails on
   `*` and never reaches the predicate.

3. **`one_of()`**, absent from the leaf-helper list. It reads names and nothing
   else, and `one_of(1)` is refused for its input type before any name is
   consulted. It is superseded rather than deprecated, and tidyselect keeps it.

`where()` stays `FALSE`: it is the one helper that reads column data.

## The two readers

`is_name_only_expr()` and `is_grouping_spec_predicate()` do not describe the
same grammar and must not be made to enumerate the same set. They take
different subsets of one fact — tidyselect's `walk_data_tree()` dispatch table
— because they answer different questions, and each adds what its own position
requires:

| | descends-into set | `(` | `*` `^` `&&` `\|\|` | leaf helpers |
|---|---|---|---|---|
| `is_name_only_expr()` | needs | **needs** — it reads the caller's expression as written, which is why `grouping_set((zzz))` already reports locally | needs, for the reason above | needs |
| `is_grouping_spec_predicate()` | needs | **must not have** — `nested_arg_expr()` has already dropped a redundant pair | needs, for its own reason: a specification nested under a refused spelling reaches it too | no use for them — it finds helpers by frame inspection |

What the two share is the descends-into set. Give that set one name both read,
and let each site add its own extras.

## Acceptance criteria

- [ ] `is_name_only_expr()` carries a header stating the sentence above. It is
      the only AST reader in the module without one.
- [ ] A `.grouping` or `.by` selection using `/`, `*`, `^`, `&&`, `||`, or
      `one_of()` whose failure is settled by names raises before any query
      reaches the backend, for both arguments.
- [ ] The refused spellings answer without descending into their operands.
- [ ] The descends-into operator set is named once and read by both readers;
      neither list is defined as the other.
- [ ] A selection carrying a predicate still resolves against the typed
      snapshot, as it does today — including one under `/`, as in
      `everything() / where(is.numeric)`.
- [ ] A test counts backend reads rather than only asserting the diagnostic.

## Out of scope

Two shapes the same invariant settles as `TRUE` that this reader cannot
reach, each held out by a decision living elsewhere: a formula, and `.data$x`.
Both are #346.

## Blocked by

None.

sayuks:
Closes #342.

`is_name_only_expr()` answers whether tidyselect settles an expression from the
column names alone, so a failure the names decide is raised before typed
metadata is acquired (ADR 0005). Three classes of expression answered `FALSE`
that the question settles as `TRUE`, and each made preparation send the
zero-row selection-proxy query before reporting an error the names had already
decided.

- **`/`** — tidyselect walks it as a set difference, in `eval_slash()`, exactly
  as it walks `c()`, `:`, `!`, `-`, `|`, and `&`. Its `language.Rd` records no
  entry for it, which is why `selection_walk_operators()` says where the fact
  is read.
- **`*`, `^`, `&&`, `||`** — refused outright, so their failure depends on
  nothing. They answer without descending: tidyselect refuses them before
  walking anything under them, so `where(is.numeric) * a` fails on `*` and
  never reaches the predicate.
- **`one_of()`** — reads names and nothing else, as `any_of()` does, and
  `one_of(1)` is refused for its input type before a name is consulted. It is
  superseded rather than deprecated, so tidyselect keeps it.

`where()` still answers `FALSE`, including under `/`, so a selection carrying a
predicate resolves against the typed snapshot as it did.

## The issue asked for something else

#342's second acceptance criterion asked that `is_name_only_expr()` and
`is_grouping_spec_predicate()` enumerate the same operators. That would be
wrong, and the issue has been rewritten to say so. The two are not copies of
one list: they take different subsets of tidyselect's dispatch table because
they answer different questions. `is_name_only_expr()` reads the caller's
expression as written and needs `(` — which is what makes
`grouping_set((zzz))` report locally today — while
`is_grouping_spec_predicate()` is reached through `nested_arg_expr()`, which
has already dropped the pair. What they share is the two operator sets, now
named once and read by both.

The issue also offered "no, deliberately" as a valid close for `one_of()`.
That is not available: `one_of(1)` is a locally settled failure that reached
the backend first.

## Tests

Two blocks in `tests/testthat/test-grouping-plan.R`:

- the reader's classification, shape by shape, including `/` over a predicate;
- the ordering, counting calls to `grouping_selection_proxy()` — the one place
  typed metadata is acquired — for `.grouping` and `.by` alike, with a
  predicate selection as the positive control that still reads once.

The count is taken through `local_mocked_bindings()` against a local data
frame, so it needs no optional backend and asserts the rule rather than one
backend's behaviour.

## Out of scope

Two shapes the same invariant settles as `TRUE` that this reader cannot reach —
a formula, and `.data$x` — are #346. Each is held out by a decision living
elsewhere rather than by the operator sets this changes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -153,6 +153,15 @@
   returned the right answer, having read the whole input to get it — so on
   those versions this is a breaking change, and the rewrites the refusal names
   reproduce the old result while letting you choose what is read.
+* A `.grouping` or `.by` selection whose failure the column names already
+  settle is now refused without querying your input. This reaches the
+  set-difference operator `/` that tidyselect reads, the arithmetic and
+  scalar-boolean spellings it refuses in a selection — `*`, `^`, `&&`, and
+  `||` — and `one_of()`. Each of these used to send the zero-row query first
+  on a backend that needs one, so on a disconnected or slow connection you
+  were handed the connection's failure instead of the column diagnostic. A
+  selection carrying `where()` still resolves against your input's column
+  types, including one written under `/`.
 * Every error marginplyr raises for a correctable call now inherits the
   `"marginplyr_error"` class, so `tryCatch(marginplyr_error = )` catches them
   all. It is the only promised class; narrower subclasses and message wording
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -185,7 +185,11 @@ prepare_grouping_plan <- function(.data,
         # against the resolved keys below. Withholding the pass instead would
         # make a `.grouping` error determinable from names alone wait for a
         # backend read, which ADR-0005 forbids.
-        compile_grouping_spec(
+        # Discarded with the plan it compiles, so a warning it signals is one
+        # the canonical pass signals again from the same names. Suppressing
+        # them is what keeps this pass invisible but for the failure it exists
+        # to raise; `one_of()` is where that is observable.
+        suppressWarnings(compile_grouping_spec(
           grouping_spec,
           data_vars = data_vars,
           data_proxy = grouping_name_proxy(data_vars),
@@ -193,7 +197,7 @@ prepare_grouping_plan <- function(.data,
           .duplicates = .duplicates,
           duplicates_choices = duplicates_choices,
           preflight = preflight
-        )
+        ))
       }
       data_proxy <- grouping_selection_proxy(data, backend = backend)
       if (is.null(by)) {
@@ -295,8 +299,8 @@ validate_nested_grouping_units <- function(parent, nested) {
 
 # The operators tidyselect's walk combines a selection from. `/` is one:
 # tidyselect reads it as a set difference in `eval_slash()`, which its
-# `language.Rd` does not record. `(` is walked too but combines nothing, so
-# each reader adds it where it meets one.
+# `language.Rd` does not record. `(` is walked too but combines nothing, so it
+# is not here.
 selection_walk_operators <- function() {
   c("c", ":", "!", "-", "|", "&", "/")
 }
@@ -311,7 +315,9 @@ selection_refused_operators <- function() {
 # its selection where it has one, its failure where it has none. `TRUE` is what
 # lets a failure the names decide be raised before typed metadata is acquired
 # (ADR 0005); a shape this cannot answer for is `FALSE`, which resolves it
-# against the typed snapshot and is correct however that shape reads.
+# against the typed snapshot and is correct however that shape reads. Callers
+# hold `env` as the expression's own environment and `data_vars` as the input's
+# column names, which is the pair the symbol branch decides from.
 is_name_only_expr <- function(expr, env, data_vars) {
   if (is.symbol(expr)) {
     name <- as.character(expr)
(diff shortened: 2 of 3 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes, against `origin/main` (`d3a3455`),
reading `d159b9b`. Answered by `d4470b5`.

### Spec

| finding | disposition |
|---|---|
| `one_of()` in the leaf helpers made `.grouping` report its unknown-name warning twice — the name-only pass compiles a plan it discards, then the canonical pass resolves the same selection again | fixed in code, `d4470b5` |
| `^`, `&&`, `||`, and `one_of()` on `.by` were asserted only at the classifier level, not by the read count the acceptance criterion names | fixed in code, `d4470b5` |
| the four gate questions — ordering for both arguments, "answer without descending", a name-only classification the names do not settle, and whether the tests assert the rule rather than a backend | none found; the third was probed over 23 shapes against `grouping_name_proxy()` versus the real frame, and every expression classified `TRUE` gave an identical selection or an identical error |
| scope creep | none |

The `one_of()` finding also falsified a sentence in the issue's evidence
comment. It is deleted there rather than restated.

### Standards

| finding | disposition |
|---|---|
| the new ordering test asserted three External refusals with a bare `expect_error()`, where `design/architecture.md` requires a test to assert that an External condition keeps its class beside the behaviour that raises it (ADR 0015) | fixed in code, `d4470b5` |
| `is_name_only_expr()`'s new header carried what it answers but not what its callers hold, which *Code comments* in `AGENTS.md` requires of a header | fixed in prose, `d4470b5` |
| `selection_walk_operators()`'s header claimed each reader adds `(` where it meets one; only one does, and `is_grouping_spec_predicate()` already says why it must not | claim deleted, `d4470b5` |
| every factual claim the new comments make — `/` → `eval_slash()`, `&&`/`||` → `stop_bad_bool_op()`, `*`/`^` → `stop_bad_arith_op()`, operands unwalked, `language.Rd` recording no `/` | verified against the installed tidyselect; none false |
| baseline smells | none material; the nullary-constant extraction matches the repo's idiom |

Issue #217, opened 2026-08-17
Deduplication of a repeated warning does not hold under cli's rendering options


sayuks:
Split out of the review of #199. The Repeated-condition contract ADR 0021
states — one report, however many grouping sets raised it — does not hold in a
session with colour, which is most interactive sessions.

## What is wrong

`branch_warning_identity()` computes its deduplication key by matching patterns
against the message dplyr rendered: `^There (were|was) [0-9]+ warnings? in `,
`^[iℹ] In group `, `^Caused by `, and the `last_dplyr_warnings()` pointer.
When cli styles the output it wraps those markers in ANSI escapes, so every
pattern misses, nothing is removed from the key, and each branch's key differs
by its own grouping values.

Measured on `main` with the reproduction #141 was filed with:

```r
data <- data.frame(
  region = c("East", "East", "West"),
  grade  = c("a", "b", "a"),
  units  = c(1, 3, 6)
)

count <- function(colours) {
  op <- options(cli.num_colors = if (colours) 256L else 1L)
  on.exit(options(op), add = TRUE)
  n <- 0L
  withCallingHandlers(
    summarize_with_margins(
      data,
      total = sum(as.numeric(grade)),
      .grouping = cube(region, grade)
    ),
    warning = function(cnd) {
      n <<- n + 1L
      invokeRestart("muffleWarning")
    }
  )
  n
}
```

| session | reports |
| --- | --- |
| `cli.num_colors = 1` | 1 |
| `cli.num_colors = 256` | 4 |

The count is the number of grouping sets again, which is what #141 fixed for an
unstyled session only. Nothing about the warning differs between the four.

## Why it was not fixed with #199

#199 restores the caller's spelling in the argument bullet, and reads the same
rendered text through the same helper, so it inherits the same bound: in a
colour session the restatement no-ops too. Fixing the reading is one change in
`written_message_lines()` — strip the styling before either reading — and it
would fix both at once, but it changes the identity path, which is the part of
#141 with the most behaviour resting on it. ADR 0022 records the measurement
and says the fix belongs to ADR 0021's identity rather than to that decision.

## What a fix has to establish

- Which reading strips the styling. `cli::ansi_strip()` is the obvious one, and
  cli is already in the dependency closure through rlang — but this needs
  checking against `AGENTS.md`'s *Dependency metadata* rules rather than
  assumed, and a guard through `marginplyr_suggest_available()` if cli turns out
  not to be an Import of an Import.
[...]

PR #347, opened 2026-08-31
Keep a repeated warning one report past a caller's pointer line

### LINKED ISSUE #341: Keep a Repeated condition one report when the caller's own message carries dplyr's pointer line
## What to build

A warning the caller's summary expression raises in every grouping set is reported once,
with a count of the further sets that raised it, whatever the warning's own text says.

Today the identity a Repeated condition is matched on drops every line from the first one
matching dplyr's `last_dplyr_warnings()` pointer onward, wherever in the message that
line sits. dplyr appends such a pointer only at the end, and only to a message whose
header said there was more than one warning — but the search is not restricted to that
position, so a line the caller's own warning text contains is matched too.

The consequence is a split. In a branch where dplyr did append a real trailing pointer,
the last match is the real one and the caller's tail survives into the identity. In a
branch with one group, where dplyr appended none, the caller's own line is the only match
and everything after it is stripped. The two branches then compute different identities
for one warning, and it is reported twice instead of once.

The failure direction is under-collapsing: it reports more than it should and never
merges two genuinely distinct warnings. ADR 0021 accepts that direction for a dplyr
wording change, but this instance is triggered by caller-controlled text rather than by a
dplyr version, so it sits outside what that paragraph covers.

## Acceptance criteria

- [ ] An identical warning raised in every grouping set is reported once with the count of
      further grouping sets, including when its own text contains a line matching dplyr's
      pointer.
- [ ] Two genuinely different warnings are still never collapsed into one report.
- [ ] The pointer removal reaches only a pointer dplyr appended — the one at the end of a
      message whose header said there was more than one warning. That header is already
      read in the same place.
- [ ] The behaviour is pinned by a test whose warning text carries the pointer line, since
      nothing in the suite reaches this today.

## Blocked by

None — can start immediately.

sayuks:
Closes #341.

`branch_warning_identity()` looked for dplyr's pointer at `last_dplyr_warnings()`
anywhere in the rendered message. dplyr appends one only at the end of a message
whose header counted more than one warning, so a line the caller's own warning
text spelled was read as dplyr's — and the two branches of a `rollup()` then
computed different identities for one warning, reporting it twice.

The removal now reads the count out of the header it already matches, and takes
the last written line only where that count is more than one.

`tests/testthat/test-execution-conditions.R` pins it with a warning whose text
carries the pointer line; nothing in the suite reached this before. The pair of
distinct diagnostics that both spell the pointer line is still two reports.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -178,14 +178,13 @@
   wrote, and an error blames the Margin verb you called rather than an internal
   summary. A warning every grouping set raises is reported once and says how
   many further grouping sets raised it, in place of one identical warning per
-  set — `2^k` of them for a `cube()` of `k` dimensions — while warnings that
-  differ from each other are still reported one by one. That holds whatever
-  your own warning text says, including a diagnostic of yours that spells the
-  line dplyr writes to point at `dplyr::last_dplyr_warnings()`. Only that
-  context changes: the class, the diagnostic, and the cause you receive are
-  the ones raised. A lazy input is unaffected, because its summary
-  expressions run when you collect the result rather than while the verb runs
-  (#141, #108).
+  set — `2^k` of them for a `cube()` of `k` dimensions — including a warning
+  whose own text spells the line dplyr writes to point at
+  `dplyr::last_dplyr_warnings()`, while warnings that differ from each other
+  are still reported one by one. Only that context changes: the class, the
+  diagnostic, and the cause you receive are the ones raised. A lazy input is
+  unaffected, because its summary expressions run when you collect the result
+  rather than while the verb runs (#141, #108).
 * `nest_with_margins()` and `nest_by_with_margins()` now use collision-free
   internal columns. `.keep = TRUE` retains original pre-margin key values,
   and nesting rejects duplicate sets with `.duplicates = "keep"` because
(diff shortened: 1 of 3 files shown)


sayuks:
## Review record

One round, `mattpocock-skills:code-review` over `git diff main...HEAD` at 12d18de (merge-base `e80fa5f`), both axes. Answered in 99afe68.

### Standards — 3 findings, 3 dispositions

| finding | disposition |
|---|---|
| `NEWS.md`: "That holds whatever your own warning text says" is false — an empty caller warning is still two reports, the *Left whole* degradation ADR 0022 records | fixed in prose: the clause is deleted, the pointer-line case is what the sentence says (99afe68) |
| One rule in three copies inside `branch_warning_identity()` — the header block that owns it, plus two inline restatements | fixed in prose: the restatements are gone; each site keeps only what it alone knows (99afe68) |
| `counted` held a match vector, not a count, and needed a comment to say what `[[3L]]` was — the naming is the finding | fixed in code: `header_match`, read once into `warning_count`; `last` names the position `written` and `removed` were indexed at under two spellings (99afe68) |

### Spec — 1 finding, 1 disposition

| finding | disposition |
|---|---|
| The new test sampled one rendering where every other collapse fixture crosses them (ADR 0021: "a green suite under one rendering says nothing about another"), and the pointer's spelling is what `cli.hyperlink_run` changes (#217) | fixed in code: the fixture runs over the twenty configurations, asserting the markers before the collapse (99afe68) |

No requirement of #341 was found missing, and no scope creep. The round reproduced the split against `main`'s `branch_warning_identity()` and confirmed against dplyr that the pointer appears only for a count above one and only as the final line.

PR #352, opened 2026-08-31
Scope the factor-preservation claim to the backends that restore factors

### LINKED ISSUE #343: Scope the factor-preservation claim to the backends that restore factors
## What to build

The reference tells a reader which backends preserve a factor dimension's levels and
ordering, rather than claiming every backend does.

The *Result class and attributes* section currently states that factor and ordered-factor
columns are the one exception to marginplyr not preserving input attributes, and that
their levels and ordering are preserved. It is written without qualification, and
`expand_with_margins()` and the nesting verbs inherit the section as written.

On Arrow it is false under the default Margin label: a factor dimension comes back as a
character column, ordered by its values rather than by its levels. Plain dplyr over the
same Arrow table keeps the dictionary type through `summarise()` and `collect()`, so the
levels are lost by marginplyr rather than by Arrow.

The code is deliberate and already pinned — Arrow can read a schema but cannot restore
factors, because the schema a proxy is built from carries a factor column with no levels
in it — so this is the documentation's claim being wrong, not the behaviour. Two existing
tests assert the character result and one states in a comment that level order is
deliberately not asserted there.

The `.margin_label` section makes the same unqualified claim about reconstruction a second
time.

## Acceptance criteria

- [ ] The *Result class and attributes* section names the condition under which levels and
      ordering are preserved instead of claiming it unconditionally, and the claim agrees
      with what the backend capability table actually decides.
- [ ] The second unqualified statement, in the Margin label contract, agrees with the
      first — one of them owns the condition and the other cites it rather than restating
      it.
- [ ] A reader on Arrow can tell from the reference alone what type a factor dimension
      comes back as.
- [ ] Generated documentation is regenerated in the same commit, per *Documentation* in
      `AGENTS.md`.

## Blocked by

None — can start immediately.

sayuks:
Closes #343.

*Result class and attributes* claimed without qualification that a factor
dimension's levels and ordering survive a Margin operation, and
`expand_with_margins()` and the nesting verbs inherit the section as written.
On Arrow the claim is false under the default Margin label: `can_restore_factors`
is false there, so the dimension comes back as character ordered by its values,
while plain `dplyr::summarize()` over the same table keeps the dictionary type.

The section now names the three backends `backend_capabilities()` grants
restoration to — local data frames, `dtplyr`, DuckDB — says what Arrow returns
instead and why the schema a prototype is read from cannot supply levels, notes
that `NA_character_`/`NULL` insert no label and convert nothing, and says a SQL
backend other than DuckDB never held a factor to lose.

The Margin label contract and the Margin order key now cite that condition
rather than restating it, so one section owns it.

`man/` regenerated in the same commit. Lint, spelling, the context budget, and
the full suite under `NOT_CRAN=true` are green locally.

### DIFF
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -239,11 +239,27 @@
 #' attributes a result must carry after the Margin operation, as with any
 #' dplyr pipeline.
 #'
-#' Factor and ordered-factor columns are the one exception, because
-#' marginplyr decomposes them to insert `.margin_label` and rebuilds them
-#' itself. Their levels and ordering are preserved; see *Display labels and
-#' grouping identity*. Classed columns such as [Date] and [POSIXct], including
-#' its `tzone`, are carried through by dplyr and vctrs unchanged.
+#' Factor and ordered-factor columns are the one exception, on the backends
+#' whose factors marginplyr can rebuild: local data frames, `dtplyr`, and
+#' DuckDB. There it decomposes them to insert `.margin_label` and rebuilds
+#' them itself, so their levels and ordering are preserved; see *Display
+#' labels and grouping identity*. On DuckDB it is the levels alone: an `ENUM`
+#' carries no ordered flag, so an ordered factor is already unordered when
+#' marginplyr reads the column and unordered is what it rebuilds.
+#'
+#' Arrow is the one backend that can hold a factor without marginplyr being
+#' able to rebuild one, because the schema a dimension's prototype is read
+#' from carries a dictionary column with no levels in it. A non-missing
+#' `.margin_label` there converts the dimension to character as it does any
+#' ordinary dimension, so the levels and the ordered status are gone and a
+#' Margin order sorts the dimension by its values; plain
+#' [dplyr::summarize()] over the same table keeps the dictionary type, so
+#' this is marginplyr's loss rather than Arrow's. `NA_character_` and `NULL`
+#' insert no label and so convert nothing. A SQL backend other than DuckDB
+#' reaches none of this, storing a factor as text before marginplyr sees it.
+#'
+#' Classed columns such as [Date] and [POSIXct], including its `tzone`, are
+#' carried through by dplyr and vctrs unchanged.
 #'
 #' @section Grouping set identifiers:
 #' When `.id` names an output column, each result row receives the one-based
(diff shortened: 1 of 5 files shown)


sayuks:
## Review record

One round, `mattpocock-skills:code-review`, two axes over
`git diff origin/main...HEAD` at the merge-base `60fa93c`, reading
`cca1581` (now `9ce26ef`). Standards was passed *Documentation*, *Linting*,
and *Code comments* from `AGENTS.md`; both axes were passed *Answering a
review* from `design/architecture.md` for what a finding about prose may be.
`man/*.Rd` is generated and outside the review.

Three findings.

**DuckDB cannot carry an ordered factor — fixed in prose.** Raised by both
axes. `dbWriteTable()` stores an ordered factor as a bare `ENUM`, which has no
ordered flag, so `collect()` returns an unordered factor before marginplyr
reads the column. Naming DuckDB alongside local and `dtplyr` for "levels and
ordering are preserved" promised a case DuckDB cannot reach. The owning
section now says the levels alone are preserved there and why.

**`expand_with_margins()` refuses an ordered factor dimension on `dtplyr` —
filed as #354.** Confirmed: `collect()` raises
`<dt_join_type_mismatch_error>` from data.table's `.set_ops_arg_check()`,
where the unordered column, `summarize_with_margins()`, and the local data
frame all succeed. `can_restore_factors` is true for `dtplyr`, so the
sentence is right about what the capability table decides and the defect is
in the expand path. Per *Answering a review*, a finding this branch's
Acceptance does not reach becomes a ticket.

**The three-backend list duplicates `design/adr/0018`'s — rejected.** ADR 0018
names the same set as a supporting clause of its ordering argument, so a
fourth backend gaining the capability touches both. The reference already
carries user-facing backend lists derived from `backend_capabilities()` in two
other sections — *Database backend coverage* for `native_grouping_sets`, and
*When marginplyr queries your data* for `collect_selection_proxy` — so a list
here is the established shape and not a new drift site. A user-facing
reference cannot cite `design/`, and the capability names are not public.

Everything else the axes checked verified: the three backends against
`backend_capabilities()`, the empty-levels dictionary the Arrow schema proxy
yields, Arrow returning character under a non-missing label and an ordered
factor under `NA_character_`/`NULL`, plain `dplyr::summarize()` keeping the
dictionary type on the same table, SQLite storing the column as text, and
`roxygen2::roxygenise()` leaving the tree clean. The *Margin order* edit,
which #343 does not name, was judged warranted by criterion 1 rather than
scope creep: a third unqualified sentence in the same reference would have
left the claim disagreeing with the capability table.

PR #353, opened 2026-08-31
Settle a formula and the `.data$name` pronoun from the names (#346)

### LINKED ISSUE #346: Settle the two selection shapes the name-only reader still cannot reach
## What to build

`is_name_only_expr()` answers whether tidyselect settles an expression from the
column names alone. #342 established that question and closed the three shapes
it named. Two shapes still answer `FALSE` that the question settles as `TRUE`,
and neither is reachable by the edit #342 made — each is held out by a decision
that lives somewhere else.

### 1. `~`, which the reader cannot see

tidyselect refuses a formula in a selection outright, in `stop_formula()`, so
its failure depends on nothing and a fortiori not on column types. Measured
against `grouping_selection_proxy()`:

```
grouping_set(~c(region))   reads=1   Formula shorthand must be wrapped in `where()`.
.by = ~c(region)           reads=1   Formula shorthand must be wrapped in `where()`.
```

That is the ordering ADR 0005 forbids, and it is the same defect #342 fixed for
`*`, `^`, `&&`, and `||`. Adding `"~"` to `selection_refused_operators()` does
not reach it: `is_nameable_call()` declines a call to `~`, so
`static_call_name()` answers `NULL` and `is_name_only_expr()` returns `FALSE`
before any operator set is consulted. That refusal is deliberate and has its
own reasons (ADR 0019, #163, and the formula tests in
`tests/testthat/test-static-expression-analysis.R`), so this needs a reading
that does not route through the name, not a wider list.

### 2. `.data$x`, which the reader declines conservatively

`.data$region` is a valid selection settled by the names — tidyselect's walk
has a `.data` branch of its own — and it currently costs a snapshot:

```
grouping_set(.data$region)   reads=1   (succeeds)
```

No error is deferred, so this is not an ADR 0005 violation; the fast pass is
skipped where the names would have settled the plan. It is narrower than it
looks: `.data$x` with a literal name is settled, while `.data[[var]]` is not,
so whatever reads this has to tell a symbol right-hand side from an expression
one. `.data$x` is also deprecated in a tidyselect selection, which is a reason
to weigh before spending anything on it.

## Acceptance criteria

- [ ] A `.grouping` or `.by` selection written as a formula raises before any
      query reaches the backend, for both arguments, without widening what
      `is_nameable_call()` names.
- [ ] `.data$x` is either settled by names or deliberately left `FALSE` with
      the reason recorded at the site. Answering "left alone" is a valid close;
      answering it by silence is not.
- [ ] Whatever reads `.data$` distinguishes a literal name from `.data[[var]]`,
      or does not claim to read `.data` at all.
- [ ] The read count is what a test asserts, as
      `tests/testthat/test-grouping-plan.R`'s ordering block already does.

## Context

#342 and PR #345 establish the invariant, the two readers' relationship, and
the read-counting test these would extend. `d159b9b`'s
`selection_refused_operators()` is the set item 1 would join if it could.

sayuks:
Closes #346.

`is_name_only_expr()` answers whether tidyselect settles an expression from the
column names alone, so a failure the names decide is raised before typed
metadata is acquired (ADR 0005). #342 closed three shapes; this closes the two
it could not reach.

## A formula

tidyselect refuses one in a selection on sight, in `stop_formula()`, whatever
it wraps, so the refusal depends on no column type. `"~"` cannot join
`selection_refused_operators()`: `is_nameable_call()` declines a call to `~`
(ADR 0019, #163), so the name that set is consulted by is never read. The
shape is read directly, before the name is taken. A quosure is a `~` call too
and is excluded — tidyselect evaluates one rather than refusing it.

## `.data$name`

Settled by the names: tidyselect's walk has a `.data` branch that looks up the
name written in the call. The reader takes the `$` call whose left operand is
the `.data` symbol *as written* — tidyselect refuses `(.data)$region`, so the
pair is not read through there — and whose right operand is a symbol.
`.data[[...]]` is not read, and `FALSE` for it resolves against the snapshot.

## The warning that had to be kept

`.data$x` is deprecated in a selection and lifecycle signals that once per
session. The discarded name-only pass muffles what it signals, so settling
`.data$region` made the pass spend the caller's only signal and the canonical
pass report nothing — less than the second pass reports, where the pass's rule
had only ever been that it reports no *more*. `lifecycle_verbosity = "quiet"`
over the discarded pass fixes the signalling rather than the display; every
other warning is still suppressed.

## Read counts

| call | before | after |
|---|---|---|
| `grouping_set(~c(region))` | 1 | 0 |
| `rollup(~c(region))` | 1 | 0 |
| `.by = ~c(region)` | 1 | 0 |
| `grouping_set(.data$absent)` | 1 | 0 |
| `rollup(.data$absent)` | 1 | 0 |
| `.by = .data$absent` | 1 | 0 |
| `rollup(everything() / where(is.numeric))` | 1 | 1 |
| `grouping_set(.data[["absent"]])` | 1 | 1 |

Local: full `testthat` suite green with `NOT_CRAN=true`, `lintr::lint_package()`
clean.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -163,8 +163,10 @@
   selection carrying `where()` still resolves against your input's column
   types, including one written under `/`. It reaches two more spellings: a
   formula, which tidyselect refuses on sight whatever it wraps, and the
-  `.data$name` pronoun, whose name your input either holds or does not.
-  `.data[[name]]` is unchanged and still resolves against your input.
+  `.data` pronoun, whose name your input either holds or does not. Where you
+  have turned a deprecation into an error with
+  `options(lifecycle_verbosity = "error")`, the deprecation `.data` carries in
+  a selection is now raised before the query too.
 * Every error marginplyr raises for a correctable call now inherits the
   `"marginplyr_error"` class, so `tryCatch(marginplyr_error = )` catches them
   all. It is the only promised class; narrower subclasses and message wording
--- a/NEWS.md
+++ b/NEWS.md
@@ -161,7 +161,10 @@
   on a backend that needs one, so on a disconnected or slow connection you
   were handed the connection's failure instead of the column diagnostic. A
   selection carrying `where()` still resolves against your input's column
-  types, including one written under `/`.
+  types, including one written under `/`. It reaches two more spellings: a
+  formula, which tidyselect refuses on sight whatever it wraps, and the
+  `.data$name` pronoun, whose name your input either holds or does not.
+  `.data[[name]]` is unchanged and still resolves against your input.
 * Every error marginplyr raises for a correctable call now inherits the
   `"marginplyr_error"` class, so `tryCatch(marginplyr_error = )` catches them
   all. It is the only promised class; narrower subclasses and message wording
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -190,12 +190,15 @@ prepare_grouping_plan <- function(.data,
         # them is what keeps this pass invisible but for the failure it exists
         # to raise; `one_of()` is where that is observable.
         #
-        # A deprecation warning is the one kind suppressing is not enough for,
-        # and the option is what covers it. lifecycle signals one once per
-        # session, so a pass that signals it and muffles it spends the only
-        # signal the caller had coming and the canonical pass then reports
-        # nothing -- less than the second pass reports, where every other
-        # warning here is more. `.data$region` is where that is observable.
+        # A deprecation warning is the one kind suppressing is not enough for.
+        # lifecycle signals one once per session, so a pass that signals it
+        # and muffles it spends the only signal the caller had coming and the
+        # canonical pass then reports nothing. The option withholds the signal
+        # instead of hiding it; `.data$region` is where that is observable.
+        #
+        # `"error"` is the one verbosity it does not replace, because under it
+        # the deprecation is a failure the names decide and quieting it would
+        # put the read below in front of it (ADR 0005).
         rlang::with_options(
           suppressWarnings(compile_grouping_spec(
             grouping_spec,
@@ -206,7 +209,7 @@ prepare_grouping_plan <- function(.data,
             duplicates_choices = duplicates_choices,
             preflight = preflight
           )),
-          lifecycle_verbosity = "quiet"
+          lifecycle_verbosity = discarded_pass_verbosity()
         )
       }
       data_proxy <- grouping_selection_proxy(data, backend = backend)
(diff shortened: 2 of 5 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` against `main...HEAD`, two axes in parallel, over `10924d1`. Three findings; all answered in `9b907f4`.

|---|---|---|---|
| 1 | Standards | The discarded pass set `lifecycle_verbosity = "quiet"` unconditionally, so under `options(lifecycle_verbosity = "error")` the `.data` deprecation — a failure the names decide — was raised only after `grouping_selection_proxy()` had run. ADR 0005. | Fixed in code. `discarded_pass_verbosity()` keeps `"error"`. Gated by *a deprecation that is an error precedes the read*. |
| 2 | Standards + Spec (both axes, independently) | The new warning-count block pinned `lifecycle_verbosity = "warning"` — lifecycle's always-warn mode, where the once-per-session spend the option prevents cannot occur. Deleting the option left the whole file passing. | Fixed in code and prose. The count keeps only what it measures; *the discarded pass withholds a deprecation, not muffles it* observes `getOption("lifecycle_verbosity")` per pass through a mock over `compile_grouping_spec()`. Both mutations (no option; unconditional `"quiet"`) now fail. |
| 3 | Spec | The recorded reason for leaving `.data[[...]]` `FALSE` was false — tidyselect settles it from the names too. | Fixed in code. `is_data_pronoun()` takes the whole pronoun; the premise is measured in `investigation/what-the-data-pronoun-settles-in-a-selection.md`. |

Two judgement calls, both answered in prose: the ADR 0019 citation beside the operand comparison is dropped (that ADR's boundary is a call's *head*; the site reason carries the decision), and a restating sentence is dropped from the pass's comment. The `~` branch not joining `selection_refused_operators()` was raised as mild duplication and left as considered — the header says why, and `is_nameable_call()` is what makes the merge impossible.

One reported finding did not survive: a `~` call built with `rlang::call2()` was said to escape `stop_formula()`. Measured at zero, one, two, and three arguments, tidyselect refuses all of them.

Read counts after the answer:

| call | before | after |
|---|---|---|
| `grouping_set(~c(region))`, `rollup(~c(region))`, `.by = ~c(region)` | 1 | 0 |
| `grouping_set(.data$absent)`, `rollup(.data$absent)`, `.by = .data$absent` | 1 | 0 |
| `grouping_set(.data[["absent"]])`, `rollup(.data[["absent"]])` | 1 | 0 |
| `.data$region` under `lifecycle_verbosity = "error"` | 1 | 0 |
| `rollup(everything() / where(is.numeric))` | 1 | 1 |

Full suite green with `NOT_CRAN=true`; `lintr::lint_package()` clean.

PR #355, opened 2026-08-31
Terminate the readers that restart on a pair holding the empty argument

### LINKED ISSUE #349: Terminate the share walk on a pair of parentheses holding the empty argument
## What to build

A summary expression holding `()` — a pair of parentheses wrapping R's empty
argument — is answered by dplyr or by marginplyr, rather than by exhausting the
evaluation stack.

`contains_share_helper()` in `R/share.R` does not terminate on such an
expression. Every path into it fails the same way:

```r
summarize_with_margins(d, z = sum(!!rlang::call2("(", rlang::missing_arg())), .grouping = rollup(region))
#> Error: evaluation nested too deeply: infinite recursion / options(expressions=)?
```

`dplyr::summarise()` answers the same input with its own condition, class
intact — `In argument: z = sum(())` / `Caused by error in ()`: `argument 1 is
empty`. An `expressionStackOverflowError` is catchable by no class a caller
would write and names nothing they wrote.

The same pair reaches the walk through an `across()` argument, which
`parse_across_arguments()` reads for R's marker alone and does not see through
a pair:

```r
summarize_with_margins(d, !!rlang::call2("across", quote(value), rlang::call2("(", rlang::missing_arg()), .ns = "dplyr"), .grouping = rollup(region))
```

Only an injection or a constructed call spells it, since the parser rejects
`f(())`.

This is a walk that fails to terminate, not a position that misreads an
argument, so #340's `is_empty_argument()` does not reach it: that predicate is
asked of a verb's or a constructor's own argument, and this pair sits inside a
caller's expression. Found while answering #340's third review round;
`R/share.R` is untouched by that branch and the defect predates it.

## Acceptance criteria

- [ ] A summary expression holding a pair of parentheses wrapping the empty
      argument reaches dplyr and receives dplyr's own answer with its class
      intact, or raises a condition inheriting `marginplyr_error`.
- [ ] The same holds wherever the pair sits: a bare call argument, an `across()`
      selection, an `across()` `.fns`.
- [ ] No walk in `R/share.R` recurses on a node that is not smaller than the one
      it was called with.
- [ ] The dplyr oracle the `across()` tests already use decides what an empty
      `across()` argument answers; a pair wrapping one is not read as omitted
      unless dplyr reads it that way.

## Blocked by

None — can start immediately.

sayuks:
Closes #349.

## What was wrong

`unparenthesized_value()` stops at a pair of parentheses wrapping R's empty argument rather than unwrapping to it — a documented contract, since a pair holding nothing wraps no value to read and unwrapping would put R's missing marker where a local could bind it. It therefore answers the node it was given.

Two readers restarted on that answer under an `is_redundant_parens()` guard, which is true again on re-entry:

- `static_spelling_reference_name()`, reached from the share walk;
- `static_character_value()`, reached from the walk over a statically known string — **with no share helper and no `across()` anywhere in the expression**, so `z = get(())` reached it too. The issue named one reader; the branch found the second.

Both recursed without bound. The caller was answered with an `expressionStackOverflowError`, which is catchable by no class they would write and names nothing they wrote, where dplyr answers `argument 1 is empty`.

Only an injection or a constructed call spells the pair, since the parser rejects `f(())`.

A third position had a different cause and the same symptom. `across(value, sum, .names = ())` terminated once the guards were fixed and then raised a bare `simpleError` from marginplyr's own frame: the read of `.names` in the selection rewrite was the one of three `eval_tidy()` calls with no fallback, so the template was evaluated here instead of by dplyr. Found by the review of the first commit, not by the issue.

## The fix

Both readers guard with `is_parenthesized()` — "whether unwrapping answers a different node, which is what a reader that restarts on what the parentheses wrap has to know before it restarts". It is `FALSE` for exactly the pair the peeler stops at, so the restart cannot repeat. This is the guard the three readers that already restart correctly use, and `grouping_arg_spec()`'s comment cites the rule *while naming `static_spelling_reference_name()` as following it* — which it did not.

An unevaluable `.names` template keeps the argument the caller wrote, as the reads of `.names` and `.unpack` beside it keep the value they use for an absent argument, so the caller sees the condition their own argument produces (ADR 0015).

## What holds the rule

Two halves, because they establish different things.

A **structural gate** in `tests/testthat/test-utils.R` refuses the shape: a body that tests for a pair — named as `is_redundant_parens()` or written out — and hands the peeler's answer to another call, which is the re-entry. It names both readers before the fix and neither after. Scanned rather than listed, for the reason the two gates beside it are: two readers were already written this way, so a list would be a list the third is not on.

A **companion test** calls every reader that reads through a pair with the node a restart cannot shrink. That is where one that recursed would exhaust the stack rather than answer, and it is what establishes termination for the readers already here rather than only for a reader written tomorrow.

What neither can see is a restart on a peeler other than `unparenthesized_value()`: `unparenthesized_name()` answers its input for a pair wrapping anything but a name, so a restart on it under `is_parenthesized()` would not terminate. No reader is written that way, and the companion test is what says so for each reader that exists.

## Where dplyr is the oracle, and where it is not

Six of the seven positions answer with dplyr's own condition, class and message identical:

| expression | before | after |
|---|---|---|
| `sum(())` | `expressionStackOverflowError` | dplyr's `rlang_error`, identical |
| `dplyr::across(value, ())` — `.fns` | `expressionStackOverflowError` | dplyr's `rlang_error`, identical |
| `dplyr::across(value, sum, .names = ())` | `expressionStackOverflowError` | dplyr's `rlang_error`, identical |
| `dplyr::across(value, sum, .unpack = ())` | `expressionStackOverflowError` | dplyr's `rlang_error`, identical |
| `get(())` | `expressionStackOverflowError` | dplyr's `rlang_error`, identical |
| `sum(get(()))` | `expressionStackOverflowError` | dplyr's `rlang_error`, identical |
| `dplyr::across((), sum)` — `.cols` | `expressionStackOverflowError` | the cause dplyr reports, unwrapped |

**The `across()` *selection* is the one position dplyr is not the whole oracle for, and it is a deviation from the issue's first acceptance criterion that this branch is not the right place to close.** A Margin verb resolves a selection itself before the call is sent, so a selection that cannot be resolved is answered by the condition the selection produces and not by the one dplyr wraps around it. That is not specific to the empty pair — no selection error matches dplyr's class here:

| selection | dplyr | marginplyr, on this branch and before it |
|---|---|---|
| `across(nope, sum)` | `rlang_error` | `vctrs_error_subscript_oob` |
| `across(1:99, sum)` | `rlang_error` | `vctrs_error_subscript_oob` |
| `across((), sum)` | `rlang_error` | `simpleError` from tidyselect's `env_has()` |

The three differ from each other as well as from dplyr, because each is whatever tidyselect raises for that selection. Meeting the criterion literally would mean changing how every selection error in the package is reported, which is a contract change and not a termination fix.

The test holds the pair to what its neighbours are held to: it terminates, and it carries the cause dplyr reports underneath its own wrapper. It also asserts the pair is **not** read as an omitted selection — read as omitted it would have selected every column and returned a result where dplyr raises, which is the failure the issue's fourth criterion is about.

## Checks

`lintr::lint_package()` clean; full suite green with no failures and no skips; `verify-context-budget.R` and `verify-verifier-invocation.R` pass. No roxygen touched, so `man/` is unchanged.

### DIFF
--- a/R/contextual-helpers.R
+++ b/R/contextual-helpers.R
@@ -214,8 +214,14 @@ is_any_static_spelling_call <- function(expr, families) {
 # may have left empty -- an `across()` written `across(units, )` reaches here
 # with R's missing marker, and assigning that to a local raises
 # `missingArgError` on the next read of it (#174).
+#
+# What stops the recursion is the guard rather than the branches below it, so
+# the guard is `is_parenthesized()`: the same argument may be a pair wrapping
+# the empty one, `unparenthesized_value()` stops at such a pair rather than
+# unwrapping to it, and under `is_redundant_parens()` this restarted on the
+# node it was called with (#349).
 static_spelling_reference_name <- function(expr, family) {
-  if (is_redundant_parens(expr)) {
+  if (is_parenthesized(expr)) {
     return(static_spelling_reference_name(
       unparenthesized_value(expr),
       family
--- a/R/summary-selections.R
+++ b/R/summary-selections.R
@@ -556,10 +556,18 @@ rewrite_across_selection <- function(expr,
   if (identical(call_name, "across")) {
     parsed <- parse_across_arguments(rebuild_static_call(expr, call_args))
     if (!is.null(parsed$names)) {
-      call_args[[parsed$names_index]] <- rlang::eval_tidy(
-        parsed$names,
-        env = env
+      # A template that does not evaluate is left as the caller wrote it, so
+      # dplyr raises the condition their own argument produces rather than this
+      # frame raising it first (ADR 0015). The two reads of `.names` and
+      # `.unpack` above answer a failure the same way, with the value they use
+      # when the argument is absent; here the value is the argument itself.
+      template <- tryCatch(
+        list(rlang::eval_tidy(parsed$names, env = env)),
+        error = function(cnd) NULL
       )
+      if (!is.null(template)) {
+        call_args[[parsed$names_index]] <- template[[1L]]
+      }
     }
   }
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -1113,7 +1113,13 @@ static_character_value <- function(expr) {
   # one are read through as they are everywhere else: `get(("share"))` names
   # what `get("share")` names. Read by recursion rather than by rebinding
   # `expr`, which is an argument a `c()` above may have left empty (#174).
-  if (is_redundant_parens(expr)) {
+  #
+  # `is_parenthesized()` and not `is_redundant_parens()`, because that argument
+  # may also be a pair wrapping the empty one. `unparenthesized_value()` stops
+  # at such a pair rather than unwrapping to it, so the weaker test restarted
+  # this reader on the node it was called with, and `get(())` exhausted the
+  # evaluation stack before dplyr saw the call (#349).
+  if (is_parenthesized(expr)) {
     return(static_character_value(unparenthesized_value(expr)))
   }
   if (is.character(expr)) {
--- a/tests/testthat/helper-namespace-walk.R
+++ b/tests/testthat/helper-namespace-walk.R
@@ -19,7 +19,8 @@
 # - `test-query-policy.R`, ADR 0020's rule about which functions reach an
 #   execution entry point, and the capability table read out of the source;
 # - `test-utils.R`, that no walk over a call's arguments binds one to a name,
-#   in either the `for` spelling or the `<-` spelling;
+#   in either the `for` spelling or the `<-` spelling, and that no reader
+#   restarts on a pair of parentheses it cannot shrink;
 # - `helper-diagnostic-sites.R`, for the two diagnostics gates it serves.
 #
 # That list is a courtesy and not the record: `test-namespace-walk.R` derives
--- a/tests/testthat/test-static-expression-analysis.R
+++ b/tests/testthat/test-static-expression-analysis.R
@@ -2653,11 +2653,27 @@ test_that("a pair holding nothing in an expression evaluates as dplyr does", {
   # A bare call argument and an `across()` `.fns` reach the reference-name read
   # the share walk takes; `get()` reaches the reader of a statically known
   # string, with no share helper and no `across()` anywhere in the expression.
-  # Each of these is sent to dplyr as the caller wrote it, so dplyr's answer is
-  # the whole answer.
+  # `.names` and `.unpack` are read by evaluating them, which is a third way to
+  # reach the pair and the one that raises before any walk sees it. Each of
+  # these is sent to dplyr as the caller wrote it, so dplyr's answer is the
+  # whole answer.
   expressions <- list(
     bare = rlang::call2("sum", empty),
     fns = rlang::call2("across", quote(value), empty, .ns = "dplyr"),
+    names = rlang::call2(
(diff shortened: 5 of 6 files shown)


sayuks:
## Review record

Round run with `mattpocock-skills:code-review` against the merge-base with `main`, on `f33bbe4`. Both axes; five findings, all answered in `b970fa4`.

### Standards — 3 findings, all about prose

| finding | disposition |
|---|---|
| The two comments beside the changed guards re-derive what `is_parenthesized()`'s header owns, and say it in the past tense | fixed in prose — cut to the citation, which is what the three already-correct call sites carry |
| `walks_matching()`'s header enumerates two gates, two witnesses, and "both gates here are in one file"; there are three of each now | fixed in prose — corrected there and in the new gate's header, which set up the contradiction |
| The gate's predicate binds a call head (`head <- node[[1L]]`) while the file's rule is subscript throughout | fixed in code — the two predicates are one traversal, reading by subscript |

The first finding is the one worth naming: `grouping_arg_spec()`'s comment re-derives the same argument *and names `static_spelling_reference_name()` as following it*, which is exactly the drift **Code comments** describes — the comment stood while the function it named did the opposite. Adding two more copies would have been the same bet.

### Spec — 2 findings

**`across(value, sum, .names = ())` was not covered, and was not fixed.** After the guard change it terminated and then raised a bare `simpleError` from marginplyr's own frame: the read of `.names` in the selection rewrite was the one of three `eval_tidy()` calls with no fallback, so the template was evaluated here instead of by dplyr. Fixed in code — an unevaluable template keeps the argument the caller wrote, as the `.names` and `.unpack` reads beside it keep the value they use for an absent argument, and the caller sees dplyr's condition (ADR 0015). `.names` and `.unpack` now both answer identically to dplyr and both are in the oracle.

This means the deviation argued in the PR description is narrower than it was written: it is the `.cols` *selection* alone, not "the `across()` position". `.fns`, `.names`, and `.unpack` all match dplyr exactly.

**The gate established the rule for a reader written tomorrow and nothing for the readers already here** — their termination rested on prose. Fixed in code: a companion test calls every reader that reads through a pair with the node a restart cannot shrink, which is where one that recursed would exhaust the stack rather than answer. The gate also now matches the pair test written out rather than named.

### Left standing, deliberately

- **A restart on a peeler other than `unparenthesized_value()`.** `unparenthesized_name()` answers its input for a pair wrapping anything but a name, so a restart on it under `is_parenthesized()` would not terminate. The gate cannot see that shape. No reader is written that way, and the companion test is what says so for each reader that exists — recorded here and in `b970fa4`'s message rather than as a comment, since it is a property of the gate and not of any site.
- **`.by = ()` returns a result where `dplyr::summarise(.by = ())` errors.** Pre-existing on both trees and outside this branch: reading a pair wrapping the empty argument as an omitted verb argument is the decision #259/#261 made, and `test-static-expression-analysis.R`'s *a pair of parentheses holding nothing is the empty argument* is what owns it. AC4 is about `across()`, where dplyr is the oracle and the pair is not read as omitted.

### Checks

`lintr::lint_package()` clean; full suite green, no failures and no skips; `verify-context-budget.R` and `verify-verifier-invocation.R` pass. No roxygen touched, so `man/` is unchanged.

Issue #169, opened 2026-08-10
A helper taking bare column names rejects an injected quosure, naming a fault the caller does not have


sayuks:
## Parent

Part of #23. Found while implementing #165, by a review of that branch. Not
#165: this reproduces unchanged on `main` (`b178763`), and it is about a
quosure the caller passes *to* a helper rather than one the walk rebuilds.

---

**File / line:** `R/grouping-context.R:206` (`grouping_helper_vars()`), and
`R/share.R:1444` (`validate_share_direct_syntax()`).

## What is wrong

Four helpers take bare column names and test for one with
`rlang::is_symbol()` / `is.symbol()`. A quosure carrying a bare symbol is not a
symbol, so each rejects it — with a message describing a mistake the caller did
not make.

The spelling that reaches this is the standard tidy-eval wrapper, where
`rlang::enquo()` is what the idiom hands you:

```r
level_by <- function(data, col) {
  q <- rlang::enquo(col)
  summarize_with_margins(
    data,
    level = grouping_id(!!q),
    .grouping = rollup(region)
  )
}

level_by(d, region)
#> Error: `grouping_id()` only accepts bare grouping columns.
```

The caller did pass a bare grouping column. All four helpers behave the same
way:

| Helper | Diagnostic |
|---|---|
| `grouping_id()` | `` `grouping_id()` only accepts bare grouping columns. `` |
| `grouping_bit()` | `` `grouping_bit()` only accepts bare grouping columns. `` |
| `share_of_total()` | `requires exactly one bare name of a preceding ordinary summary` |
| `share_of_parent()` | as above |

**There is a workaround, and it is the reason this is not urgent:** injecting a
symbol rather than a quosure works today.

```r
summarize_with_margins(
  d,
  level = grouping_id(!!rlang::sym("region")),
  .grouping = rollup(region)
)
#> works
```

So a wrapper must reach for `rlang::ensym()` where `enquo()` is what the
surrounding tidy-eval idiom suggests, and nothing in the diagnostic says so.

## What the fix has to decide

This is why it is filed for triage rather than as agent-ready work: the
mechanical change (unwrap with `rlang::quo_get_expr()` before the symbol test)
is not the decision.

1. **Which injected forms are accepted.** A quosure wrapping a bare symbol is
   the clear case. A quosure wrapping `.data$region`, a string, or a call is
   not, and each already has an answer for the un-injected spelling that the
   injected one should presumably match.
[...]

PR #356, opened 2026-08-31
Answer an injected `across()` selection and `.names` (#350)

### LINKED ISSUE #350: Answer an empty across() .names without evaluating the empty symbol
## What to build

An `across()` `.names` argument a caller left empty is answered by dplyr or by
marginplyr, rather than by R evaluating the empty symbol.

`R/summary-selections.R` evaluates the caller's `.names` argument unguarded:

```r
call_args[[parsed$names_index]] <- rlang::eval_tidy(parsed$names, env = env)
```

An injected empty argument reaches it and raises a bare `simpleError` whose
`conditionCall()` is `NULL`:

```r
f <- function(data, template) {
  summarize_with_margins(
    data,
    dplyr::across(v, sum, .names = {{ template }}),
    .grouping = grouping_sets(region)
  )
}
f(d)
#> Error: object '' not found
```

dplyr's answer to the identical call is an `rlang_error` carrying its own
context — ``In argument: `dplyr::across(v, sum, .names = )`. Caused by error: !
object '' not found`` — so this position answers a caller with strictly less
than dplyr does. A written `.names = ` is unaffected, because R matches a named
formal's empty argument to that formal's default.

The sibling reader in the same file already wraps its `eval_tidy()` in
`tryCatch()`, so what is missing here is a decision rather than a mechanism:
whether an injected empty `.names` is read as omitted, which is what
`parse_across_arguments()` answers for the written spelling, or reaches dplyr
and receives dplyr's own condition. The `across()` tests hold dplyr as the
oracle for every other omitted argument, and those two answers differ here.

#340 does not reach this: `is_empty_argument()` is asked of a verb's or a
constructor's own argument, and this one sits inside a caller's expression.
Found while answering #340's fourth review round; `R/summary-selections.R`'s
`across()` machinery is untouched by that branch and the defect predates it.
Related to #349, which is the other empty-argument position inside a summary
expression, by a different mechanism.

## Acceptance criteria

- [ ] An injected empty `across()` `.names` raises a condition inheriting
      `marginplyr_error`, or reaches dplyr and receives dplyr's own answer with
      its class intact.
- [ ] Which of the two it is, is decided against the dplyr oracle the `across()`
      tests already use, and the decision is written where the reader is.
- [ ] A written `.names = ` keeps the reading it has, which is the formal's
      default.
- [ ] The other `across()` argument readers are checked for the same shape in
      the same pass.

## Blocked by

None — can start immediately.

sayuks:
Closes #350.

A wrapper forwarding a selection injects a quosure into `.cols`, which is the
ordinary tidy-eval idiom. `resolve_summary_selection()` wrapped whatever it was
given in a fresh quosure, so a quosure arrived at `tidyselect::eval_select()`
inside another one, where the inner `~` reads as the lambda shorthand:
`across({{ cols }}, sum)` was refused with tidyselect's `where()` advice
whatever columns it named. A quosure is now passed on, carrying the environment
the wrapping would have supplied.

Forwarding an argument the caller omitted is one value of that spelling, and it
now selects what dplyr selects for it, which is nothing. `rlang::is_missing()`
is false of such a quosure, so the `across()` parse reads it as supplied —
correctly, since dplyr tells the two spellings apart the same way: a written
empty `.cols` takes the formal's default and deprecates, an injected one
resolves to the empty selection. Folding them together at the parse would
answer a call the caller did not write, so the two sites that read the argument
answer it instead.

The second is `.names`. The template is evaluated so that one written against
the caller's own bindings reaches dplyr as its string; an empty quosure has no
value, and evaluating it looked up the empty symbol for base R's untyped
`simpleError` reading `object '' not found`, whose `conditionCall()` was
`NULL`. dplyr raises on the same argument and names it, so the argument is left
for dplyr (ADR 0015).

`.fns`, `.unpack`, an argument forwarded to the function, and `pick()` were
checked in the same pass and already agree with dplyr; tests hold them there.

## Local checks

- `testthat::test_local()` with `NOT_CRAN=true` — passes.
- `lintr::lint_package()` after `pkgload::load_all()` — no lints.
- `roxygen2::roxygenise()` — no diff.
- `spelling::spell_check_package()` — clean.
- `Rscript .github/scripts/verify-context-budget.R` — within budget.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -197,10 +197,17 @@
   selection was read as the formula shorthand for a predicate. Forwarding an
   argument your caller omitted is one value of the same spelling, and it now
   selects no columns, which is what dplyr selects for it; a selection you write
-  empty yourself still means every eligible column, as it did. An omitted
-  `.names` template forwarded the same way reaches dplyr, which names the
-  argument you wrote, in place of R's `object '' not found` naming nothing
-  (#350).
+  empty yourself still means every eligible column, as it did. dplyr reads a
+  nested `across()` — one inside another expression rather than standing as the
+  summary — differently, taking its deprecated `everything()` default for a
+  forwarded omission; marginplyr resolves the selection wherever it is written,
+  so the reading above is the one you get at both positions. A share's
+  `across()` reads its selection the same way, so a forwarded name that is not
+  an eligible preceding summary is now refused by that name rather than
+  reported as an unusable selection. An `across()` `.names` template that
+  cannot be evaluated — forwarded empty, or naming something absent — reaches
+  dplyr, which names the argument you wrote, in place of R's bare
+  `object '' not found` naming nothing (#350).
 * A condition raised while your summary expression runs now reports its
   context in names you can act on. A margin operation summarizes that
   expression once per grouping set, so the grouping values it reported were
--- a/NEWS.md
+++ b/NEWS.md
@@ -190,6 +190,17 @@
   Redundant parentheses are transparent at all three positions, as they are
   elsewhere: an injected pair wrapping nothing is the empty argument it wraps
   (#340).
+* A selection a wrapper forwards into `across()` is now the selection dplyr
+  makes of it. `across({{ cols }}, sum)` — the ordinary way to pass on a
+  selection your own caller wrote — was refused by tidyselect, which advised
+  wrapping it in `where()`, whatever columns it named, because the forwarded
+  selection was read as the formula shorthand for a predicate. Forwarding an
+  argument your caller omitted is one value of the same spelling, and it now
+  selects no columns, which is what dplyr selects for it; a selection you write
+  empty yourself still means every eligible column, as it did. An omitted
+  `.names` template forwarded the same way reaches dplyr, which names the
+  argument you wrote, in place of R's `object '' not found` naming nothing
+  (#350).
 * A condition raised while your summary expression runs now reports its
   context in names you can act on. A margin operation summarizes that
   expression once per grouping set, so the grouping values it reported were
--- a/R/share.R
+++ b/R/share.R
@@ -3037,8 +3037,14 @@ resolve_share_selection <- function(expr,
                                     context,
                                     kind,
                                     error_call) {
-  if (rlang::is_symbol(expr)) {
-    source <- rlang::as_name(expr)
+  # Through the injected reading, so that a selection a wrapper forwarded is
+  # asked the question the written one is asked, rather than reaching
+  # `eval_select()` as an unusable selection (ADR 0019's amendment, #169).
+  # Asked with `is_name_part()` for the reason `validate_share_across_syntax()`
+  # gives, and read twice rather than bound, because what it carries can be the
+  # empty argument.
+  if (is_name_part(unwrap_injected_quosure(expr))) {
+    source <- rlang::as_name(unwrap_injected_quosure(expr))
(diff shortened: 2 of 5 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, one round, against `origin/main` at
`3d4c7d2` (three-dot merge-base) over `8c9d252`. Standards and Spec ran as
parallel sub-agents. Answers are `7d1368f`.

### Standards — 3 breaches, 2 judgement calls

| # | Finding | Disposition |
|---|---|---|
| 1 | `is_injected_empty_argument()`'s header and the test section comment carried the same argument about the two spellings of an empty argument — a second copy nothing compares (*Code comments*, *Answering a review*) | fixed in code: the predicate is gone with the special case it existed for, and the test prose that re-derived it is deleted |
| 2 | The `.names` code comment was reproduced almost sentence-for-sentence in the test | fixed in prose: the test states what it asserts, the site keeps the reason |
| 3 | Two comments narrated the removed defect in the past tense — "the wrapping refused every forwarded selection", "evaluating it looked up the empty symbol" (*Code comments*: present tense, and what a review found is not something the code holds) | fixed in prose |
| 4 | Judgement call — one shape ("an argument I cannot evaluate") answered two ways ten lines apart: `tryCatch` for `.unpack`, a predicate for `.names` | accepted, and it was the stronger finding: `.names` now uses the sibling's answer, which also covers a template naming something absent — that shape got the same bare `simpleError` and no test had reached it |
| 5 | Judgement call — three near-identical `data`/`summarize`/`oracle` triples | fixed: the two `.cols` tests are one |

### Spec — 2 confirmed against the brief, 1 not as described

| # | Finding | Disposition |
|---|---|---|
| 1 | The brief's criterion "the `across()` argument parser's header distinguishes the written empty argument from the injected one … no second copy of that argument is added elsewhere" was met by adding a copy rather than by amending the header | fixed in code, jointly with Standards 1: the distinction has one home, the `.cols` comment at the site that acts on it |
| 2 | A nested `across()`/`if_any()`/`if_all()` with a forwarded omission answers 0 columns where dplyr answers `everything()`. Measured; converts a refusal on `main` into a divergent answer | accepted as a deliberate divergence, not removed. dplyr's own two readings disagree — it expands a top-level `across()` statically and resolves the injected empty `.cols` to the empty selection, while a nested one reaches its run-time `across()` and takes the *deprecated* `everything()` default. Every position is rewritten here, so one reading applies to both; it is the one dplyr gives the position `across()` is written in as a summary. Asserted in `test-static-expression-analysis.R` with dplyr's lifecycle warning beside it, which is the signal saying which of dplyr's two paths answered |
| 3 | "`across({{ sel }}, …)` still fails when a share helper is present … only `{{ }}` fails" | premise measured false in that form, and a real defect underneath it. `!!rlang::ensym(sel)` — which injects a bare symbol and carries no quosure — fails identically, and reproduces on `main` at `3d4c7d2`, so it is neither `{{ }}`-specific nor this branch's. Two things were found in the same place and split: the share planner's *selection reader* had both shapes the summary reader had (#169's `rlang::is_symbol()` test, and the same `eval_select()` wrapping) and is fixed here with a test; the *staging* beneath it re-expands any injection and is #357 |

Scope creep: none found on either axis.

### What this branch does not answer

#357. `AGENTS.md`'s *Answering a review* — a finding this branch's Acceptance
does not reach becomes a ticket.

PR #358, opened 2026-08-31
Fix dtplyr ordered-factor expansion

### LINKED ISSUE #354: expand_with_margins() refuses an ordered factor dimension on dtplyr
## What happens

`expand_with_margins()` on a `dtplyr_step` whose Grouping dimension is an
*ordered* factor fails when the result is collected:

```r
d <- data.frame(
  a = c("x", "y"),
  o = factor(c("s", "l"), levels = c("s", "l"), ordered = TRUE),
  u = 1:2
)
dplyr::collect(expand_with_margins(
  dtplyr::lazy_dt(d),
  .grouping = rollup(a, o),
  .margin_label = "(all)"
))
#> Error in .set_ops_arg_check(x, y, all, block_list = !all) :
#>   Item 2 of x is 'character' but the corresponding item of y is 'ordered'.
```

The condition is `<dt_join_type_mismatch_error>`, raised by data.table's
`.set_ops_arg_check()` from the `union` the expand path builds. The Margin
label is added to a branch as `character` while the source branch still carries
the `ordered` column, and data.table refuses that pairing where it accepts the
unordered one.

## What is not affected

- the same call with `o` unordered succeeds;
- `summarize_with_margins()` on the same dtplyr step succeeds and returns
  `<ord>`;
- the same `expand_with_margins()` call on the local data frame succeeds.

So this is the `dtplyr` expand path alone, and only for `ordered = TRUE`.

## Why it is filed rather than fixed in #343

Found by the review round on #343's branch (PR #352), which is documentation
only: that branch scopes the factor-preservation claim to the backends that
restore factors, and it names `dtplyr` among them. The claim is right about
what the backends decide — `can_restore_factors` is true for `dtplyr` — and
this is a defect in the expand path rather than a wrong sentence, so per
*Answering a review* in `design/architecture.md` it becomes a ticket.

## Acceptance criteria

- [ ] `expand_with_margins()` and the nesting verbs accept an ordered factor
      dimension on `dtplyr` and return it with its levels and ordered status.
- [ ] A test covers the ordered case on `dtplyr` for every verb built on the
      `union` path, and fails without the fix. It requires no backend other
      than `dtplyr`, per the one-backend policy in `AGENTS.md`.
- [ ] Whether the same pairing is reachable on any other backend the portable
      adapter serves is established, and either covered or recorded as not
      reachable.

sayuks:
## Summary

- replace omitted factor dimensions with full-length label values on dtplyr branches so data.table changes the column type before `funion()`
- keep factor reconstruction after the portable union, preserving ordered status, source level order, and Margin-label position
- cover `expand_with_margins()`, `nest_with_margins()`, and `nest_by_with_margins()` through their public interfaces

## Testing

- `Rscript -e 'testthat::test_dir("tests/testthat", filter = "^(margin-label|expand-operation|nest-operation|factor)$", load_package = "source", stop_on_failure = TRUE)'`
- `Rscript -e 'pkgload::load_all(".", quiet = TRUE); print(lintr::lint_package())'`
- `Rscript .github/scripts/verify-context-budget.R`
- `git diff --check`

Closes #354

### DIFF
--- a/R/margin-label.R
+++ b/R/margin-label.R
@@ -382,11 +382,23 @@ label_margin_branch <- function(.data,
       dplyr::across(dplyr::all_of(labelled_as_character), as.character)
     )
   }
+  # dtplyr renders a scalar overwrite as data.table `:=`. On a multi-row factor
+  # column, data.table recycles it without replacing the column class, leaving
+  # an ordered factor opposite the character branch at `funion()`. A full-size
+  # value replaces the column so factor restoration remains after the union.
+  dtplyr_factor_cols <- if (inherits(.data, "dtplyr_step")) {
+    vapply(factor_info, function(info) info$col, character(1))
+  } else {
+    character()
+  }
   values <- lapply(
     omitted,
     function(col) {
       label <- margin_labels[[col]]
       if (!is_missing_margin_label(label)) {
+        if (col %in% dtplyr_factor_cols) {
+          return(rlang::expr(rep(!!label, dplyr::n())))
+        }
         return(label)
       }
       value <- prototypes[[col]]
--- a/tests/testthat/test-margin-label.R
+++ b/tests/testthat/test-margin-label.R
@@ -455,6 +455,57 @@ test_that("dtplyr applies mixed named labels lazily and restores factors", {
   expect_true(all(is.na(result$second[result$id == 1L])))
 })
 
+test_that("dtplyr expansion verbs restore ordered factor dimensions", {
+  skip_if_suggest_absent("dtplyr")
+  # Two rows are load-bearing: data.table replaces a one-row factor column but
+  # recycles a scalar assignment into a multi-row one without replacing it.
+  data <- data.frame(
+    group = ordered(c("small", "large"), levels = c("small", "large")),
+    value = 1:2
+  )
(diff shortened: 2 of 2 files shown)


sayuks:
## Review record

One round, `mattpocock-skills:code-review`, two axes over
`git diff main...HEAD` at merge-base `60fa93c`, reading `5769ac8`.

### Standards

No findings. The dtplyr-specific full-length label expression remains in the
Margin-label module that owns applying labels in portable branches; factor
reconstruction remains at its existing finalization boundary. The change adds
no lazy execution entry point, the regression requires only the `dtplyr`
optional backend, and both new comments state the location-specific data.table
behavior their code and fixture depend on. No baseline code smells were found.

### Spec

No findings. The diff covers all three expansion-union verbs on a two-row
dtplyr input, keeps the two lazy-returning verbs lazy, and asserts ordered
status, source-level order, Margin-label position, and visible values. The
implementation makes the omitted factor branch character-compatible before
`funion()` without changing a public interface or another backend.

Summary: Standards 0 findings; Spec 0 findings. The full suite passed after
the round: 5,925 assertions, 0 failures, 0 warnings, 0 skips.

PR #359, opened 2026-08-31
Read an empty argument inside a selection as tidyselect reads one (#351)

### LINKED ISSUE #351: Read an empty argument inside a .by or .grouping selection as tidyselect reads one
## What to build

An empty argument inside a column selection is answered as tidyselect answers
it, rather than by R raising for a zero-length variable name.

`is_name_only_expr()` walks the parts of a selection and asks each one whether
it names a column. R's empty argument is a symbol whose name is `""`, so
`rlang::env_has()` raises:

```r
summarize_with_margins(d, n = dplyr::n(), .by = c(, region), .grouping = rollup(grade))
#> Error: attempt to use zero-length variable name
```

Both positions the empty argument can take are affected — `c(, region)` and
`c(region, )` — and so are all five public verbs. `dplyr::summarise()` groups by
`region` for both.

This is not the empty `.by` #340 answered: there the argument itself is empty,
and `resolve_fixed_keys()` now selects no columns for it. Here the argument is a
selection that names a column, with an empty argument one level down, so
`is_empty_argument()` is not asked and `is_name_only_selection()` reads it.
Pre-existing on `main` and unchanged by that branch, which is why it is filed
rather than folded into it.

The same reader serves `.grouping`, so the decision has a second half:
`rollup(c(, region))` reaches it too, and a Grouping specification constructor
already refuses `rollup(, region)` by position. Whether a selection *inside* a
constructor argument follows tidyselect or that refusal is what this ticket has
to settle; the two answers are not interchangeable.

Found while answering #340's sixth review round. Related to #349 and #350, the
other two empty-argument positions that sit inside a caller's expression rather
than at an argument of a verb.

## Acceptance criteria

- [ ] An empty argument inside a `.by` selection is answered as
      `dplyr::summarise()`'s `.by` answers it, on all five public verbs and in
      every position of the selection.
- [ ] What an empty argument inside a Grouping specification argument's
      selection is answered with is decided explicitly, against the refusal
      `preflight_grouping_spec()` already gives for an empty constructor
      argument, and the decision is written where the reader is.
- [ ] `is_name_only_expr()` raises no untyped condition for any part a caller
      can write or inject.
- [ ] The existing refusal of an empty constructor argument is unchanged,
      including that a trailing comma is not one.

## Blocked by

None — can start immediately.

sayuks:
Closes #351.

`is_name_only_expr()` read a name off every symbol, and R's empty argument is a symbol whose name is `""`, so `rlang::env_has()` raised `attempt to use zero-length variable name` for any selection carrying one. Reproduced on `main` for `.by = c(, region)`, `c(region, )`, `c(region, , grade)`, and for the same selection written at a Grouping specification argument — `rollup(c(, grade))`. All three spellings parse, so none needed injection to reach the reader.

This was the last site in the package reading a name off a symbol without first asking whether the part names anything; every other one already goes through `is_name_part()`.

## What the reader answers

`TRUE`, which is what its contract already gives an empty part: the column names settle it in both directions. tidyselect drops one under `c()` and fails on one under every other walk operator, exactly as `dplyr::select()` does, and neither branch reads a column's data. `investigation/an-empty-argument-under-a-selection-walk.md` measures which per operator, and records that the non-`c()` failures are tidyselect's rather than a divergence to close.

Refusing an empty part with a Package condition was weighed and rejected: it would make this package stricter than dplyr for shapes dplyr accepts, and would break the reported case.

## The boundary it settles for `.grouping`

`CONTEXT.md`'s *Nested specification position* now says that the refusal of an empty constructor argument reaches the argument's own position and not what a selection written there contains. So `rollup(, grade)` is refused, unchanged, and `rollup(c(, grade))` is not. Its existing test passes unmodified, which is what carries that half.

## Tests

- `is_name_only_expr()` answered directly for the part and through a walk in all three positions.
- End to end over every verb `verbs_taking(".by")` derives — six, not the five the issue body says; it undercounts `inspect_grouping()`.
- `rollup(c(, grade))`, `rollup(c(grade, ))`, and one nested `grouping_sets(grouping_set(c(, region)))`.
- For `-`, `!`, `|`, and `/` carrying an empty operand, that no `marginplyr_error` is raised. The message is tidyselect's and version-dependent, so it is not snapshotted here; the note holds it with the versions it was measured on.

Full suite: 6114 passing, 0 failures, 0 skips. `lintr::lint_package()` clean. Context budget unchanged (`CONTEXT.md` is not in the always-loaded closure).

No `man/`, `NAMESPACE`, or `README.md` change — nothing exported or documented changed.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -26,7 +26,10 @@ function. A spelling the position does not recognize is never
 evaluated to find out, because a selection such as `starts_with("re")` has no
 meaning outside a selection context. An argument the caller left empty gets
 neither reading, and is refused naming the constructor and the position; a
-trailing comma is not one, because R captures no argument for it. Both
+trailing comma is not one, because R captures no argument for it. That refusal
+reaches the position and not what a selection written there contains: an empty
+argument inside such a selection leaves the position filled, and is read as
+tidyselect reads one. Both
 readings claim one argument at once: a bare name that is a column of the
 input and is also bound, in the caller's environment, to a specification of a
 kind the position admits — which kinds those are varies by constructor. How
--- a/NEWS.md
+++ b/NEWS.md
@@ -189,7 +189,12 @@
   get from omitting it, which is what writing `.grouping = ` already means.
   Redundant parentheses are transparent at all three positions, as they are
   elsewhere: an injected pair wrapping nothing is the empty argument it wraps
-  (#340).
+  (#340). An empty argument one level further down — inside a selection, as in
+  `.by = c(, region)` or `rollup(c(, grade))` — is read as tidyselect reads
+  one, which under `c()` is to select nothing for it, in place of the same
+  `attempt to use zero-length variable name`. The refusal above reaches a
+  constructor argument's own position and not a selection written there, so
+  `rollup(, grade)` is still refused (#351).
 * A selection a wrapper forwards into `across()` is now the selection dplyr
   makes of it. `across({{ cols }}, sum)` — the ordinary way to pass on a
   selection your own caller wrote — was refused by tidyselect, which advised
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -349,25 +349,9 @@ selection_refused_operators <- function() {
 # hold `env` as the expression's own environment and `data_vars` as the input's
 # column names, which is the pair the symbol branch decides from.
 is_name_only_expr <- function(expr, env, data_vars) {
-  # R's empty argument, answered before anything asks what the part names --
-  # it is a symbol whose name is `""`, so the branch below would put that name
-  # to `rlang::env_has()`, which raises for a zero-length variable name
-  # (#351). This is the site `is_name_part()`'s header names: the rule is
-  # satisfied by answering the empty argument outright rather than by
-  # declining it.
-  #
-  # `TRUE` is what the contract above gives it. tidyselect settles an empty
-  # part from the names in both directions -- it drops one under `c()` and
-  # fails on one under every other walk operator, as `dplyr::select()` does --
-  # and `investigation/an-empty-argument-under-a-selection-walk.md` measures
-  # which per operator. A selection written at a Grouping specification
-  # argument reaches here too, and `CONTEXT.md`'s *Nested specification
-  # position* holds why that is not the empty constructor argument
-  # `preflight_grouping_spec()` refuses.
-  if (rlang::is_missing(expr)) {
-    return(TRUE)
-  }
-  if (is.symbol(expr)) {
+  # `is_name_part()` rather than `rlang::is_symbol()`, for the reason its
+  # header gives: a part read by subscript may be R's empty argument.
+  if (is_name_part(expr)) {
     name <- as.character(expr)
     return(
       name %in% data_vars ||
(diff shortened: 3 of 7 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel, against the merge-base with `origin/main` (`46c8467...d1f6018`). Answered in `3c518da`.

The fixed point needed repinning: the local `main` was at `60fa93c`, so `main...HEAD` carried #346, #349, #350, and #354 as though they were this branch's — the gap `design/agents/code-review.md` names.

### Standards — 7 raised, 5 answered in code, 2 rejected

| finding | disposition |
|---|---|
| The sentence added to `is_name_part()`'s header re-derived nothing checkable and claimed uniqueness: "`is_name_only_expr()` is the one site that satisfies the rule without asking this" | **False, deleted.** `call_part_label()` and `check_grouping_helper_columns()` both answer the empty argument with `rlang::is_missing()` without asking `is_name_part()` |
| The reader's comment re-derived the `""` argument `is_name_part()`'s header owns | **Fixed in code.** The branch now asks `is_name_part()`, so the ordering is local and the comment carries only its two citations |
| The investigation note described this repository's present — "which is the contract `is_name_only_expr()` states", "not a marginplyr divergence to close", "what the reader's answer rests on" | **Fixed in prose.** Deleted. `investigation/README.md`: a note "is not a description of how things currently are" |
| The note's per-operator conclusion was restated in three artifacts | **Fixed in prose.** The note holds the table; the reader and both tests cite it |
| The new test built unary `\|( )` and `/( )` where the note measured `\|(, region)` and `/(, region)` | **Fixed in code.** Both raised the same `simpleError`, so the assertion held, but the shape under test was not the one recorded |
| Duplicated Code — "the third near-identical six-verb table (873, 969, new)" | **Miscounted, and fixed anyway.** There were two, not three; `969` was this branch's own. The table is now file level and shared, and this branch adds none |
| Graduation: `R/` now rests on a note, so the durable part belongs in `AGENTS.md` or an ADR | **Rejected.** Answered by the finding above it. With the comment cut to citing the note for the measurement, the durable claim is the reader's own contract, which its header already states; what the note supplies is evidence, which is what a note is authoritative for |

### Spec — 5 raised, 4 answered in code, 1 rejected

| finding | disposition |
|---|---|
| "asserted directly against the function and not only end to end" — the direct test covered only `rlang::missing_arg()` and the three `c()` spellings | **Fixed in code.** Each shape the note measured is now put to the reader directly |
| No `NEWS.md` bullet, where #340, #349, and #350 each carry one | **Fixed in prose.** The empty-argument bullet gains the position one level down and the boundary against the constructor refusal |
| The reader's comment was 14 lines and re-derived the per-operator argument instead of quoting the note once | **Fixed in code**, with the Standards finding of the same shape |
| Unary vs binary operator shapes | **Fixed in code**, as above |
| The note adds a `( )` row absent from the brief's table, for a shape *Out of scope* reserves to #340 | **Rejected.** The row is a true measurement of tidyselect and the note is authoritative for exactly that. What marginplyr does with `( )` is #340's decision and is recorded where #340 recorded it; the note recording it too would be the second copy. The operator test now says why `( )` is excluded from it |

### After

`6124` passing, 0 failures, 0 skips. `lintr::lint_package()` clean. Always-loaded context 24198/24198, unchanged.

Issue #360, opened 2026-08-31
A dtplyr share re-parses a caller call that deparse cannot round-trip


sayuks:
> *This was generated by AI during triage.*

## What happens

A dtplyr share carries the caller's verb call into the summary as deparsed text and re-parses it with `str2lang()` when it raises. A call holding an object rather than the source that produced it does not survive that round trip: the re-parse raises R's own parse error, so the caller receives an internal error instead of the Package condition the check exists to raise.

`do.call()` is the ordinary way to reach it — it records the evaluated arguments in the call:

```r
d <- data.frame(region = c("East", "East", "West"), units = c(1, 3, 6))
dt <- dtplyr::lazy_dt(d)

do.call(summarize_with_margins, list(
  dt,
  units = quote(range(units)),
  units_share = quote(share_of_total(units)),
  .grouping = rollup(region)
))
#> Error in str2lang(call_text) : <text>:20:183: unexpected '<'
#> 19:     finalize_margin_operation(operation, execution)
#> 20: })(structure(list(parent = structure(list(region = c("East", "East", "West"), units = c(1, 3, 6)), row.names = c(NA, -3L), class = c("data.table", "data.frame"), .internal.selfref = <
```

The same call on a local input raises what it should:

```r
do.call(summarize_with_margins, list(
  d,
  units = quote(range(units)),
  units_share = quote(share_of_total(units)),
  .grouping = rollup(region)
))
#> Error: Total share `units_share` requires source summary `units` to return
#>   exactly one value per grouping row.
```

Measured on `main` at `46c8467`.

## Why dtplyr carries text at all

A dtplyr share raises at `collect()`, outside the verb call and outside the handler that sets a Package condition's call for everything raised inside it, so the call has to travel with the expression. Text is how it travels today. The failure is that `deparse()` is not injective over calls holding objects — an environment, a connection, a `data.table` — and `str2lang()` refuses what it produces.

Two directions worth weighing, neither settled:

- carry the call as an object the expression holds as a constant rather than as language, so nothing re-parses it (a language object embedded inside a non-language constant is not descended into by rlang's unquote walker, which is what makes this different from `quote()`);
- keep the text but stop the re-parse being able to fail — fall back to the un-parsed call, or to no call, rather than raising R's own error over the caller's diagnostic.

The second is the smaller change and degrades rather than crashes; the first removes the round trip. Whichever is chosen, ADR 0015 is the constraint: an internal parse failure must not replace a Package condition.

[...]

PR #361, opened 2026-08-31
Stage a share without the caller's own call (#357)

### LINKED ISSUE #357: Stage a share across() without re-expanding an injected selection

## What to build

A summary a wrapper injects into a share `across()` reaches dplyr as the
expression the caller wrote, rather than as one re-expanded in a mask.

Every injection into a share `across()` fails, whatever it carries and whichever
spelling writes it:

```r
d <- data.frame(region = c("East", "East", "West"), units = c(1, 3, 6))

f <- function(data, selection) {
  summarize_with_margins(
    data,
    units = sum(units),
    dplyr::across({{ selection }}, share_of_total, .names = "{.col}_share"),
    .grouping = rollup(region)
  )
}
f(d, units)
#> Error: object 'selection' not found

g <- function(data, selection) {
  summarize_with_margins(
    data,
    units = sum(units),
    dplyr::across(
      !!rlang::ensym(selection),
      share_of_total,
      .names = "{.col}_share"
    ),
    .grouping = rollup(region)
  )
}
g(d, units)
#> Error: object 'selection' not found
```

`!!rlang::ensym()` injects a bare symbol and carries no quosure, so this is not
the injected-quosure reading: what fails is downstream of the selection. The
share staging wraps the caller's `across()` in `check_share_across()` and passes
`call = rlang::call2("quote", call)`, where `call` is the caller's own verb call
— which still holds the unexpanded `{{ selection }}` or `!!rlang::ensym(...)`.
The staged dot is spliced back with `!!!` and defused again by
`dplyr::summarise()`, and the second defuse expands the injection in a mask
where the wrapper's formal is not bound. The condition names that formal, which
is a name the caller did write but not one they can bind where the message
sends them.

An ordinary (non-share) summary is unaffected: nothing re-quotes the caller's
call into the staged expression there.

This predates #350's branch. `!!rlang::ensym()` reproduces on `main` at
`3d4c7d2`; `!!rlang::enquo()` and `{{ }}` reached a different refusal there,
which #350 removed, so all three spellings now reach this one. #350's branch is
where it was found and where the share selection reader was checked; the reader
is correct and this is the staging beneath it.

## Acceptance criteria

- [ ] A share `across()` whose selection is injected — `{{ }}`,
      `!!rlang::enquo()`, or `!!rlang::ensym()` — computes the shares it names,
      as the written spelling does.
- [ ] The same holds for a direct share summary, `name = share_of_total(!!q)`,
      if it is reached by the same staging.
- [ ] No condition a share raises names a binding from the caller's frame that
      the caller cannot act on, in place of the summary they wrote.
- [ ] Whatever the quoted caller call is carried for keeps working: the
      diagnostics that quote the caller's own spelling (ADR 0022, ADR 0024) name
      what they named before.
- [ ] Covered on every backend a share reaches, within the one-backend-per-test
      policy.

## Blocked by

None — can start immediately.

sayuks:
Closes #357.

A wrapper forwarding a selection or a source into a share is the ordinary tidy-eval idiom, and every spelling of it failed on the local backend. The share staging wrapped each source expression in a check carrying the caller's own verb call as `quote(<call>)` — which still holds `{{ selection }}` or `!!rlang::ensym(...)` unexpanded — and `dplyr::summarise()` defuses the staged dot a second time, expanding the injection in a mask where the wrapper's formal is not bound.

The call is removed rather than re-encoded. On the local backend it decides nothing: replacing it with a sentinel and running all four local paths reports the caller's verb call every time, never the sentinel, because `with_margin_error_call()` already sets `$call` on any Package condition raised inside the operation. The same sentinel on the dtplyr path does surface — a dtplyr share raises at `collect()`, after the verb call has returned — so there the call still travels with the expression, as text.

`check_share_across()` therefore loses its `call` argument outright; `check_share_scalar()` keeps one defaulting to `NULL`, supplied by `check_dtplyr_share_source()`, which is its one caller that has to name a call itself.

## Tests

- Local: an injected share `across()` selection and an injected direct share source, each over `{{ }}`, `!!rlang::enquo()`, and `!!rlang::ensym()`, compared against the written spelling.
- Local: the whole call a share's condition reports, for a plain wrapper and for one whose call holds the injection — this is what makes the removal safe rather than merely green.
- dtplyr and DuckDB: an injected share `across()`. Both passed before this change, for reasons that are not the local backend's; nothing else would report a change to either.

`lintr::lint_package()` clean, full suite green with `NOT_CRAN=true`.

## Not in this PR

The dtplyr `str2lang(call_text)` failure filed as #360 — a caller call holding an inlined object does not survive the deparse round trip. Different cause, not fixed here.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -208,6 +208,16 @@
   cannot be evaluated — forwarded empty, naming something absent, or a pair of
   parentheses holding nothing — reaches dplyr, which names the argument you
   wrote, in place of R's bare error naming nothing (#349, #350).
+* A selection or a source your wrapper forwards into a share now reaches the
+  share it names. `across({{ cols }}, share_of_total)` and
+  `share_of_total({{ col }})` — with the `!!rlang::enquo()` and
+  `!!rlang::ensym()` spellings of each — reported `object 'cols' not found`,
+  naming your own wrapper's argument in place of the summary you wrote: the
+  check marginplyr wraps around a share's source carried your call back into
+  the summary, where dplyr defused it a second time and expanded the
+  forwarding where that argument is not bound. The check carries no call
+  there now, and a share's conditions still name the call you wrote,
+  forwarding and all (#357).
 * A condition raised while your summary expression runs now reports its
   context in names you can act on. A margin operation summarizes that
   expression once per grouping set, so the grouping values it reported were
--- a/R/share.R
+++ b/R/share.R
@@ -970,6 +970,15 @@ wraps_share_sources_in_summary <- function(backend_kind) {
   backend_kind %in% c("local", "dtplyr")
 }
 
+# The dots with each share's source expression wrapped in the check that
+# applies the eligible-type and cardinality rules to it. Only dtplyr is given
+# the caller's call here, and as text: its checks raise at `collect()`, after
+# the verb call has returned, so nothing else can supply one. A local check
+# raises inside the verb call, where `with_margin_error_call()` sets the call
+# -- and staging one into the expression is worse than redundant there,
+# because `dplyr::summarise()` defuses the staged dot a second time and
+# expands whatever injection the caller's own spelling holds, in a mask where
+# their wrapper's formal is not bound (#357).
 wrap_share_sources <- function(dots,
                                cardinality,
                                call,
@@ -1019,8 +1028,7 @@ wrap_share_sources <- function(dots,
         share_private_call("check_share_across"),
         expr,
         share_outputs = share_outputs,
-        share_kinds = share_kinds,
-        call = rlang::call2("quote", call)
+        share_kinds = share_kinds
       )
     } else {
       check <- checks[[1L]]
@@ -1035,11 +1043,7 @@ wrap_share_sources <- function(dots,
         share_output = check$share_output,
         source_summary = check$source_summary,
         share_kind = check$share_kind,
-        !!!if (is_dtplyr) {
-          list(call_text = share_call_text(call))
-        } else {
-          list(call = rlang::call2("quote", call))
-        }
+        !!!if (is_dtplyr) list(call_text = share_call_text(call))
       )
     }
     dots[[position]] <- rlang::new_quosure(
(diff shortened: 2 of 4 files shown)


sayuks:
> *This was generated by AI.*

## Review record

One round, `mattpocock-skills:code-review`, against the merge-base with `main` (`46c8467`), read at `6ed669a`. Answers are at `4271eb8`.

**Standards — 4 findings.**

| finding | disposition |
|---|---|
| `check_share_scalar()` gains `call = NULL` with no header, so the invariant that makes the default safe is only readable 270 lines upstream (`AGENTS.md` *Code comments*: an internal name's header is the only place its contract is written) | fixed in code — the function has a header now |
| the commit message said the local post-collection re-check supplies `check_share_scalar()`'s `call`; it calls `abort_share_source_type()` directly and never reaches it | measured, finding correct; fixed in prose — the commit message and this PR's body name `check_dtplyr_share_source()` as the one supplier |
| Duplicated Code — the two new backend tests carry an identical `summarize` wrapper and fixture | rejected. Every backend test in that file defines its own local `summarize`, including the two Total-share tests directly above these; hoisting one wrapper out would leave the file with two conventions |
| `!!!if (is_dtplyr) list(...) else list()` splices one argument or none and reads heavier than the conditional it now is | fixed in code — `!!!if (is_dtplyr) list(...)`, since `!!!NULL` splices nothing |

**Spec — 3 findings, 0 blocking.** Acceptance criteria 1–6 all met, verified independently of the tests.

| finding | disposition |
|---|---|
| the brief asked for the `check_wrapped_share_sources()` comment to be corrected, and it was not | rejected, and the brief was wrong to ask. The sentence scopes itself to "the backends that evaluate summaries in R **and stay lazy**" and says the rule "raises at execution with the caller's own call" — true of dtplyr, which carries the call as text, and true of what a local caller receives, since the handler names the same call. Nothing this change made false |
| the `test-share.R` block comment re-derives the second-defusal mechanism that `wrap_share_sources()`'s header owns | fixed in prose — the comment now cites that header and states only what the two tests below it are over |
| the brief's "the dtplyr source check and the local post-collection re-check both supply one" is inaccurate | same measurement as the Standards finding above; the acceptance criterion itself (default `NULL`) was met |

**Noted, not actioned.** `parent_sql_count()` in the backend share tests returns `1` for an absent pattern, because `gregexpr()` answers `-1` and `lengths()` counts that as one element. The batching assertion in *dtplyr batches validated summaries and parent mapping* compares two such counts for `"check_share_scalar(sum(revenue)"`, a string the dtplyr rendering has never held — it renders `check_dtplyr_share_source()` — so both sides are `1` and the comparison passes vacuously. Measured on this branch and unchanged from `main`; this diff neither caused it nor touches that test.

PR #363, opened 2026-09-01
Count an absent pattern as none (#362)

### LINKED ISSUE #362: parent_sql_count() answers 1 for a pattern it does not find, and one share assertion reads it

## What happens

`parent_sql_count()` in the backend share tests counts a pattern's occurrences as

```r
parent_sql_count <- function(sql, pattern) {
  lengths(gregexpr(pattern, sql, fixed = TRUE))
}
```

`gregexpr()` answers `-1` for a pattern it does not find, and `lengths()` counts that as one element. So the helper returns `1` for a pattern that occurs once **and** for a pattern that does not occur at all, and no caller can tell the two apart.

One caller is already reading the wrong side of that. *dtplyr batches validated summaries and parent mapping* asserts that a two-share query stages the same number of source checks as a one-share query:

```r
expect_identical(
  parent_sql_count(many_call, "check_share_scalar(sum(revenue)"),
  parent_sql_count(one_call, "check_share_scalar(sum(revenue)")
)
```

The dtplyr rendering has never held that string — it stages `check_dtplyr_share_source(sum(revenue), ...)`, `check_share_scalar()` being the local backend's wrapper — so both sides are `1`, the comparison passes, and the batching property it exists to hold is asserted by nothing.

Measured on `main` at `c6ab914`:

```r
parent_sql_count("abc", "zzz")
#> [1] 1

# on the query the test builds
grepl("check_share_scalar(sum(revenue)", many_call, fixed = TRUE)
#> [1] FALSE
grepl("check_dtplyr_share_source(sum(revenue)", many_call, fixed = TRUE)
#> [1] TRUE
```

## The rest of the exposure

Every other caller compares against `1L`, which is the one value the helper cannot distinguish from absence:

- `expect_identical(parent_sql_count(many_call, "allow.cartesian = TRUE"), 1L)` — the string is present, so the assertion happens to be true today, but it would keep passing if the join stopped emitting it.
- `expect_identical(parent_sql_count(many_sql, "LEFT JOIN"), 1L)`, in the general-dbplyr test and again in the fallback-simulator test — same shape, present today.

The comparison sites that read a count against another count (`"GROUP BY GROUPING SETS"`) are only vacuous when the pattern is absent from both, which is the case above.

## What a fix has to answer

The helper's contract is a count, so `0` is the answer for an absent pattern; `sum(gregexpr(...)[[1]] > 0L)` gives it, and the assertions that currently read `1L` then mean what they say. Whether the batching assertion's pattern should be corrected to what dtplyr renders, or the assertion rewritten against something the staged expression carries more durably than a deparsed function name, is the open question — a name in a `fixed = TRUE` string is what let this go unnoticed through the renames that produced `check_dtplyr_share_source()`.

## Acceptance criteria

- [ ] `parent_sql_count()` returns `0` for a pattern its input does not contain.
- [ ] The batching assertion in *dtplyr batches validated summaries and parent mapping* fails when the query stages a different number of source checks than it claims, and names a string the dtplyr rendering actually holds.
- [ ] Every existing `parent_sql_count()` assertion still passes, or is corrected in the same change where it does not.
- [ ] A test holds the helper itself to answering `0` for an absent pattern, since that is the property no caller can observe.

## Found by

The review round on #361 (#357's branch), which read the assertion while checking that removing the staged `call` argument had not changed what that test matches. It had not: the pattern has never matched.

sayuks:
Closes #362.

The counting helper in the backend share tests counted with `lengths(gregexpr(...))`. `gregexpr()` answers `-1` for a pattern it does not find and `lengths()` counts that as one element, so the helper returned `1` for one occurrence and for none alike.

*dtplyr batches validated summaries and parent mapping* was reading the wrong side of that: it matched `check_share_scalar(sum(revenue)`, the local backend's wrapper, while a dtplyr share stages `check_dtplyr_share_source()`. The helper answered `1` on both sides, and the batching property was asserted by nothing.

Correcting the helper alone would not have fixed it — both sides become `0`, and `0 == 0` holds for the same reason `1 == 1` did. So the assertion also names what the rendering holds, and holds the count away from zero as well as equal. Checked by renaming the staged function in the pattern: the assertion fails, where before it passed.

## What changed

- The helper counts through `regmatches()`, which drops what did not match. Its header states the contract and the obligation it puts on a caller comparing two of its counts.
- A test holds it to answering `0` for an absent pattern. No assertion that uses it can observe that property, which is how this survived.
- The batching assertion names `check_dtplyr_share_source(sum(revenue)` and asserts `expect_gt(staged_checks, 0L)` alongside the equality. No occurrence count is written down — four is what both sides measure today, one per rollup branch, and that follows from the plan's shape rather than from batching.
- The `LEFT JOIN` count spelled out with `regmatches()` in the same file was the corrected helper written inline; it calls the helper now. That call is from a Total share's test, which is what makes `parent_sql_count` the wrong name — it is `share_query_count` now, since every caller is a share test and one of the patterns is a data.table argument rather than SQL.

## Weighed and not taken

Parsing the rendering into an expression tree, reaching dtplyr's unexported call builder, and counting invocations with mocked bindings. `show_query()` is dtplyr's public API, the call builder is a Suggest's internal, and counting invocations would turn a data-independent structural claim into a runtime one. What made the assertion vacuous was the missing positivity check, not what it read.

Tests only; no user-visible behaviour changes, so no `NEWS.md` entry. `lintr::lint_package()` clean, full suite green with `NOT_CRAN=true`.

### DIFF
--- a/tests/testthat/test-share-backends.R
+++ b/tests/testthat/test-share-backends.R
@@ -548,9 +548,25 @@ test_that("dtplyr batches Parent shares with missing-safe matching", {
   )
 })
 
-parent_sql_count <- function(sql, pattern) {
-  lengths(gregexpr(pattern, sql, fixed = TRUE))
+# How many times a rendered query contains a pattern, and zero where it
+# contains none. `gregexpr()` answers `-1` for a pattern it does not find
+# and `lengths()` counts that as one element, so the count goes through
+# `regmatches()`, which drops what did not match (#362).
+#
+# A caller comparing one of these counts with another asserts that the count is
+# positive as well: an equality on its own is satisfied by two absences. One
+# comparing against a positive literal already has.
+share_query_count <- function(sql, pattern) {
+  lengths(regmatches(sql, gregexpr(pattern, sql, fixed = TRUE)))
 }
+
+test_that("a pattern is counted once per occurrence, and none where absent", {
+  # Held here because no caller can observe it: an absence counted as one
+  # occurrence reads exactly like one occurrence.
+  expect_identical(share_query_count("SELECT one FROM two", "absent"), 0L)
+  expect_identical(share_query_count("SELECT one FROM two", "one"), 1L)
+  expect_identical(share_query_count("one and one", "one"), 2L)
+})
 
 test_that("dtplyr batches validated summaries and parent mapping", {
   skip_if_suggest_absent("dtplyr")
@@ -578,17 +594,13 @@ test_that("dtplyr batches validated summaries and parent mapping", {
   one_call <- paste(capture.output(dplyr::show_query(one)), collapse = "\n")
   many_call <- paste(capture.output(dplyr::show_query(many)), collapse = "\n")
 
-  expect_identical(
-    parent_sql_count(
-      many_call,
-      "check_share_scalar(sum(revenue)"
-    ),
-    parent_sql_count(
-      one_call,
-      "check_share_scalar(sum(revenue)"
-    )
-  )
-  expect_identical(parent_sql_count(many_call, "allow.cartesian = TRUE"), 1L)
+  # A dtplyr share stages `check_dtplyr_share_source()`. `check_share_scalar()`
+  # is the local backend's wrapper and appears in no dtplyr rendering.
+  staged_check <- "check_dtplyr_share_source(sum(revenue)"
+  staged_checks <- share_query_count(many_call, staged_check)
+  expect_gt(staged_checks, 0L)
+  expect_identical(staged_checks, share_query_count(one_call, staged_check))
+  expect_identical(share_query_count(many_call, "allow.cartesian = TRUE"), 1L)
 })
 
 parent_lazy_probe_capture <- new.env(parent = emptyenv())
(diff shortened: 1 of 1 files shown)


sayuks:
> *This was generated by AI.*

## Review record

One round, `mattpocock-skills:code-review`, against the merge-base with `main` (`c6ab914`), read at `486c7b3`. Answers are at `0e14793`.

**Standards — 5 findings.**

| finding | disposition |
|---|---|
| the new test's comment and the batching test's comment narrate what a review found and describe a different test, against `AGENTS.md` *Code comments* ("A comment describes the code it sits beside, in the present tense") | fixed in prose — both are present-tense statements about the site now |
| the positivity reasoning appears in the helper's header and again at the batching test, which is the second copy that rule refuses | fixed in prose — the reason lives in the header, the test site states only which name a dtplyr share stages |
| `parent_sql_count` is the wrong name: nothing in it is parent-specific, this diff adds a call from a Total share's test, and one pattern it counts is a data.table argument rather than SQL | fixed in code — `share_query_count`. In scope because this diff is what widened its reach |
| the pattern string is written twice inside one test | fixed in code — bound once and passed to both counts |
| the helper's contract paragraph states a rule for callers that nothing enforces | rejected. `AGENTS.md` *Code comments* makes a header "what it answers and what its caller holds before calling it" — a caller obligation is the documented shape of a header, not an unenforced rule out of place |

**Spec — 1 finding, 0 blocking.** All eight acceptance criteria met, verified independently of the suite: the helper's three answers, both sides counting 4, the mutation failing at the positivity assertion, and no remaining comparison two absences would satisfy.

| finding | disposition |
|---|---|
| "matching that here counted zero on both sides" misstates why the old assertion passed — the helper *answered* `1` on both sides; zero was the true occurrence count | fixed in prose — that sentence is gone with the historical comment above, and the commit message says the helper answered `1` |
| the helper header's opening was ungrammatical | fixed in prose |

**Confirmed by the Spec axis, not a finding.** The `"GROUP BY GROUPING SETS"` comparison the brief scoped out is held away from zero by the `expect_match()` on the line above it; every other call compares against `1L`, which asserts positivity once an absence answers `0`. No other `lengths(gregexpr(` survives anywhere in the repository.

PR #364, opened 2026-09-01
Write a dtplyr share's call as source it can read back (#360)

### LINKED ISSUE #360: A dtplyr share re-parses a caller call that deparse cannot round-trip

sayuks:
Closes #360.

A dtplyr share carries the caller's call to `collect()` as deparsed text and reads it back with `str2lang()`. `deparse()` is not total over calls, so a call holding the input itself — as `do.call()` records — was answered with R's own parse error in place of the share's Package condition.

The text is made total rather than the read guarded: the call is walked before it is written, and every part `deparse()` cannot write as source is replaced by a name spelling that part's class. The test is a trial rather than a type, so a `data.frame` or a Grouping specification, both of which round-trip, is left exactly as the caller wrote it.

The alternative #360 raised — carrying the call as an object — was measured and is closed: `dtplyr:::dt_squash()` raises `abort("Invalid input")` for anything that is not atomic, `NULL`, a symbol, a quosure, or a call, and descends into and rewrites a bare call object.

`.check_share_source = FALSE` did not avoid the crash, so there was no workaround short of not using `do.call()`.

## Checks

- `testthat::test_dir("tests/testthat")` under `NOT_CRAN=true`: 6165 pass, 0 fail, 0 skip
- `lintr::lint_package()` after `pkgload::load_all()`: no lints
- No roxygen, `README.Rmd`, or `DESCRIPTION` change, so nothing generated needs regenerating

### DIFF
--- a/R/share.R
+++ b/R/share.R
@@ -1399,36 +1399,55 @@ abort_share_source_type <- function(value,
 # source. The parts that fail are replaced before the call is written, rather
 # than the read being allowed to fail (#360).
 share_call_text <- function(call) {
-  text <- deparse_call_text(round_trippable_call(call))
-  # An invariant, not a Package condition (ADR 0015): `round_trippable_call()`
-  # answers every part that does not parse back, so a text that still does not
-  # is a defect here rather than something a call rewrite avoids. It is stated
-  # at the site that writes the text, inside the verb call, because the site
-  # that reads it is past the point where stopping is affordable.
-  stopifnot(!is.null(parse_call_text(text)))
+  text <- deparse_call_text(writable_call(call))
+  # An invariant, not a Package condition (ADR 0015): `writable_call()` answers
+  # every part that does not read back, so a text that still does not is a
+  # defect here. It is stated at the site that writes the text, inside the verb
+  # call, because the site that reads it is past the point where stopping is
+  # affordable.
+  stopifnot(call_text_reads_back(text))
   text
 }
 
+# One expression as the source that would write it.
 deparse_call_text <- function(expr) {
   paste(deparse(expr, width.cutoff = 500L), collapse = "\n")
 }
 
+# Whether `text` can be read back as one expression.
+#
+# Asked instead of testing `parse_call_text()` for `NULL`, because `NULL` is
+# also what the text `"NULL"` reads as. A caller's `.margin_label = NULL` is a
+# part `deparse()` writes perfectly well, and conflating the two answers put a
+# name in its place.
+call_text_reads_back <- function(text) {
+  tryCatch(
+    {
+      str2lang(text)
+      TRUE
+    },
+    error = function(cnd) FALSE
+  )
+}
+
 # The call a text stands for, or `NULL` where the text cannot be read as one.
+# The two answers are one answer at the only site that asks: a text this cannot
+# read and a text reading as `NULL` both leave the condition no call to name.
 parse_call_text <- function(text) {
   tryCatch(str2lang(text), error = function(cnd) NULL)
 }
 
-# `expr` with every part `deparse()` cannot write as source replaced by a name
-# spelling the class of what stood there. Only the parts that fail are touched,
-# which is why the test is a trial rather than a type: a `data.frame` and a
-# Grouping specification both round-trip, and replacing them would thin a call
-# that reads correctly today.
+# `expr` -- a call -- with every part `deparse()` cannot write as source
+# replaced by a name spelling the class of what stood there. Only the parts
+# that fail are touched, which is why the test is a trial rather than a type: a
+# `data.frame` and a Grouping specification both round-trip, and replacing them
+# would thin a call that reads correctly today.
 #
 # A symbol is answered before the trial. `deparse()` writes a non-syntactic
 # name bare when it is asked for one alone -- `unit count`, which does not
 # parse -- and backquotes it inside the call it is written in, so trying one on
 # its own reports a failure the text this produces does not have.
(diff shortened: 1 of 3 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel, against the merge-base with `main` (`git diff origin/main...HEAD`, merge-base `24958b0`), reading `e7a3001`. Standards sources: `AGENTS.md` (*Code comments*, *Linting*, *Release matrix*, *Queries against a lazy input*, *Optional-dependency guards*, *Dependency metadata*), `design/architecture.md` *Answering a review*, ADR 0015, ADR 0023, plus the skill's Fowler smell baseline. Spec source: issue #360 and its agent brief.

The local `main` ref was six branches stale; reviewing against it would have reported #346, #349, #350, #351, #354, #357 and #362 as this branch's work. The fixed point is `origin/main`.

### Standards — 5 findings

| Finding | Disposition |
|---|---|
| `parse_call_text()` uses `NULL` as its failure sentinel, but `str2lang("NULL")` *is* `NULL`, so every `NULL` argument was replaced with `` `<NULL>` `` once any sibling failed | fixed in `46e5c2a` |
| The header's "only the parts that fail are touched" was false as a consequence | the code was what was wrong; carried by the fix above |
| `deparse_call_text()` and `deparse_round_trips()` carried no header | fixed in `46e5c2a` |
| "rather than something a call rewrite avoids" re-derives ADR 0015's predicate beside the citation | deleted in `46e5c2a` |
| `round_trippable_call()` reads as a predicate but returns a call, beside `deparse_round_trips()` which is one | renamed `writable_call()`; `unrepresentable_name()` → `unwritable_name()` to share the vocabulary |

One-backend-per-test, the lazy-input query policy, and dependency metadata raised nothing.

### Spec — 2 findings

| Finding | Disposition |
|---|---|
| Same `NULL` defect, against "everything else the caller wrote left intact" | fixed in `46e5c2a` |
| The `across()` end-to-end test asserted only that *a* call was reported, not the substitution or the source summary the criterion names | both asserted in `46e5c2a` |

No scope creep. Criteria 1–5 and the whole *Out of scope* list verified as honoured, including that a dtplyr share whose call round-trips produces byte-identical text.

### What the fix exposed

Correcting the predicate turned up a second defect neither axis could see, because it was masked by the first: `expr[[position]] <- NULL` *deletes* an argument rather than setting it, so a `NULL` the walk correctly returned unchanged then dropped out of the call. `test-share.R`'s "a call text keeps a `NULL` the caller wrote" fails without both halves.

### Checks after the answers

- `testthat::test_dir("tests/testthat")` under `NOT_CRAN=true`: 6171 pass, 0 fail, 0 skip
- `lintr::lint_package()` after `pkgload::load_all()`: no lints

PR #368, opened 2026-09-01
Refuse a name on a grouping constructor's argument (#365)

### LINKED ISSUE #365: A verb argument written inside rollup() becomes a grouping dimension
Found by using the package on `ggplot2::diamonds` rather than by reading the code.

`rollup()`, `cube()`, `grouping_sets()` and `grouping_spec()` take only `...`, and pass it to tidyselect. The verb arguments an analyst is holding in their head at the same moment — `.by`, `.id`, `.sort` — are arguments of `summarize_with_margins()`, one line below. Writing one in the wrong pair of parentheses is not caught:

```r
# intended: color is a fixed partition, cut is rolled up
ggplot2::diamonds |>
  summarize_with_margins(n = n(), .grouping = rollup(cut, .by = color))
#> # A tibble: 3 × 3
#>   cut     color     n
#>   <ord>   <ord> <int>
#> 1 Ideal   E      3903
#> 2 Premium E      2337
#> 3 Good    E       933
```

No error. `color` became a second rollup dimension, so the report has `(cut, color)`, `(cut)` and `()` levels where the caller asked for a per-`color` report over `rollup(cut)`. The numbers are all correct for the grouping that ran, and nothing in the result says which grouping that was.

The other two fail, but not in a way that names the mistake:

```r
rollup(cut, .sort = "last")      #> Can't select columns that don't exist. ✖ Column `last` doesn't exist.
cube(cut, color, .id = "level")  #> Can't select columns that don't exist. ✖ Column `level` doesn't exist.
```

A reader takes that as a claim about their data.

A dot-prefixed name in `...` is never a column a caller means to select or rename here, so the helpers can reject one and say where the argument belongs. Whether that is a general rule for `^\.` or a list of the verbs' own argument names is the design question.

sayuks:
Closes #365.

The grouping constructors capture `...` as quosures and never read the names R bound to those arguments. A name inside a selection is refused — `rollup(c(area = cut))` reports a renaming dimension — while the same name one level out was dropped: `rollup(area = cut)` resolved to `rollup(cut)`, and `rollup(cut, .by = color)` took `color` as a second dimension, producing a report correct for the grouping that ran with nothing saying which grouping that was.

`preflight_grouping_spec()` now refuses a named argument over the whole call before the loop reads any argument, with its own diagnostic. The design was settled on the issue; the agent brief there carries the decisions and what each rests on.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,7 +27,11 @@
   constructor and the position — `grouping_sets(, region)` reports that its
   first argument is empty — rather than reporting an internal name you never
   wrote. A trailing comma is not an empty argument, because R captures no
-  argument for it.
+  argument for it. A name on the constructor's own argument is refused too:
+  `rollup(area = region)` reports the name rather than dropping it, and so a
+  Margin verb's own argument written one pair of parentheses in —
+  `rollup(region, .by = year)` — is reported rather than taken as a second
+  dimension (#365).
 * Added contextual `grouping_bit()` and `grouping_id()` summary helpers.
 * Added the contextual `share_of_parent()` summary helper, which divides a
   preceding numeric scalar summary by the same measure one `rollup()` level
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -272,6 +272,26 @@ abort_empty_grouping_arg <- function(constructor, position) {
   ))
 }
 
+# The refusal of a name R bound to a constructor's own argument. Every
+# constructor takes `...` alone, so such a name is not a selection renaming a
+# dimension -- `abort_grouping_rename()` answers that one level in -- and
+# nothing reads it: expansion takes the argument's value and the name reaches
+# no position. A Margin verb's own argument written one pair of parentheses in
+# arrives here, which is the spelling that made the position silent (#365).
+#
+# The pairs arrive alone in an `i` bullet, per ADR 0023's condition 2: how many
+# of them there are is the caller's decision.
+abort_named_grouping_arg <- function(constructor, pairs) {
+  abort_marginplyr(c(
+    "{.fun {constructor}} takes no named arguments:",
+    i = "{.code {pairs}}.",
+    i = paste0(
+      "Remove the {cli::qty(length(pairs))}name{?s}, or move a Margin verb's ",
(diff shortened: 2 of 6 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` over `git diff origin/main...HEAD` at `ff8d0fe`, two axes in parallel. The local `main` ref was stale, so the fixed point is `origin/main` (`d7cbd37`); against local `main` the diff would have carried 24 files of other branches' work.

Five findings. Answered at `749e0c6`.

|---|---|---|---|
| 1 | Standards, Spec | The scan's header claimed an empty argument gives "the name and nothing after the `=`". `rlang::as_label()` answers one with `<empty>`, so `rollup(area = )` reported a token the caller never wrote, and the named-and-empty case silently took precedence over `abort_empty_grouping_arg()` with nothing pinning it. | Fixed in code |
| 2 | Standards | The same sentence cited ADR 0024 for a right half that is a deparse in every case, which ADR 0024's closing paragraph puts outside its rule. | Fixed in prose |
| 3 | Standards | The header restated `abort_empty_grouping_arg()`'s premise from eleven lines above. | Fixed in prose |
| 4 | Standards | Byte-identical copy of the ADR 0023 condition 2 comment carried by `abort_grouping_rename()`. | Fixed in prose |
| 5 | Standards | `check_named_grouping_args()`'s pair formatting has `selection_rename_pairs()`'s shape. | Rejected |

Finding 1 was answered by leaving an empty argument to the refusal that points by position (#261) rather than by correcting the sentence: an empty argument has no spelling for a pair to quote, so the scan cannot speak for one. The false sentence is deleted rather than restated. The test that fails without it is `test-grouping-plan.R`'s *a name on an empty argument is refused by position*.

Finding 5 is rejected on `selection_rename_pairs()`'s own header: its contract is the renaming a selection asked for, in the spelling the caller would write to fix it. These pairs are neither a renaming nor a spelling that fixes one, so one helper could not state one contract.

Both axes also reported the diff clean against *Release matrix* (the new tests need no backend), *Optional-dependency guards*, and *Linting*, and the Spec axis verified all eight acceptance criteria by running them.

PR #369, opened 2026-09-01
Read the plan's dimensions for a bare grouping_id() (#366)

### LINKED ISSUE #366: grouping_id() with no arguments is rejected although the call knows the grouping
Found by using the package on `ggplot2::diamonds`.

The natural way to keep a level identifier in the result is to ask for one:

```r
ggplot2::diamonds |>
  summarize_with_margins(gid = grouping_id(), .grouping = rollup(cut, color))
#> Error: `grouping_id()` requires at least one column.
```

What works is retyping the specification:

```r
summarize_with_margins(gid = grouping_id(cut, color), .grouping = rollup(cut, color))
```

The columns are already written in `.grouping`, in the same call, and the mask `grouping_id()` returns is only meaningful against that plan's column order — `inspect_grouping()` prints the same order. So the second spelling is a copy of the first that a later edit to `.grouping` can leave stale: changing `rollup(cut, color)` to `rollup(cut, color, clarity)` renumbers the levels while `grouping_id(cut, color)` keeps returning a two-bit mask, silently.

`grouping_id()` over every `.grouping` column, in plan order, is the reading with no other candidate. SQL's `GROUPING_ID()` requires its arguments because the statement has no other place to look; here `.grouping` is that place.

`grouping_bit()` is different and should stay as it is — it asks about one named column, so no-argument has no meaning for it.

sayuks:
Closes #366.

`grouping_id()` written with no columns now reads every Grouping dimension of the resolved plan, in plan order — the order `inspect_grouping()` reports — so a bare call equals `inspect_grouping()$grouping_id` for the row's Grouping set. Fixed `.by` columns are excluded; a plan with no dimensions gives `0L`; a plan past the 31-column cap keeps the existing refusal. `grouping_bit()` is unchanged.

ADR 0009 carries the amendment. Review record to follow.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -33,6 +33,18 @@
   `rollup(region, .by = year)` — is reported rather than taken as a second
   dimension (#365).
 * Added contextual `grouping_bit()` and `grouping_id()` summary helpers.
+  `grouping_id()` written with no columns reads every `.grouping` column of the
+  resolved plan, in plan order, which is the order `inspect_grouping()` reports:
+  a bare `grouping_id()` is `inspect_grouping()$grouping_id` for the Grouping
+  set the row came from, as `.id` is `set_id` for its occurrence. Retyping the
+  columns is still accepted, and is how a subset of them or an order other than
+  the plan's is encoded, but a retyped list is a copy of a specification written
+  in the same call: widening `rollup(cut, color)` to
+  `rollup(cut, color, clarity)` renumbered the levels while
+  `grouping_id(cut, color)` went on returning a two-bit mask (#366). Columns
+  fixed by `.by` are not part of the default, each contributing a bit that is
+  always zero. A plan with no dimensions gives `0L`, and one past the 31-column
+  cap is refused as a written call of that width is.
 * Added the contextual `share_of_parent()` summary helper, which divides a
   preceding numeric scalar summary by the same measure one `rollup()` level
   up, partitioned by the fixed `.by` keys. Local data frames, dbplyr, and
--- a/R/grouping-context.R
+++ b/R/grouping-context.R
@@ -13,8 +13,7 @@
 #' the resolved plan, in plan order. That is the order [inspect_grouping()]
 #' reports, so a bare [grouping_id()] equals that function's `grouping_id` for
 #' the grouping set the row came from. Columns fixed by `.by` are not part of
-#' that default: each belongs to every grouping set, so it would widen the
-#' mask by a bit that is always zero.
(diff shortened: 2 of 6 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel against `origin/main...HEAD`
at af4c7df. Standards sources: `AGENTS.md` (*Linting*, *Documentation*, *Code
comments*, *Release matrix*, *Optional-dependency guards*, *Queries against a
lazy input*, *Dependency metadata*), `design/architecture.md`, ADRs 0009, 0015,
0023, `CONTEXT.md`. Spec source: #366 and its agent brief.

### Standards — 3 findings, 3 baseline smells

| Finding | Disposition |
|---|---|
| The default's comment re-derives ADR 0009's own argument beside the citation, against *Code comments* (hard) | fixed in 21fda00 — collapsed to which way the site falls under the ADR |
| The same zero-bit argument again in two test comments | fixed in 21fda00 — deleted |
| The new roxygen sentence restates a clause the same block already carries | fixed in 21fda00 — sentence stops earlier |
| Mysterious Name: a test named for a bare `grouping_id()` that calls none | fixed in 21fda00 |
| Mysterious Name: `one_set()` names neither what it builds nor its shape | fixed in 21fda00 — `single_grouping_set()` |
| Duplicated Code: the correspondence test writes each `.grouping` twice | fixed in 21fda00 — one specification is injected into both calls, which is the condition ADR 0009 states the correspondence under |

No finding under ADR 0015, ADR 0023, *Release matrix*, *Queries against a lazy
input*, or *Dependency metadata*. The removed `requires at least one column`
diagnostic has no stale reference in `R/`, `vignettes/`, `man/`,
`.github/scripts/`, or the snapshots.

### Spec — 1 finding

Every acceptance criterion verified by running, not by reading. The reviewer
additionally checked `.duplicates = "keep"`, a composite dimension, a nested
specification, and the emulated `UNION ALL` path on SQLite, none of which the
brief named and all of which hold.

| Finding | Disposition |
|---|---|
| The `@examples` block's `level = grouping_id(region, store)` became `level = grouping_id()`, an edit the brief did not ask for | rejected — a reference example is where a new default is shown, the first example keeps the written spelling, and the two produce the same values |

### Raised on both axes

`NEWS.md` gained no entry, though this adds a public spelling and every
preceding user-visible change on this line recorded one. Fixed in 21fda00.

Both reviewers found nothing implemented wrongly and nothing on the brief's
*Out of scope* list touched.

PR #370, opened 2026-09-01
Name .sort where the README's first output is read (#367)

### LINKED ISSUE #367: README's first output shows a row order the default does not produce
Found by copying the README's first example onto a real dataset.

The headline block prints detail rows grouped tidily under their region, then the subtotals, then the grand total. Within a reporting level that order is the source's order of first appearance, not a sort. Shuffling the fixture's rows — same data, same call — reorders the report:

```r
js <- retail_sales |> dplyr::filter(year == 2026L, month == "Jan")
set.seed(1)
js[sample(nrow(js)), ] |>
  summarize_with_margins(revenue = sum(revenue), .grouping = rollup(region, store))
#>   region         store revenue
#> 1   East        Boston    6000
#> 2   West San Francisco    7200
#> 3   East          <NA>    2200
#> 4   West       Seattle    4900
#> 5   East      New York    3000
#> 6   East         Total   11200
#> 7   West         Total   12100
#> 8  Total         Total   23300
```

On `ggplot2::diamonds` the same call gives `Ideal/E`, `Premium/E`, `Good/E`, `Premium/I`, … — the first thing a reader sees is a scrambled report where the README promised a finished one.

`.sort` is the answer and it is already there: `.sort = "last"` produces what the README shows, for any input. The gap is that `.sort` appears nowhere in `README.Rmd`, which never mentions row order at all, so a reader has no reason to look for it. `README.md` is the page with the widest reach and the one a new user meets first.

Smallest fix is a sentence after the first output naming `.sort` and linking the vignette that covers it. Whether the headline example itself should pass `.sort = "last"` — showing the argument that makes the order real, at the cost of a longer first call — is the question worth deciding.

sayuks:
Closes #367.

The README's headline block prints detail rows, then subtotals, then the grand
total, and says nothing about how they got there. Within a reporting level that
arrangement is the input's own row order, so the same call on the same data
shuffled reads differently.

`.sort` is what asks for an order, and it appeared nowhere in either half of the
README. The reference, Get Started, and the recipes vignette all state the rule;
the page with the widest reach was the one that left a reader no reason to look
for it.

The added sentence states what those pages state and cites the recipes section
that shows it. The headline call keeps its arguments: `.sort = "last"` does not
produce the block the README shows — it puts each subtotal directly beneath the
rows it summarizes — so passing it would change the first output a reader sees
in exchange for a longer first call. That was the decision the issue named as
the one worth making.

`README.md` is regenerated from `README.Rmd` against the installed working tree
and pandoc 3.10.1, the version `document.yaml` pins. Its diff is the prose and
nothing else; every output block is byte-identical.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -54,11 +54,10 @@ january_sales |>
 The result is one regular data frame—not three reports that you have to build
 and combine yourself.
 
-Row order is unspecified. The arrangement above happens to follow the input's
-own row order, so the same call on the same data shuffled reads differently.
-`.sort` asks for a Margin order, which puts each subtotal with the rows it
-summarizes, and `arrange()` gives any other presentation order. [Put each
-subtotal with the rows it
+Row order is unspecified: the same call on the same data in a different order
+can arrange the result differently. `.sort` asks for a Margin order, which puts
+each subtotal with the rows it summarizes, and `arrange()` gives any other
+presentation order. [Put each subtotal with the rows it
 summarizes](https://sayuks.github.io/marginplyr/vignettes/recipes.html) shows
 `.sort` at work and what it does not promise.
 
--- a/README.Rmd
+++ b/README.Rmd
@@ -54,6 +54,14 @@ january_sales |>
 The result is one regular data frame—not three reports that you have to build
 and combine yourself.
 
+Row order is unspecified. The arrangement above happens to follow the input's
+own row order, so the same call on the same data shuffled reads differently.
+`.sort` asks for a Margin order, which puts each subtotal with the rows it
+summarizes, and `arrange()` gives any other presentation order. [Put each
+subtotal with the rows it
+summarizes](https://sayuks.github.io/marginplyr/vignettes/recipes.html) shows
+`.sort` at work and what it does not promise.
+
 - Works with local data frames and lazy `dbplyr` tables.
 
 - Uses native `GROUP BY GROUPING SETS` on DuckDB and PostgreSQL.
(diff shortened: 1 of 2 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel, against the merge-base with `main` (`ea39026...HEAD`), at `05ee474`. `README.md` is generated and `document.yaml` checks it against its source, so it was outside the review; `README.Rmd` was the subject.

**Standards — no findings.** One judgement call was raised and not reported as a finding: the sentence described behaviour the package does not promise, in the document with the widest reach, with nothing failing when it stopped holding. The Spec axis found the same sentence false, so it is answered below.

**Spec — two findings.**

| finding | disposition |
|---|---|
| *"The arrangement above happens to follow the input's own row order"* is false. Only the detail rows do: shuffling the fixture reorders those five and leaves the two subtotals and the grand total in positions 6 to 8. | Fixed in prose (`2c2a16c`). The claim is deleted rather than narrowed, per *Answering a review*. Naming the mechanism was not what the fix required, and the deletion also closes the Standards axis's judgement call. |
| The link resolves to the recipes vignette rather than to its *Put each subtotal with the rows it summarizes* section, while the link text names the section. | Rejected. `vignettes/get_started.qmd` links this same section by the same construction — section title as link text, page URL as target — and no README vignette link carries a fragment. The brief's other half, *reached the way the README's other vignette links are*, is what the repository actually holds; nothing here validates a fragment, so one would be an unchecked claim of exactly the kind the first finding was. |

PR #376, opened 2026-09-01
Spell a per-dimension NULL Margin label (#371)

### LINKED ISSUE #371: `.margin_label` cannot express a per-dimension `NULL`, and both halves of the collision check depend on one

*Rewritten during triage: the original report framed this as a missing hint on the observed half of the collision check. Triage established a different root, and the two symptoms it describes both hang off it. The original text is in this issue's edit history.*

## The defect

`.margin_label` chooses a label per dimension, so a named vector is the natural spelling. `NULL` — the value the reference documents as producing a typed-missing Margin value while opting out of collision checks — is the one choice that cannot be made per dimension. A named character vector cannot carry a `NULL` element:

```r
.margin_label = c(region = "All", store = NULL)
#> Error:
#> `.margin_label` must name every Margin dimension.
#> ℹ Missing `store`.
```

`c()` drops the `NULL` before the argument is ever seen, and a named list is rejected by the guard. So `NULL` is reachable only as an unnamed scalar applying to every dimension at once, and *both halves of the Margin label collision check rest on a per-dimension `NULL` that cannot be written.*

## Symptom 1 — the observed half refuses a result the package permits by another spelling

Found while building a two-level management report where `region` should print a custom label and `store` should stay typed-missing, because `store` carries source `NA` and `grouping_bit()` is what tells the two apart:

```r
retail_sales |>
  summarize_with_margins(
    revenue = sum(revenue),
    .grouping = rollup(region, store),
    .margin_label = c(region = "All", store = NA_character_)
  )
#> Error:
#> NA is already present in grouping column:
#> ℹ `store`.
#> ℹ Choose another `.margin_label` or set `.check_margin_label = FALSE`.
```

Neither remedy is available. There is no other `NA` to choose, and `NULL` is all-or-nothing. That leaves `.check_margin_label = FALSE`, which is a property of the call rather than of the dimension, so it also turns off the observed check on `region`:

```r
collides <- retail_sales |>
  dplyr::mutate(region = ifelse(store == "Boston" & !is.na(store), "All", region))

collides |>
  summarize_with_margins(
    revenue = sum(revenue),
    region_is_margin = grouping_bit(region),
    .grouping = rollup(region, store),
    .margin_label = c(region = "All", store = NA_character_),
    .check_margin_label = FALSE,
    .sort = "last"
  ) |>
  dplyr::filter(region == "All")
#>   region  store revenue region_is_margin
#> 1    All Boston   22800                0
#> 2    All   <NA>   22800                0
#> 3    All   <NA>   87400                1
```

Rows 2 and 3 print identically — a source region named `All` with its store subtotal, and the Grand total set. That is the collision `.check_margin_label` exists to report, and the report is silent, because reaching a typed-missing label on a *different* dimension required switching the check off.

**What the refusal is actually refusing.** `NULL` and `NA_character_` produce identical output on a column holding source `NA`:

```r
d <- data.frame(region = c("E","E","W"), store = c("a", NA, "b"), v = 1:3)
# .margin_label = NULL, and .margin_label = NA_character_ with the check off:
#   region store n
# 2      E  <NA> 2   <- source NA
# 3      E  <NA> 3   <- margin
```

ADR 0012's own eight-case table carries the asymmetry. Row 3 (`NA_character_`, no NA level, missing value present) reads *Error*; row 7 (`NULL`, same column) reads *Allowed*. Same premise, opposite verdict, identical result. So the refusal rejects one of two spellings of a result the package already permits — and it reaches the caller who picked the spelling the documentation calls the checked one, not the caller whose result is ambiguous.

## Symptom 2 — the declared half's remedy names a spelling that cannot be written

The original report held the declared half up as the branch that recognises this shape and redirects instead of refusing. It redirects to a dead end:

```r
status <- structure(
  c(1L, 2L, NA_integer_), levels = c("kept", NA_character_), class = "factor"
)
d <- data.frame(region = c("E","E","W"), status = status, v = 1:3)

d |>
  summarize_with_margins(
    n = sum(v), .grouping = rollup(region, status),
    .margin_label = c(region = "All", status = NA_character_)
  )
#> Error:
#> `NA_character_` is already a factor level in grouping column:
#> ℹ `status`.
#> ℹ Use `NULL` for a typed-missing Margin label while preserving the NA level.

# following that hint:
d |>
  summarize_with_margins(
    n = sum(v), .grouping = rollup(region, status),
    .margin_label = c(region = "All", status = NULL)
  )
#> Error:
#> `.margin_label` must name every Margin dimension.
#> ℹ Missing `status`.
```

The remedy works only where there is exactly one dimension. `R/margin-label.R` states the rule this breaks, beside the same pair of refusals: `.check_margin_label = FALSE` is offered only where it is a remedy, because naming an argument that changes nothing sends the caller nowhere.

## Decisions taken in triage

Recorded as two amendments to ADR 0012 and a narrowing of *Margin label collision* in `CONTEXT.md`:

1. **`.margin_label` also accepts a named list** whose elements are character scalars or `NULL`. The named character vector stays accepted as the shorthand for a list with no `NULL` element. The requirement to name every Margin dimension is unchanged — an unnamed dimension stays a refusal, since the alternative makes a misspelled name turn every dimension typed-missing with no diagnostic.
2. **A typed-missing Margin label is not an observed collision.** `NA_character_` is excluded from the observed half before it selects a column, so ADR 0012's third row becomes *Allowed*, matching its seventh, and a call whose every label is typed-missing sends no query — as `NULL` already did. Nothing is reported in its place: a message on one spelling and silence on the other would move the asymmetry into the diagnostics rather than remove it.
3. **The declared half is unchanged.** An NA factor level is part of the column's domain and `levels()` records it, so `NA_character_` proposed against one stays a structural refusal (rows 1 and 2 still read *Error*). That is the whole of what separates the two spellings once 2 lands, and it is the difference that survives to something a caller can observe — which is why it is kept.
4. **`.check_margin_label` stays a logical scalar.** Making it per dimension was considered and is unnecessary once `NULL` is per-dimension spellable and an `NA` label no longer reaches the observed check.
5. **The eight-case table stays eight rows.** It exists to be indexed by the value the reader wrote and the column they have; collapsing it saves four lines across two copies and costs every reader a derivation.

sayuks:
Closes #371.

`.margin_label` chooses a label per dimension, but `NULL` — the value that selects a typed-missing Margin value and opts out of collision checking — was the one choice that could not be made per dimension. `c()` drops a `NULL` element before the argument is reached, and a named list was refused, so `NULL` applied to every dimension or to none. Both halves of the Margin label collision check rested on a per-dimension `NULL` that could not be written: the observed half refused `NA_character_` against a source missing value and offered only a per-call escape, and the declared half's refusal named the unwritable `NULL` as its remedy.

## What changed

**`.margin_label` accepts a named list** whose elements are character scalars or `NULL`. A named character vector is the shorthand for a list holding no `NULL` and is normalized to one, so nothing downstream carries two spellings. The name rules are unchanged, including the requirement to name every Margin dimension — reading an unnamed dimension as `NULL` would turn a misspelled name into a silent whole-call change.

**A typed-missing label is no longer an observed collision.** `NULL` and `NA_character_` produce the same displayed result on a column already holding missing values, and ADR 0012's table permitted that result under one spelling (row 7) while refusing it under the other (row 3). The refusal reached the caller who picked the spelling the documentation calls the checked one, not the caller whose result was ambiguous. The observed check now selects no column for either spelling, so a call whose every label is typed-missing sends no query — as `NULL` already did (ADR 0020). Nothing is reported in its place.

**The declared half is unchanged.** An NA factor level is recorded in `levels()`, so `NA_character_` proposed against one is still refused on every backend whatever `.check_margin_label` says. Its remedy names the list spelling, because the bare `NULL` it named before is the whole of `.margin_label` and so was a remedy only where there was one dimension — the rule the same file states for `.check_margin_label = FALSE`, run backwards.

`.check_margin_label` remains a logical scalar. The eight-case table keeps eight rows.

## What it looks like

```r
retail_sales |>
  summarize_with_margins(
    revenue = sum(revenue),
    store_is_total = grouping_bit(store),
    .grouping = rollup(region, store),
    .margin_label = list(region = "All", store = NULL)
  )
```

The check stays on for `region`, `store` stays typed missing, and `grouping_bit()` tells a source `NA` from the margin over it.

## Design record

Two amendments to ADR 0012 and a narrowing of *Margin label collision* in `CONTEXT.md`, committed separately as `224da50`. The decisions were settled in a grilling session on #371 and are listed in that issue's body.

## Checks

- `lintr::lint_package()` — no lints.
- `NOT_CRAN=true testthat::test_dir("tests/testthat")` — 6260 pass, 0 fail, 0 skip.
- `roxygen2::roxygenise()` — four Rd pages regenerated and committed; no `# nolint` leaked into `man/`.
- `verify-context-budget.R` — 24198 of 24198 bytes, unchanged.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -73,12 +73,14 @@ factor level.
 _Avoid_: Total value, missing-value replacement
 
 **Margin label collision**:
-A grouping dimension holding, as one of its own values, the Margin label
-chosen for it — so that a margin row and a row of the source data cannot be
-told apart by that column. A collision is *declared* when the colliding value
-is a factor level, which the column's type records, and *observed* when it
-appears only among the column's values.
-_Avoid_: Duplicate label, ambiguous margin
+A grouping dimension holding, as one of its own values, the non-missing Margin
+label chosen for it — so that a margin row and a row of the source data cannot
+be told apart by that column. A collision is *declared* when the colliding
+value is a factor level, which the column's type records, and *observed* when
+it appears only among the column's values. A typed-missing Margin label is not
+a collision: it displays as missing wherever the column already holds missing
+values, and a Grouping bit or Grouping identifier is what tells the two apart.
+_Avoid_: Duplicate label, ambiguous margin, typed-missing ambiguity
 
 **Margin order**:
 The row order a Margin operation produces when it is asked to order its
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -80,7 +80,7 @@ value is a factor level, which the column's type records, and *observed* when
 it appears only among the column's values. A typed-missing Margin label is not
 a collision: it displays as missing wherever the column already holds missing
 values, and a Grouping bit or Grouping identifier is what tells the two apart.
-_Avoid_: Duplicate label, ambiguous margin, typed-missing ambiguity
+_Avoid_: Duplicate label, ambiguous margin
 
 **Margin order**:
 The row order a Margin operation produces when it is asked to order its
--- a/R/margin-label.R
+++ b/R/margin-label.R
@@ -1,18 +1,16 @@
-# The Margin label as one shape: `NULL`, an unnamed character scalar, or a
-# named list holding one label per dimension. A named character vector is the
-# shorthand for a list with no `NULL` element and is converted to one here, so
-# nothing downstream carries two spellings of the same argument.
-#
-# The list exists because `c()` drops a `NULL` element before this is reached,
-# which left a per-dimension `NULL` unwritable while both halves of the
-# collision check rested on one (ADR 0012).
+# The Margin label a caller wrote, as the shape the rest of the operation
+# reads: `NULL`, an unnamed character scalar, or a named list of per-dimension
+# labels. A named character vector is converted to that list here, so the two
+# named spellings are one downstream (ADR 0012).
 normalize_margin_label <- function(.margin_label) {
   if (is.null(.margin_label)) {
     return(NULL)
   }
   # An unnamed list is not the list form: its elements name no dimension, and
-  # a caller who meant one label wrote the scalar.
-  is_label_list <- is.list(.margin_label) && !is.null(names(.margin_label))
+  # a caller who meant one label wrote the scalar. Bare, because a one-row
+  # data frame is otherwise a named list whose columns pass as labels.
+  is_label_list <- rlang::is_bare_list(.margin_label) &&
+    !is.null(names(.margin_label))
   if (!is_label_list &&
         (!is.character(.margin_label) || length(.margin_label) == 0L)) {
     abort_marginplyr(paste0(
@@ -48,8 +46,6 @@ normalize_margin_label <- function(.margin_label) {
 
 # Each element of a named-list `.margin_label` is one dimension's label, so it
 # is what an unnamed scalar may be, plus the `NULL` the list exists to carry.
-# The offending names travel in an `i` bullet because how many of them arrive
-# is the caller's decision (ADR 0023).
 validate_margin_label_elements <- function(.margin_label) {
   bad <- vapply(
     .margin_label,
@@ -171,8 +167,8 @@ validate_margin_label <- function(.data,
           "{cli::qty(length(na_level_cols))}column{?s}:"
         ),
         i = "{.var {na_level_cols}}.",
-        # A bare `NULL` is the whole of `.margin_label`, so it is a remedy
-        # only where there is one dimension; the list is one either way.
+        # The list spelling, because a bare `NULL` is the whole of
+        # `.margin_label` and so is not one dimension's answer (ADR 0012).
         i = paste0(
           "Use {.code NULL} in a named-list {.arg .margin_label} for a ",
           "typed-missing Margin label while preserving the NA level."
@@ -238,9 +234,7 @@ check_observed_label_collision <- function(data,
   # A factor dimension states its values in its levels, and a label equal to
   # one of them was rejected above, so reading its column could only find a
   # value the levels do not contain. A typed-missing label is not a collision
-  # (ADR 0012): it displays as missing wherever the column already holds
-  # missing values, which is the result `NULL` has always been allowed to
-  # produce, so neither spelling of it selects a column.
+  # (ADR 0012), so neither spelling of one selects a column either.
   read_cols <- Filter(
     function(col) {
       !is_missing_margin_label(margin_labels[[col]]) && !(col %in% factor_cols)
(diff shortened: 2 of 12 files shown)


sayuks:
> *This was generated by AI during review.*

## Review record

One round through `mattpocock-skills:code-review`, both axes, against `git diff origin/main...HEAD` at `1738a32`. Answered in `f6053ea`.

### Standards — 4 hard, 4 judgement calls

| Finding | Disposition |
|---|---|
| The eight-case table's heading conditions it on `.check_margin_label = TRUE` while the paragraph below says every row is decided whatever the argument says | fixed in prose — the condition is what became false, so it is deleted |
| `normalize_margin_label()`'s header re-derives ADR 0012's amendment beside its citation, in past tense | fixed in prose — the header states the shape it returns and cites |
| The observed check's comment re-derives the same argument, a third copy after `CONTEXT.md` and the ADR | fixed in prose |
| The NA-level remedy's comment reproduces the amendment's closing paragraph, uncited, while that paragraph points back at this file | fixed in prose — it states which way this site falls and stops |
| `validate_margin_label_elements()`'s header repeats the ADR 0023 `i`-bullet rule the same file states 25 lines below | fixed in prose |
| `is.list(x) && !is.null(names(x))` admits a data frame as the list form | fixed in code — `rlang::is_bare_list()`, with the case added to the shape refusal's table in `test-margin-label.R`. A one-row data frame was not merely mis-diagnosed but **accepted**: its columns are character scalars |
| The header's "nothing downstream carries two spellings" overclaims — the unnamed scalar stays a character vector | fixed in prose, in the same rewrite |
| The table's *Missing value* column decides nothing now that row 3 reads Allowed | **rejected** — it separates the results in three of its four pairs (3/4, 5/6, 7/8); only rows 1 and 2 agree |

### Spec — 0 findings

Every acceptance criterion in the issue's agent brief was verified by running R, and every *Out of scope* item held: `.check_margin_label` is still an `assert_logical_scalar()` per call, backend defaults are untouched, both collision refusals are unchanged in the diff, nothing replaces the removed refusal, and `.margin_label_position` is untouched. Rows 1, 2, 5, and 6 of the table are byte-identical.

Two findings were raised and rejected:

- That the `.check_margin_label` parameter should show `list(<dimension> = NULL)` as code rather than name the spelling. How prose reads is not a finding (`design/architecture.md`, *Answering a review*), and the examples show the literal.
- That `CONTEXT.md` adding `typed-missing ambiguity` to the term's *Avoid* list was unasked-for scope. Rejected as scope, **fixed for a different reason**: the entry's body already states that a typed-missing label is not a collision, so the list was a second copy of that sentence inside one entry.

### Checks at `f6053ea`

- `lintr::lint_package()` — no lints.
- `NOT_CRAN=true testthat::test_dir("tests/testthat")` — 6262 pass, 0 fail, 0 skip.
- `roxygen2::roxygenise()` — regenerated and committed.
- `verify-context-budget.R` — 24198 of 24198 bytes, unchanged.

Issue #386, opened 2026-09-02
Decide whether the record needs a SQL-detection predicate


sayuks:
Part of #377

## Question

Does the record need a predicate saying whether the object in hand is a SQL
query, and if so is `is_sql_lazy_query()` extracted out of what
`grouping_backend()` computes inline at `R/grouping-backend.R:14`?

The map's fog listed this as "a judgement that rides on whichever features
survive". They have all now been decided — #381 rejected the condition, #382
rejected `"explain"`, #383 fixed `last_sent_queries()` and the reset in
`prepare_grouping_plan()`, #384 rejected the on-disk log — so the question has a
subject and can be put.

What has to be decided:

1. **Whether the record needs the predicate at all.** #318's warrant for it was
   "self-detection from the query object means no call site needs to thread
   backend/operation through just for this" — an argument about the
   `signal_marginplyr_query_sql()` helper, which #381 removed. #383 put the
   reset in `prepare_grouping_plan()`, where `grouping_backend()` has already
   run, so the sites may be *handed* the kind rather than recomputing it. ADR
   0020's predicate for whether a query is sent is `is.data.frame()`, not a SQL
   test, and #383 recorded that `grouping_selection_proxy()` sends nothing on
   arrow and nothing on a backend without `collect_selection_proxy` — neither of
   which a `tbl_lazy` test answers.
2. **If it is needed, whether extraction is the shape.** #378 answered the
   mechanics and they cut against it: `is_dtplyr` and `is_arrow` are locals
   `grouping_backend()` computes for its own `kind` cascade whether or not
   `is_sql` moves, so extracting buys no arithmetic; and the helper must call
   `arrow_input_classes()` rather than spell the vector, since that function's
   header says it exists *because* a second copy of the vector is what would
   drift (`R/grouping-backend.R:1-8`).
3. **Whether it stands on its own merits regardless.** #318 concedes
   `grouping_backend()` needs no helper for its own sake. If the record does not
   need one either, #318's section is deleted rather than kept, and that is the
   answer.

The map's *Out of scope* rules out writing the R code. What this decides is
whether the specification #385 assembles contains the helper at all.

## Resolution

A verdict, and — if the helper survives — the contract its header would state,
so #385 can write it into #318 rather than re-deciding it.

sayuks:
Resolved. **The record needs no SQL-detection predicate.** `is_sql_lazy_query()` is rejected, and #318's section for it is deleted rather than kept.

[...]

Issue #384, opened 2026-09-02
Decide the on-disk audit log: whether, what format, and whose failure


sayuks:
Part of #377

## Question

Does marginplyr write an on-disk log, and if so, in what format and under whose
failure policy?

#318 proposes `options(marginplyr.audit_sql_file = "path")`, unset by default,
appending every captured entry to that path when `marginplyr.audit_sql` is also
`TRUE`. Its stated warrant is that a corporate audit needs a record surviving
the session, which in-memory retrieval alone cannot give.

What has to be decided:

1. **Whether at all.** #381 rejected the condition, which removes the
   comparator this question was originally framed against — there is no handler
   for a caller to write. It also *raises the stakes here*: #381 handed the
   multi-call batch case to this ticket, on the reasoning that if the file log
   is rejected too, a caller loops and appends the per-call record themselves.
   So the question is whether a package-supplied sink beats that loop, and is a
   job a data-manipulation package should hold.
2. **The format.** #318 proposes a plain block per entry — a timestamp/context/
   method header line, the SQL, and a `----` separator — chosen so no
   serialization dependency is added. That format is not parseable without
   assumptions when the SQL itself contains a line reading `----`. Is that
   acceptable, or does a format that cannot be corrupted by its own payload
   matter for an audit record specifically?
3. **The failure policy.** #318 says a write failure raises a base `warning()`,
   never a `marginplyr_error`, and the operation continues — "nothing about the
   call itself is wrong". ADR 0015 governs which conditions this package raises
   and why; check the reasoning against it rather than asserting it. And decide
   the case #318 explicitly defers: whether the warning is throttled to once per
   session, since an unthrottled one spams every iteration of a loop.
   ADR 0021 already decided how this package reports a repeated condition once —
   does it reach here?
4. **What an audit record is worth if it can silently stop being written.** A
   warning that the caller suppressed leaves an audit trail with a hole and
   nothing saying so. Is "never disrupts the operation it observes" the right
   principle for a record whose entire purpose is compliance?
5. **The timestamp.** #318 adds `time` to the record shape specifically because
   a file log needs it. If the file log is rejected, does `time` stay?

This ticket is not moot: #381 rejected the condition but kept the record, and
this ticket decides whether that record also reaches a file. It does become
moot if #383 rejects the record entirely.

## Resolution

The verdict and the shape — or a recorded rejection for ADR 0027's *Considered
options*, naming what a caller writes instead.

sayuks:
[...]

Issue #383, opened 2026-09-02
Decide whether a caller reads back the last call's SQL, and from what


sayuks:
Part of #377

## Question

Does a caller read back the last call's SQL from an accessor, or only from a
handler they wrote themselves?

#318 proposes `last_marginplyr_sql()` over a package environment that is reset
at the start of `prepare_margin_operation()` and appended to by each signal —
a last-call accessor, never an accumulator — justified by `rlang::last_error()`
and by the claim that this package has no package-environment pattern.

That claim is false: `share_dialect_verdicts` at `R/share.R:2301` is one. So the
question is not whether to introduce the pattern but **whether a second
environment, of the opposite shape, is warranted**. The precedent is a cache
that is never reset and is keyed by dialect; #318's is a log cleared on every
call. They agree on the mechanism and disagree on everything the mechanism is
for.

## What #380 established

#380 is closed. Its note is
`investigation/session-state-and-last-call-accessors.md` on branch
`research/session-state-precedents` (`c5bc14e`). Three findings bear directly:

- **`dbplyr::last_sql()` already exists**, new in dbplyr 2.6.0, and dbplyr is an
  Import here so it is never absent. Its whole body is `the$last_sql`, written
  by the single line `the$last_sql <- out` in `db_sql_render()` — which
  `collect()`, `compute()`, `show_query()` and `sql_render()` all funnel
  through. This is not merely an analogous precedent: it is the very package
  marginplyr renders through, doing this exact thing. What it does not do is
  distinguish anything — one slot, overwritten, no context, no method, no time —
  and marginplyr's own internal renders write to that same slot, so what a
  caller finds there after a margin operation is whichever render happened
  last, not necessarily their result. **That gap is the whole case for a
  marginplyr accessor, and this ticket has to make it explicitly.**
- **No precedent has #318's shape.** Eight accessors across seven packages:
  `rlang::last_error()`/`last_warnings()`, `dplyr::last_dplyr_warnings()`,
  `lifecycle::last_lifecycle_warnings()`, `dbplyr::last_sql()`,
  `httr2::last_request()`/`last_response()`, `duckplyr::last_rel()`,
  `ggplot2::last_plot()`. One clears at call start (httr2, and only the single
  slot it is about to fill), three flush lazily on the first write of a new
  top-level command, four never reset. **None** is a log emptied at call start
  and appended to several times within one call. `rlang::last_error()` — #318's
  sole cited precedent — also *errors* when empty, where #318 specifies the
[...]

Issue #382, opened 2026-09-02
Decide the "explain" capture method against ADR 0020


sayuks:
Part of #377

## Question

Does marginplyr gain `options(marginplyr.audit_sql_method = "explain")`, and if
so, what does ADR 0020 have to say to permit it?

This is the heaviest judgement on the map. ADR 0020 states that marginplyr sends
no query reading a lazy input's data unless the caller asked for one, and
exempts exactly two queries — **enumerated rather than derived from a shape**,
which is the sentence this ticket turns on.

#318 argues `EXPLAIN` is "a plan, not a result", so it is schema-shaped rather
than a data read. That is a shape argument, and ADR 0020 declines shape
arguments by construction.

Three readings are open, and the map's Notes forbid taking the third silently:

- **(a) Amend ADR 0020 with a third exemption.** `explain()` is a query against
  the caller's table, and permitting it is an amendment argued as one. The
  existing exemptions each carry their own warrant; this one would need its own.
- **(b) Reject `"explain"`.** Keep `"render"` as the only capture method, and
  record `"explain"` in ADR 0027's *Considered options*. `dbplyr::sql_render()`
  is purely client-side, so render-only leaves ADR 0020 untouched.
- **(c) Read the opt-in as the caller asking.** A caller who sets the option has
  asked for the query, so ADR 0020's rule is satisfied unamended. Attractive,
  but ADR 0020's exemptions are a list, and adding an interpretation that admits
  a query without adding to the list is itself an amendment — a silent one.

## What #379 established

#379 is closed. Its note is
`investigation/what-dplyr-explain-sends-per-backend.md` on branch
`research/explain-per-backend` (`4dff142`). Four findings bear directly:

- **The bare statement is invalid on at least three dbplyr-reachable dialects.**
  dbplyr's default is `EXPLAIN {sql}`; only postgres, sqlite, oracle and
  redshift override it. **GoogleSQL has no `EXPLAIN` statement at all**, so on
  BigQuery — the backend the cost argument was about — the question has no
  subject and the statement simply fails. SQL Server has none either outside
  Azure Synapse dedicated SQL pool; its equivalent, `SET SHOWPLAN_XML`, must be
  the only statement in its batch, which is not a shape dbplyr can prefix to a
  query. SAP HANA documents only `EXPLAIN PLAN ... FOR`, which writes to
  `EXPLAIN_PLAN_TABLE` instead of returning rows and needs `OPTIMIZER ADMIN`.
[...]

PR #388, opened 2026-09-02
Record the SQL marginplyr sends (ADR 0027)

### LINKED ISSUE #385: Write ADR 0027 and rewrite #318 to the settled specification
Part of #377

## Question

Write ADR 0027 and rewrite #318 to the specification this map settled.

Nothing is decided here — every verdict arrives from the closed tickets. This
is the assembly step that makes the destination reachable, and it is a `task`
for that reason.

What it produces:

1. **`design/adr/0027-<slug>.md`.** The slug follows from what survived, so it
   is not fixed in advance; 0027 is the next free number as of 2026-09-02
   (0001–0026 contiguous). Shape follows 0010/0014/0015: title, prose decision,
   `## Considered options`. Every feature this map decided against goes into
   *Considered options* with the argument that rejected it, including the
   per-entry-point `.audit_sql` argument.
2. **An amendment to ADR 0020**, if and only if #382 decided one is needed, in
   the form ADR 0020's existing amendment sections take.
3. **`CONTEXT.md` glossary entries** for whatever terms the surviving features
   name, each with its `_Avoid_` line. Per the map's Notes these are written by
   the ticket that fixed the name — this ticket only checks none is missing.
4. **#318 rewritten** to the settled specification, with its stale call sites
   and its false "no precedent" claim gone (see #378), so that whoever
   implements it reads a plan describing the code they have.
5. **What the `R/marginplyr-package.R` and `vignettes/database_backends.qmd`
   sections have to say**, written into #318 as part of item 4. This graduated
   from the map's fog when #384 closed, and is an assembly step rather than a
   decision: every sentence it must carry is already fixed — #381's promised
   `"result"` purpose and record-before-send, #383's accessor with its two
   refusals and its concurrency sentence, #382's single capture method, and
   #384's append recipe in place of a file log. Two constraints carry over: the
   section cannot be called *Audit signal*, since #381 rejected the condition,
   and #382 leaves one capture method to describe rather than two.
6. **A pointer to each `investigation/` note** the research tickets produced,
   cited from the ADR where its evidence is used — `AGENTS.md` requires the
   durable part to live in the ADR rather than the note.

If every feature was rejected, this ticket still runs: an ADR recording that
marginplyr does not signal its SQL, and why, is the destination just as much as
one recording that it does.

## Resolution

The ADR merged and #318 rewritten. Closing this ticket closes the map.

sayuks:
Closes #385, and closes map #377 — this is the map's last ticket.

## What this adds

- `design/adr/0027-record-the-sql-marginplyr-sends.md`. One feature of the
  four #318 bundled survives: a per-call record of the Sent queries, opt-in
  through `marginplyr.audit_sql`, read back with `last_sent_queries()`, and
  captured by a client-side `dbplyr::sql_render()`. The condition (#381), the
  `"explain"` capture method (#382), the on-disk log and the explicit writer
  (#384), and an `is_sql_lazy_query()` predicate (#386) are rejected in
  *Considered options*, each with the argument that rejected it, along with
  the per-entry-point `.audit_sql` argument and the alternatives #318 had
  already ruled out.
- One line in ADR 0020's *Related decisions*. **ADR 0020 is not amended**: the
  capture sends nothing, so there is no third exemption and the opt-in is not
  read as the caller asking. The line records that the question was put and
  refused, so a reader of ADR 0020 can find it.
- `investigation/what-dplyr-explain-sends-per-backend.md` and
  `investigation/session-state-and-last-call-accessors.md`, from the research
  branches for #379 and #380. The ADR cites both as evidence, and a citation
  to a note on a throwaway branch does not resolve.

## What this does not add

No R code. Writing the implementation is out of the map's scope; #318 now holds
the specification for it, rewritten to what the map settled.

`CONTEXT.md` needs nothing: **Sent query** entered from #381 and was widened by
#383, and a rejected feature names no term.

## Gates

`verify-context-budget.R` and `verify-verifier-invocation.R` pass. Nothing else
in the repository reads `design/` or `investigation/`, and no generated file is
affected.

### DIFF
--- a/design/adr/0020-ask-before-reading-a-lazy-input.md
+++ b/design/adr/0020-ask-before-reading-a-lazy-input.md
@@ -221,7 +221,9 @@ is what exemption 1 protects. ADR 0014 selects the per-kind entry that reads
 the source, and is amended separately too: its two lookups stand, but the
 second no longer selects a sampler, because there is nothing left to sample.
 ADR 0025 applies this decision's other half to a read this package does not
-issue, and is what the amendment above was written for.
+issue, and is what the amendment above was written for. ADR 0027 records the
+SQL marginplyr sends, and is where a third exemption for `EXPLAIN` was put to
+this decision and refused; nothing here changed as a result.
 
 Evidence: `investigation/query-cost-across-lazy-backends.md` and
 `investigation/share-source-eligibility-on-coercing-dialects.md`.
--- a/design/adr/0027-record-the-sql-marginplyr-sends.md
+++ b/design/adr/0027-record-the-sql-marginplyr-sends.md
@@ -5,23 +5,22 @@ back with `last_sent_queries()`. The record is off unless the caller sets
 `options(marginplyr.audit_sql = TRUE)`; it holds one row per Sent query, with
 the columns `purpose` and `sql` and no others; and it belongs to a single call,
 emptied at the top of `prepare_grouping_plan()`. Each row is written before its
-query is sent, so what the record holds is what was sent rather than what
-succeeded, and a call that fails leaves readable everything it had already sent.
+query is sent, which is the promise `CONTEXT.md`'s **Sent query** entry states.
 
 The capture is `dbplyr::sql_render()` and nothing else. It is client-side, sends
 nothing, and is not one of the execution entry points
 `lazy_execution_entry_points()` catalogs, so **ADR 0020 is not amended**: this
 decision adds no third exemption to it and does not read the opt-in as the
-caller having asked for a query. The four sites that do send — the zero-row
-selection proxy, the observed-label-collision scan, and the dialect probe's
-question and control — send exactly what they sent before this decision, whether
-or not a caller is auditing.
+caller having asked for a query. The four queries that do reach a connection —
+the zero-row selection proxy, the observed-label-collision scan, and the dialect
+probe's question and control — are exactly what they were before this decision,
+whether or not a caller is auditing.
 
 `marginplyr.audit_sql` is the package's first option, and `last_sent_queries()`
 is its first export that reads back what a previous call did; all fifteen others
-are verbs, constructors, or grouping helpers. Neither has a precedent inside
-this package to cite, which is why both are settled here rather than by
-analogy.
+are verbs, constructors, grouping helpers, or the inspection surface ADR 0013
+governs. Neither has a precedent inside this package to cite, which is why both
+are settled here rather than by analogy.
 
 ## Why a record rather than a signal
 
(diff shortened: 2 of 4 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, both axes, over `git diff main...HEAD` at
`bb19a76` (merge-base `5730166`). Ten findings; one was raised by both axes.
All were about prose, and all are answered in `c129cc2` by changing the prose
the finding was about — the diff is the record for each, per *Answering a
review* in `design/architecture.md`.

### Standards

| # | finding | answer |
|---|---|---|
| 1 | The opening restated `CONTEXT.md`'s **Sent query** sentence, which owns it, and can drift from it | the restatement is deleted and the glossary entry cited |
| 2 | *"the common ancestor of all five recorded sites"* is false — only the selection proxy is inside `prepare_grouping_plan()` | the claim is replaced by the two properties that are true; see below |
| 3 | *"It is not an audit record"* uses a spelling `CONTEXT.md` lists under **Sent query**'s `_Avoid_` | reworded to what the ground says |
| 4 | *"The only method is `explain.tbl_sql`"* is true of the SQL path, not of `explain()` | narrowed to the method a `tbl_lazy` reaches |
| 5 | *"all fifteen others are verbs, constructors, or grouping helpers"* — `inspect_grouping()` is none of the three | the inspection surface ADR 0013 governs is named |
| 6 | `investigation/session-state-and-last-call-accessors.md` wrote a bare `today`, which `investigation/README.md` forbids | states the date and points at `DESCRIPTION` |

### Spec

| # | finding | answer |
|---|---|---|
| 7 | The site count contradicts #383 and the ADR's own opener | same defect as Standards 2 |
| 8 | `grouping_backend()`'s header contract is quoted here and in #318; #386 verdict 4 handed it to #318 | the verbatim quote is dropped, the decision kept |
| 9 | `investigation/share-source-schema-vs-data-read.md` is cited for a render/read distinction it does not establish | citation dropped; #381 measurement 3 is named for what it measured |
| 10 | The `"explain"` ground's two claims are welded by a *so* that does not follow — #382 states them separately | unwelded |
| 11 | `last_marginplyr_sql()`, the name the map's Destination used, is replaced without record | recorded in *Considered options* with #383 verdict 9's argument |

Two findings landed on #318 rather than on this branch, #318 being the other
half of what ticket #385 produces: the vignette section's placement named a
range spanning a heading it did not mean, and the site-count sentence carried
Standards 2's error. Both are fixed in the issue body.

### The one that changed a claim

`prepare_grouping_plan()` is not the common ancestor of the recorded sites.
Only `grouping_selection_proxy()` is inside it; `check_observed_label_collision()`
is reached through `validate_margin_label()` from a sibling of its own caller,
and the dialect probe from `execute_margin_summary()`, downstream of neither.
What decides the placement is the two properties that do hold — all five entry
points reach it, and it runs before every one of the four sites on every path to
them — and that is what the ADR and #318 now say. #383's resolution carried the
looser wording, so this corrects the ticket as well as the branch.

Issue #389, opened 2026-09-02
A caller's share-source assertion has no scope wider than one call


sayuks:
> *This was generated by AI during triage.*

Split from #372, whose two documentation fixes are correct whichever way this
goes and are tracked there.

## Problem

`.check_share_source` is scoped to one call. On a dialect that converts a value
of another type to a number rather than refusing it — SQLite today, and the
class MySQL is expected to fall in — every share request needs the argument
again, and there is no connection-level or session-level place to record the
answer once.

Two consequences, both measured against the working tree at `bce5aef`:

1. It is not a one-time surprise. `R/` holds exactly one `getOption()` call and
   it reads `lifecycle_verbosity`, so nothing today offers a caller-set
   session default of any kind.
2. The argument is all-or-nothing for a call. A report carrying one share from
   a source the caller has established and one they have not loses the rule on
   both.

marginplyr already caches the *dialect's* answer: `share_dialect_verdicts` maps
a dialect class to `"refuses"` or `"converts"`, deliberately not caching
`"unknown"` because that is a fact about one attempt rather than about the
dialect. What has no home is the *caller's* answer — the assertion that a
particular source summary is eligible on a dialect that cannot be asked.

## Worth deciding

- Whether the caller's opt-out should have a scope wider than one call
  (an option, a connection attribute, a `local_*()` helper), or whether
  per-call is the right granularity and the answer is documentation.
- Whether per-share granularity is wanted, so a caller can vouch for one
  source and keep the rule on another, rather than opting out of both.

A wider scope inherits the objection recorded beside the verdict cache: a fact
recorded once governs every later connection carrying that dialect, and there
the package concluded that asking again was cheaper than being wrong. Whether
that reasoning transfers from a measured dialect fact to a caller's assertion
is the question, not a settled point.

Needs a human: this is an API-surface decision about a rule ADR 0010,
ADR 0014, and ADR 0020 set, not a specified change.

sayuks:
決定は PR #396。`.check_share_source` は1回の呼び出しに閉じたまま、share 単位の粒度も却下。記録は ADR 0010 の *Considered options*、論証は PR とコミットメッセージにあります。

PR #390, opened 2026-09-02
Say what the SQLite Parent-share row was tested under

### LINKED ISSUE #372: README promises SQLite Parent-share execution that the default refuses
Found while moving a finished local report onto RSQLite. The README's *Backend
verification* table reads:

| Verification status | Backends |
|---|---|
| Live portable Parent-share execution tested | SQLite |

so the expectation on arriving at a SQLite connection is that the same call
works. It does not:

```r
lite <- dplyr::copy_to(con, retail_sales, "retail_sales")

lite |>
  summarize_with_margins(
    revenue = sum(revenue, na.rm = TRUE),
    revenue_share = share_of_parent(revenue),
    .grouping = rollup(region, store)
  ) |>
  dplyr::collect()
#> Error:
#> marginplyr cannot establish that the source summaries of Parent shares are
#> plain integer or double scalars on this backend, because its SQL dialect
#> converts a value of another type to a number rather than refusing it, so an
#> ineligible source summary is indistinguishable from an eligible one.
#> ℹ Set `.check_share_source = FALSE` ...
```

The refusal is the documented rule and the diagnostic names its own argument, so
nothing here is wrong at the call site. What is wrong is that the README is
where a reader forms the expectation and `.check_share_source` appears nowhere
in it. The row above is true of the *tested* path, which is the one that passes
`.check_share_source = FALSE`; a reader has no way to tell that from the table.

Two things follow from the dialect being remembered rather than re-asked
(`R/share.R`, `verdict == "converts"`). First, this is not a one-time surprise:
every share request against SQLite — and against any other dialect that answers
the probe the same way, which is the class MySQL is expected to fall in — needs
the argument again, and there is no connection-level or session-level place to
record the answer once (`R/` has one `getOption()` call and it is unrelated).
Second, `.check_share_source` is scoped to the call, so a report carrying one
share from a source the caller is sure of and one they are not loses the rule on
both.

`database_backends.qmd` explains all of this correctly in *Parent shares use a
staged lazy query*. The gap is reachability: the section that a remote user
actually reads as a checklist, *Practical rules for remote analysis*, lists the
lazy `.check_margin_label` default but not this one, and the README's table
points the other way.

Worth deciding: whether the README table row should carry the condition it was
tested under, whether the remote checklist should name `.check_share_source`
beside `.check_margin_label`, and — separately — whether a caller should be able
to record a dialect answer once instead of per call.

sayuks:
Closes #372.

The README's *Backend verification* table claimed live portable Parent-share
execution for SQLite. A reader arriving at an RSQLite connection with
documented defaults is refused: the tested path passes
`.check_share_source = FALSE`, and that argument appeared nowhere in either
half of the README. The row now carries the condition and says that the
default refuses the share on the dialects the row names, so the table alone
tells a tested path from one the reader's own defaults reach. It says that of
"these dialects" rather than of SQLite, so it stays true when another
converting dialect joins the cell — MySQL is expected to.

The database-backends vignette's *Practical rules for remote analysis*
checklist gains one bullet beside the `.check_margin_label` one. It names
`.check_share_source`, the two kinds of backend the rule cannot be applied on,
and cites *Parent shares use a staged lazy query*, which owns the argument,
rather than re-deriving it.

Nothing under `R/` changes: the refusal is correct and its diagnostic already
names its own remedy. Whether a caller can record a dialect answer once
instead of per call is #389.

## Verification

| check | result |
|---|---|
| `verify-context-budget.R` | pass, 24198 bytes, unchanged |
| `spelling::spell_check_package()` | clean |
| `lintr::lint_package()` after `pkgload::load_all(".")` | no lints |
| `testthat::test_local()` with `NOT_CRAN` set | pass |
| `rmarkdown::render("README.Rmd")` | regenerated against the installed working tree under pandoc 3.10.1, the version `document.yaml` pins |
| `altdoc::render_docs(parallel = FALSE, freeze = FALSE)` then `verify-site.R` | pass, 19 pages |

The site render was run from a copy of the tree outside this worktree: the
worktree lives under `.claude/`, and quarto ignores a dot-prefixed path.

### DIFF
--- a/README.Rmd
+++ b/README.Rmd
@@ -335,7 +335,7 @@ dplyr code may be easier.
 | Verification status | Backends |
 |---|---|
 | Live native execution tested | DuckDB |
-| Live portable Parent-share execution tested, with `.check_share_source = FALSE`; the default refuses a Parent share on any dialect that converts an ineligible share source rather than refusing it | SQLite |
+| Live portable Parent-share execution tested, with `.check_share_source = FALSE`; the default refuses the share on these dialects | SQLite |
 | Native SQL generation tested | PostgreSQL |
 | Fallback SQL generation tested | Access, SAP HANA, Hive, Impala, MariaDB, Microsoft SQL Server, MySQL, Oracle, Amazon Redshift, Snowflake, Spark SQL, SQLite, Teradata, and generic DBI/ODBC connections |
 | Non-SQL lazy backend tested | Arrow, dtplyr |
--- a/README.Rmd
+++ b/README.Rmd
@@ -335,7 +335,7 @@ dplyr code may be easier.
 | Verification status | Backends |
 |---|---|
 | Live native execution tested | DuckDB |
-| Live portable Parent-share execution tested | SQLite |
+| Live portable Parent-share execution tested, with `.check_share_source = FALSE`; the default refuses a Parent share on any dialect that converts an ineligible share source rather than refusing it | SQLite |
 | Native SQL generation tested | PostgreSQL |
 | Fallback SQL generation tested | Access, SAP HANA, Hive, Impala, MariaDB, Microsoft SQL Server, MySQL, Oracle, Amazon Redshift, Snowflake, Spark SQL, SQLite, Teradata, and generic DBI/ODBC connections |
 | Non-SQL lazy backend tested | Arrow, dtplyr |
--- a/README.md
+++ b/README.md
@@ -456,7 +456,7 @@ subtotal, explicit dplyr code may be easier.
 | Verification status | Backends |
 |----|----|
 | Live native execution tested | DuckDB |
-| Live portable Parent-share execution tested | SQLite |
+| Live portable Parent-share execution tested, with `.check_share_source = FALSE`; the default refuses a Parent share on any dialect that converts an ineligible share source rather than refusing it | SQLite |
 | Native SQL generation tested | PostgreSQL |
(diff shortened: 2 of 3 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` ran once over `git diff 130a1a3...HEAD`, the
merge-base with `main`, at `4c627c0`. Standards was passed *Documentation*,
*Code comments*, *Site verification*, and *Linting* from `AGENTS.md`, plus
*Answering a review* from `design/architecture.md` as what a finding about
prose may be; Spec was passed #372 and the agent brief in its triage comment.
`README.md` was excluded as generated, per `design/agents/code-review.md`.

Three findings, all against the two sentences the branch added, all answered
in prose by `bb2add2`:

| axis | finding | disposition |
|---|---|---|
| both | the bullet's umbrella "a backend that cannot be asked" is false — that is `share_dialect_can_be_asked()`, which a live SQLite connection passes; it is asked, answers `"converts"`, and the answer is remembered | fixed in prose |
| Standards | the bullet said "A Parent share" while `.check_share_source` governs every contextual share, `share_of_total()` included | fixed in prose |
| Standards | the README row's trailing clause re-derived why a converting dialect refuses, an argument the staged-lazy-query section and `share_of_parent()`'s roxygen each own | fixed in prose |

Spec also read the first version of the row as leaving membership in the
converts class unstated — the reader had to already know SQLite converts. The
Standards fix above answers that too: the row now says the default refuses the
share on the dialects it names, which is the fact the acceptance criterion
asked the table to carry, rather than the rule it follows from.

Spec found no scope creep. No baseline smell applied — no code changed.

The commit that answers each finding names it; the diff it produced is the
whole of what these dispositions record, per *What is reviewed, and where the
record goes*.

Issue #392, opened 2026-09-02
Vignette prose describing a diagnostic printed directly below it


sayuks:
#374 deleted a clause from `vignettes/database_backends.qmd` — "and the refusal
names the classes that can" — sitting directly above the `must_error` chunk that
renders the refusal it describes. Nothing compares the two, so the clause is
free to age into a false claim about a diagnostic printed three lines below it.
That is the argument `AGENTS.md`'s *Code comments* makes about a second copy,
and it scopes itself to `R/`.

The question is whether this is one editorial call or a form the vignettes
repeat, and it was left open in #374 because it had not been measured. It has
been now. All 21 `must_error` chunks under `vignettes/` were scanned, and six
of them carry a sentence describing what the refusal below them contains:

| chunk | the clause |
| --- | --- |
| `database_backends.qmd`, nesting a SQL table | `and the refusal names the classes that can` |
| `database_backends.qmd`, Arrow summary | `and tells you the two ways to get it computed` |
| `get_started.qmd`, duplicate grouping sets | `and names the positions that collide` |
| `get_started.qmd`, grouping helper in `filter()` | `a Package condition naming the replacement, not a silently different answer` |
| `get_started.qmd`, `cube()` parent share | `names the rewrite instead of choosing a parent` |
| `get_started.qmd`, `grouping_sets()` without an empty set | `is rejected naming that fix` |

**They are not all the same thing, and that is what this has to decide.** Two of
the six carry a contrast the diagnostic itself does not make — "not a silently
different answer", "instead of choosing a parent" — and a reader learns
something from those that the rendered refusal does not tell them. The rest are
closer to bare restatement: they promise the diagnostic names X, and the
diagnostic below names X.

So the options are not "add a clause" or "do nothing":

1. **A clause in `AGENTS.md`** widening *Code comments*' second-copy argument to
   prose standing beside a rendered diagnostic, with the contrast case named as
   what it does not reach. Its home would have to be settled — that section is
   about `R/`, and #374's own decision was that a rule with a home does not get
   a second one.
2. **Edit the bare restatements and leave the contrasts**, recording nothing.
   Three sites is not a pile, and `AGENTS.md` says a hand audit is the answer
   where a scanner cannot express what makes the text wrong — which is exactly
   the case here, since the difference between the two groups is semantic.
3. **Nothing.** The clauses are prose about the page's own argument, and a
   vignette is allowed to say what it is about to show.

[...]

Issue #391, opened 2026-09-02
expand_with_margins() refuses a RecordBatchReader without naming a way forward


sayuks:
`expand_with_margins()` refuses a `RecordBatchReader` without naming a way
forward, which is the defect #374 fixed for the nesting verbs:

```r
r <- arrow::as_record_batch_reader(arrow::as_arrow_table(data.frame(g = c("a", "b"), v = 1:2)))
expand_with_margins(r, .grouping = rollup(g))
#> Error:
#> `.data` must not be an object of the following classes: `RecordBatchReader`.
```

`assert_lazy_table()` in `R/expand_with_margins.R` names the class it refuses
and stops. The reader has `dplyr::group_vars()` and `dplyr::collect()` methods,
so it passes `assert_margin_input()` and reaches this guard from an ordinary
lazy pipeline.

This was found while measuring #374 and split out from it deliberately: the
remedy is not settled by that analysis. A `RecordBatchReader` is a one-shot
stream, so `dplyr::collect()` and `arrow::as_arrow_table()` do not leave the
caller in the same place — one consumes the reader and produces a local frame,
the other materializes an Arrow table the Margin verbs accept. Which is the
remedy, or whether both are, is what this issue has to decide, under the rule
ADR 0012's amendment states: *a remedy is offered only where it is one.*

The nesting whitelist refuses the same object for a different reason, so
whatever is decided here should be checked against #374's bullet — a caller who
follows one refusal's remedy must not arrive at the other's.

sayuks:
> *This was generated by AI during triage.*

## Triage: ready-for-agent

**Category:** enhancement. **State:** `ready-for-agent` — the decision this
issue asked for is made below, and what remains is one `i` bullet, a header
sentence, and a test pin.

**Verified against `80ff86b`**, which is `origin/main`'s content for every file
below. Four verbs, one `RecordBatchReader` each:

| verb | guard reached | diagnostic |
|---|---|---|
| `summarize_with_margins()` | `assert_lazy_table()` | refuses, no remedy |
| `expand_with_margins()` | `assert_lazy_table()` | refuses, no remedy |
| `inspect_grouping()` | `assert_lazy_table()` | refuses, no remedy |
| `nest_with_margins()` | `assert_nest_possible()` | refuses, names `dplyr::collect()` |

So the defect is not `expand_with_margins()`'s: the guard is shared by three
verbs, and the bullet added to it is read from all three. `nest_with_margins()`
[...]

PR #393, opened 2026-09-02
Name a way forward when the nesting verbs refuse a lazy table (#374)

### LINKED ISSUE #374: Nest verbs refuse a lazy table without naming either way forward
Found while working a DuckDB-backed report and wanting the source records behind
each subtotal:

```r
nest_with_margins(lite, .grouping = rollup(region, store))
#> Error:
#> `.data` must be one of the following classes, which can be nested:
#> `data.frame` or `dtplyr_step`.
```

The restriction is right and the reference states it. The diagnostic names the
classes it accepts and stops there, so it says what the input is not and nothing
about what to do, which leaves a reader at a dead end in the middle of a lazy
pipeline. Two routes exist and neither is discoverable from here:
`dplyr::collect()` first, or `expand_with_margins()`, which takes the same
grouping specification and does work against the same connection:

```r
class(expand_with_margins(lite, .grouping = rollup(region, store)))
#> [1] "tbl_SQLiteConnection" "tbl_dbi" "tbl_sql" "tbl_lazy" "tbl"
```

`assert_nest_possible()` in `R/utils.R` carries no remedy bullet, unlike most
refusals in the package. The comment above it is about how the class list is
rendered rather than about what the caller should do next.

The README contributes to the wrong expectation by pairing the two verbs in one
bullet — "Can expand or nest the source rows behind every subtotal" — without
saying that their backend ranges differ, so a lazy user has no reason to suspect
one of them is unavailable until the call fails.

Worth deciding: whether the refusal should point at `collect()` or at
`expand_with_margins()` (they answer different questions — one moves the data,
one stays lazy but returns rows rather than list columns), and whether the
README bullet should separate the two verbs.

sayuks:
Closes #374.

`assert_nest_possible()` named the classes it accepts and stopped. It now
carries a remedy bullet:

```
`.data` must be one of the following classes, which can be nested:
`data.frame` or `dtplyr_step`.
i Collect it with `dplyr::collect()` first.
```

The decision the ticket existed to take — `dplyr::collect()` and not
`expand_with_margins()` — is argued in [#374's decision
comment](https://github.com/sayuks/marginplyr/issues/374#issuecomment-5506835548),
and the commit message carries what each edit rests on. In short: the sibling
verb answers a different question (rows, not the list column this call asked
for) and is not a route every input reaching this refusal can take, which is
the case ADR 0012's amendment names under *a remedy is offered only where it is
one*.

Also here: the README bullet now carries the narrower verb's backend range, and
`database_backends.qmd` loses a clause describing the diagnostic rendered three
lines below it.

Split out, deliberately: #391 (`assert_lazy_table()`'s own missing route) and
#392 (whether that vignette clause is one site or a form).

Checks: `lintr::lint_package()` clean; full suite 6263 passing, no failures, no
skips; `verify-context-budget.R` and `verify-verifier-invocation.R` pass;
`README.md` regenerated against the installed working tree under
`document.yaml`'s pandoc 3.10.1 pin.

### DIFF
--- a/.github/scripts/verify-site.R
+++ b/.github/scripts/verify-site.R
@@ -388,7 +388,13 @@ markers <- list(
     # what ADR 0023's condition 3 keeps out of a marker. #223's phase 3 left
     # this marker in place deliberately until the file behind it was
     # re-authored, since moving it is the same edit.
-    "which can be nested"
+    "which can be nested",
+    # That refusal's remedy, added in #374, quoted whole: it interpolates
+    # nothing, so ADR 0023's condition 3 leaves every run of it available and
+    # the longest one pins the most. Its backticks are the diagnostic's own
+    # and reach the page literally, as the `Error in` marker two entries up
+    # already does.
+    "Collect it with `dplyr::collect()` first."
   ),
   "docs/man/summarize_with_margins.html" = c(
     "summarise_with_margins",
--- a/R/utils.R
+++ b/R/utils.R
@@ -51,14 +51,23 @@ assert_string_scalar <- function(x) {
 # joins a vector with by default would read as a conjunction. This is the same
 # call `match_margin_choice()` makes about an option vocabulary, and the
 # `toString()` comma both replace was neutral between the two readings.
+#
+# `expand_with_margins()` is not named beside `dplyr::collect()`, under the
+# rule ADR 0012's amendment states. It takes the same grouping specification
+# and stays lazy, but it returns rows rather than the list column this call
+# asked for, and `assert_lazy_table()` refuses a `RecordBatchReader` that
+# reaches this refusal.
 assert_nest_possible <- function(x) {
   # Read only from the cli template below, which codetools cannot see.
   nm <- deparse(substitute(x)) # nolint: object_usage_linter.
   valid_classes <- c("data.frame", "dtplyr_step")
   if (!inherits(x, valid_classes)) {
-    abort_marginplyr(paste0(
-      "{.arg {nm}} must be one of the following classes, which can be ",
-      "nested: {.or {.code {valid_classes}}}."
+    abort_marginplyr(c(
+      paste0(
+        "{.arg {nm}} must be one of the following classes, which can be ",
+        "nested: {.or {.code {valid_classes}}}."
+      ),
+      i = "Collect it with {.fun dplyr::collect} first."
     ))
   }
 }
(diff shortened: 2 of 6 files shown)


sayuks:
## Review record

Round run through `mattpocock-skills:code-review` against `origin/main...HEAD`
at `c5bad30`, both axes. Standards carried `AGENTS.md` (*Code comments*,
*Linting*, *Documentation*, *Site verification*, *Chunks that must fail*),
`design/architecture.md`'s *Answering a review*, and ADR 0012, 0023, and 0024.
Spec carried #374's decision comment.

**Standards — one hard violation, two judgement calls.**

| finding | disposition |
| --- | --- |
| `R/utils.R`'s new header said `assert_lazy_table()` refuses a `RecordBatchReader` "this guard admits". `this guard` is `assert_nest_possible()`, which refuses it — the admitting guard is `assert_margin_input()`. False prose. | Fixed in code. Reads "that reaches this refusal". |
| `verify-site.R`'s comment cited the neighbouring code-span comment, which argues about apostrophes surviving pandoc `smart`, for a claim about backticks. | Fixed in prose. The clause is deleted; the `Error in` marker two entries up is the checkable evidence and stays. |
| No test asserts the new bullet. | Fixed in code — see below. |

**Spec — no missing requirements, one unauthorized change.**

The four edits landed with the pinned text and the three negative requirements
held. The site marker ships as the whole sentence where the decision said
`"Collect it with"`. Kept: it is a superset of the specced marker, so it cannot
pass where that one would fail, and `"Error in \`nest_with_margins()\`"` — already
a marker on this page — is direct evidence that a diagnostic's backticks reach
the HTML literally. The reason the decision gave for the *shorter* marker was
false, and #374's decision comment has been corrected rather than restated.

**The rejection that collapsed.** The decision recorded that no testthat pin
would be added because "no remedy bullet in this package has one". That premise
was mine and it was false — it came from grepping two phrasings. The count:

```console
$ grep -rc '"i [A-Z]' tests/testthat/*.R | grep -v ':0'
tests/testthat/test-execution-conditions.R:18
tests/testthat/test-grouping-plan.R:19
tests/testthat/test-diagnostic-pluralization.R:8
tests/testthat/test-contextual-helpers.R:6
tests/testthat/test-inspect-grouping.R:2
tests/testthat/test-grouping-backends.R:2
tests/testthat/test-grouping-interface.R:1
tests/testthat/test-static-expression-analysis.R:1
```

Fifty-seven across eight files, including `"i Use \`grouping_set()\` for the
empty grouping set."`. So the rejection goes and `test-utils.R` pins the bullet.
It fails without the fix: the matcher returns `FALSE` against the message the
guard produced before this branch, and `TRUE` against the one it produces now.

Nothing else was raised. No baseline smell was reported: the diff is five small
hunks, and the `Collect it with` / `Convert it with` echo of the sibling guard
is deliberate parallelism.

Issue #144, opened 2026-08-09
A `NULL` option value is now rejected, where `.key` and `.id` document `NULL` as "use the default"


sayuks:
Split out of #110. That ticket asked for exact, documented option values and is complete without this; the change that delivered it (#143) also narrowed `NULL`, which the ticket did not name. Recording the narrowing so it is a decision rather than a side effect.

## What changed

`match_margin_choice()` used `match.arg()`, and `match.arg(NULL, choices)` returns `choices[1]`. Passing `NULL` to any fixed-vocabulary option therefore selected that option's default, silently. Since #143 it is an error, with the message any other invalid value gets:

```r
d <- data.frame(g = c("a", "b"), v = 1:2)

summarize_with_margins(d, s = sum(v), .grouping = rollup(g), .sort = NULL)
#> before: accepted, identical to .sort = "none"
#> now:    `.sort` must be one of "none", "last", "first".
```

The same holds for `.duplicates`, `.margin_label_position`, and `inspect_grouping()`'s `.format` — all four go through the one helper.

Nothing documented the old behaviour, and nothing documents the new one. `?summarize_with_margins` describes `.sort` as taking `"none"`, `"last"`, or `"first"`, and is silent on `NULL` either way.

## Why this needs a decision rather than a fix

The narrowing follows from "only exact, documented values are accepted", which is what #110 asked for, and it is the behaviour I would defend: a `NULL` reaching an option argument is far more often a variable that did not hold what the caller thought than a deliberate request for the default. But this package already gives `NULL` the opposite meaning on two arguments, both deliberately and both documented:

- `.key` — "As in `tidyr::nest()`, `NULL` uses `"data"`", implemented as an explicit `if (is.null(.key))` in `nest_with_margins()`.
- `.id` — `NULL` is the documented default, meaning no Grouping set identifier column.

So "`NULL` means leave it alone" is an idiom the API uses, and the fixed-vocabulary options are now the place it does not hold. That inconsistency is defensible — `.key` and `.id` are values with a natural absent case, an option vocabulary is not — but it is currently undocumented in both directions, which is the part worth settling. A caller writing `.sort = user_choice %||% NULL` gets an error whose message lists three values, none of which is the thing they passed.

Either answer implies a small documentation change rather than a code change:

- **Keep rejecting** (my recommendation): say so where the options are documented, so the difference from `.key` and `.id` is stated rather than discovered. `@param .sort` and its siblings would gain a clause.
[...]

PR #394, opened 2026-09-02
Decide that a caller writes a Margin order key of their own

### LINKED ISSUE #373: A Margin order cannot run newest-first, and nothing shows how to rebuild its sort key
Found while preparing a quarterly report for a management meeting, where the
standard presentation is newest period first. `.sort` gives the subtotal
adjacency and nothing else:

```r
retail_sales |>
  summarize_with_margins(
    revenue = sum(revenue),
    .grouping = rollup(year, region),
    .sort = "last"
  )
#>    year region revenue
#> 1  2025   East   16200
#> 2  2025   West   17700
#> 3  2025  Total   33900
#> 4  2026   East   25700
#> 5  2026   West   27800
#> 6  2026  Total   53500
#> 7 Total  Total   87400
```

`"none"`, `"last"`, and `"first"` decide where a margin sits relative to the
rows it summarizes; the value order underneath is always ascending. What the
report needs is 2026 above 2025 with the subtotals still attached, and
`arrange()` — the answer recipes.qmd gives for "some other presentation order" —
drops the adjacency that made `.sort` worth asking for.

It is reproducible, and cheaply, once the Grouping bits are in the result:

```r
retail_sales |>
  summarize_with_margins(
    revenue = sum(revenue),
    yb = grouping_bit(year), rb = grouping_bit(region),
    .grouping = rollup(year, region)
  ) |>
  dplyr::arrange(yb, is.na(year), desc(year), rb, is.na(region), region)
#>    year region revenue yb rb
#> 1  2026   East   25700  0  0
#> 2  2026   West   27800  0  0
#> 3  2026  Total   53500  0  1
#> 4  2025   East   16200  0  0
#> 5  2025   West   17700  0  0
#> 6  2025  Total   33900  0  1
#> 7 Total  Total   87400  1  1
```

The key is `bit, missingness, value` per dimension in plan order, which is what
the `.sort` argument documentation already states in prose. But it states it as
a description of what the argument does, not as something a caller is expected
to write, and no vignette shows it as a call. recipes.qmd's *Put each subtotal
with the rows it summarizes* covers the ascending case and then sends anyone
wanting a different order to a bare `arrange()`, which is the one thing that
does not work here. So a reader who needs a reversed dimension has to derive the
key from the argument prose, including that `grouping_bit()` has to be requested
in the summary for it to be available at all.

The same shape blocks the other common presentation request — regions ordered by
revenue, stores ordered by revenue inside them, subtotals still adjacent — which
needs the parent measure on every row and so a self-join back onto the subtotal
rows. That one is genuinely outside what a sort key can express, but it lands in
the same documentation hole.

Worth deciding: whether reversing a dimension belongs in `.sort`, or whether the
answer is a recipe showing the sort key written as `arrange()` so any direction —
and a measure-based order — can be built from it.

sayuks:
Closes #373.

`.sort` orders every dimension's values ascending, so a report wanting the
newest period at the top had no argument to ask for it, and the one section
covering row order sent such a reader to a bare `arrange()` — the one thing
that drops the subtotal adjacency they came for.

**The decision: the key stays a caller's to write.** `.sort`'s vocabulary stays
`"none"`, `"last"`, and `"first"`.

A direction argument would not retire the recipe it competes with. It reaches
the reversed dimension and not the request one step past it — regions ordered
by revenue, stores ordered by revenue inside them — which needs each region's
own measure on every one of its rows and so is outside what any ordering over
the result's columns expresses. A caller writes that key by hand whichever way
this went. Its own cost is the vocabulary: `match_margin_choice()` admits one
string from a fixed list and reads an untouched formal by `identical()` (#110,
#144), a per-dimension direction has to name dimensions that a composite has no
single name for, and `.sort` is shared by four verbs.

ADR 0010's `.subtotals = "one"` rejection is not the precedent. "A presentation
rule" there is about a computed value whose definition a presentation choice
would obscure; `.sort` computes no value.

## What was measured

- **Dropping a Grouping bit does not lose the order.** ADR 0018 rejected
  `.bits` partly because "`arrange()` followed by dropping a summarized column
  is exactly the shape dbplyr flattens". Against dbplyr 2.6.0 that does not
  reach a Grouping bit on DuckDB or SQLite: a bit is produced by the aggregate
  query, which the labelling projection already wraps, so it stays resolvable
  in the `FROM` of the query carrying the `ORDER BY`. The amendment narrows
  that clause; the rejection stands on its other half.
- **Dropping a column computed after the summary does.** DuckDB refuses it with
  `Binder Error: Referenced column "region_revenue" not found in FROM clause!`
- **The measure-ordered case needs no self-join**, contrary to the ticket and
  its triage. The region's subtotal row is already in the result, so a grouped
  `mutate()` puts the measure on every row and dbplyr renders it as a window
  function.
- **`compute()` returned the intended order on DuckDB without promising it.**
  A materialized table has no row order in SQL and the following `select()`
  renders a query with no `ORDER BY`, so the guide does not show it; the
  investigation note records the measurement so it is not taken for a route.

## Changes

| File | Change |
|---|---|
| `design/adr/0018-…` | One amendment: the decision, and the narrowed `.bits` clause |
| `investigation/writing-a-margin-order-key-by-hand.md` | New note — three backends, the binder error, why `compute()` is not shown |
| `vignettes/recipes.qmd` | New section *Write the sort key yourself*; the closing sentence that sent readers away is replaced |
| `.github/scripts/verify-site.R` | The section title as a marker |
| `R/summarize_with_margins.R`, `man/` | One clause on the *Margin order* pointer |
| `NEWS.md` | One clause on the existing `.sort` bullet |

The new section's `must_error` chunk is not a marker: it needs a database, and
the comment there says only the database-free ones are listed. It is
`must_error: true` and not a class — what refuses the call is dbplyr's
flattening reaching DuckDB's binder, so pinning `duckdb_error` would fix the
assertion to one dialect's condition class where the prose names neither.

## Checks

`lintr::lint_package()` clean, `spelling::spell_check_package()` clean, full
suite passing with no failures and no skips under `NOT_CRAN`,
`verify-context-budget.R` and `verify-verifier-invocation.R` pass. Every chunk
the new section adds was run against DuckDB 1.5.5 and against the working tree,
and the measure-ordered pipeline against SQLite 3.53.3 as well. `README.md`
needed no regeneration.

### DIFF
--- a/.github/scripts/verify-site.R
+++ b/.github/scripts/verify-site.R
@@ -319,6 +319,7 @@ markers <- list(
   ),
   "docs/vignettes/recipes.html" = c(
     "Put each subtotal with the rows it summarizes",
+    "Write the sort key yourself",
     "Compare each row with the right total",
     "Keep only one grouping set",
     "Name the grouping set on every row",
--- a/NEWS.md
+++ b/NEWS.md
@@ -91,9 +91,9 @@
   property of the returned object and may not survive further verbs applied to
   a lazy result. `compute()` materializes a sorted lazy result in the Margin
   order; it records no dbplyr window ordering, because the key reads Grouping
-  bits from a column the result does not expose. Another presentation order —
-  a dimension descending, or rows ordered by a measure — is the same key
-  written as an `arrange()` of your own, which the recipes guide shows.
+  bits from a column the result does not expose. A dimension descending is the
+  same key written as an `arrange()` of your own, and rows ordered by a measure
+  is that key with one column added; the recipes guide shows both.
 * Added guides for Grouping identity and explicit key completion, and made the
   function references the canonical source of the Margin, Parent-share, and
   Margin-label contracts.
--- a/NEWS.md
+++ b/NEWS.md
@@ -91,7 +91,9 @@
   property of the returned object and may not survive further verbs applied to
   a lazy result. `compute()` materializes a sorted lazy result in the Margin
   order; it records no dbplyr window ordering, because the key reads Grouping
-  bits from a column the result does not expose.
+  bits from a column the result does not expose. Another presentation order —
+  a dimension descending, or rows ordered by a measure — is the same key
+  written as an `arrange()` of your own, which the recipes guide shows.
 * Added guides for Grouping identity and explicit key completion, and made the
   function references the canonical source of the Margin, Parent-share, and
   Margin-label contracts.
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -341,9 +341,9 @@
 #' `.duplicates` policy the verb accepts with no diagnostic, and lazy inputs
 #' stay lazy.
 #'
-#' The [recipes guide][recipes] shows a Margin order before and after, and
-#' shows a join dropping one from a lazy result with no diagnostic naming this
-#' package.
+#' The [recipes guide][recipes] shows a Margin order before and after, shows
+#' the key written as an `arrange()` you control, and shows a join dropping a
+#' Margin order from a lazy result with no diagnostic naming this package.
 #'
 #' [recipes]: https://sayuks.github.io/marginplyr/vignettes/recipes.html
 #'
--- a/design/adr/0018-order-margin-results-by-grouping-structure.md
+++ b/design/adr/0018-order-margin-results-by-grouping-structure.md
@@ -177,6 +177,15 @@ order is lost on lazy backends. The trap moves out of marginplyr into caller
 code that nothing designed. Keeping the bit columns to keep the order is not
 the result the caller asked for.
 
+**A per-dimension direction on `.sort`**, so that a dimension can descend
+under a Margin order (#373). Rejected: it reaches a reversed dimension and not
+the request one step past it, ordering dimensions by a measure, which needs a
+value no ordering over the result's columns produces. A caller writes that key
+themselves either way, and the documentation carrying it carries the reversed
+dimension too, so the argument would not retire what it duplicates. It is also
+the wrong shape for `match_margin_choice()`, which admits one string from a
+fixed list, and it would have to name dimensions a composite has none of.
+
 **Leaving both out and keeping row order unspecified.** Rejected: the
 existing guidance tells a caller to add a `grouping_bit()` summary per
 dimension and drop it again, which fails on lazy backends for the reason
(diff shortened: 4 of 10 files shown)


sayuks:
## Code review record

`mattpocock-skills:code-review`, one round, two axes, against `origin/main` (merge-base `f38ca8c`), reading `5daea1c`. Seven findings; answered in `ce1e116`.

|---|---|---|---|
| 1 | Standards / Spec | The rule the amendment and the guide state — a column the summary produced sits below the query carrying the `ORDER BY`, so dropping it keeps the order — is false for a plan of one grouping-set occurrence under `.margin_label = NULL`; and the guide stated dbplyr's flattening as a durable rule where ADR 0018 declined to promise one | Fixed in prose. Measured on both adapters and recorded in the note; the amendment gains the exception, and the guide now names the signal (a dbplyr warning, or the database's refusal) instead of the optimizer |
| 2 | Standards | The amendment re-made the ticket's case — `design/architecture.md` *Repository placement*: "an amendment that makes the case again is a second copy of the ticket's" | Fixed. The rejection of a per-dimension `.sort` moved to *Considered options*, where `design/agents/code-review.md` routes a weighed alternative |
| 3 | Spec | The `.bits` entry's twin, "Leaving both out and keeping row order unspecified", still rejects this branch's recipe for the withdrawn reason and adds "unavailable at all under `.margin_label = NULL`", which the branch's own measurement contradicts | Fixed. The amendment withdraws that clause too |
| 4 | Standards | `NEWS.md` called the measure-ordered case "the same key", where the ADR in the same commit says it is outside what a sort key expresses | Fixed. It is that key with one column added |
| 5 | Spec | The guide covered half of #373's second request: regions ordered by revenue, but not the stores inside them or the adjacency | Fixed. The prose names the `sb` term that keeps each subtotal under its own stores |
| 6 | Standards | `must_error: duckdb_error` over `must_error: true`, the chunk being DuckDB-guarded | Rejected, with the reproduction. The chain reaching the reader is `rlang_error` at both levels — dbplyr's "Failed to collect lazy table." wrapping DBI's — and carries no `duckdb_error` class to match, so the class form would fail the render |
| 7 | Standards | Duplicated connection setup and its comment across three chunks | Fixed, in the narrower shape: the file had one copy of the comment and this branch added a second. There is one again, at the first connection in reading order |

Also fixed: the investigation note lost two sentences that described the present and named a consuming artifact, which `investigation/README.md` refuses under *Tense* and *Pointers*.

Checks after the answers: `lintr::lint_package()` clean, `spelling::spell_check_package()` clean, full suite passing with no failures and no skips under `NOT_CRAN`, `verify-context-budget.R` and `verify-verifier-invocation.R` pass. Every chunk the section shows was rerun, and the one-occurrence shape was measured on both adapters with and without `.margin_label`.

PR #395, opened 2026-09-02
Decide that an ancestor share is a documented join

### LINKED ISSUE #375: A three-level rollup has no share against an intermediate ancestor
Found on a region → store → channel report, where the question the reader asks
about a channel row is what fraction of its *region* it is. The two share
helpers reach the level immediately above and the top, and nothing between:

```r
retail_sales |>
  summarize_with_margins(
    revenue = sum(revenue),
    share = share_of_parent(revenue),
    .grouping = rollup(region, store, channel),
    .sort = "last"
  ) |>
  head(8)
#>   region    store channel revenue     share
#> 1   East   Boston  Online   13200 0.5789474
#> 2   East   Boston   Store    9600 0.4210526
#> 3   East   Boston   Total   22800 0.5441527
#> 4   East New York  Online    3000 0.2702703
#> 5   East New York   Store    8100 0.7297297
#> 6   East     <NA>  Online    8000 1.0000000
```

Row 1's `share` is against Boston, row 3's is against East. `share_of_total()`
covers the other end. A channel's share of East, or of anything that is neither
its parent nor the Grand total set, has to be assembled by filtering the
`grouping_bit()` pattern for the ancestor level and joining it back on the
dimensions that survive at that level — which is the structural parent matching
the package already does internally, written out again by the caller and using
`grouping_bit()` rather than labels to stay correct against source `NA` and
label collisions.

The workaround is short, so this is not urgent. What makes it worth recording is
that three levels is an ordinary depth for a management hierarchy, and the
helpers' names do not suggest a depth limit — `share_of_parent()` reads as
"relative to what contains me", and a channel is contained by a region as much
as by a store. #83 scoped the pair as parent versus grand total without
recording an intermediate variant as deferred or rejected, so nothing currently
says whether the gap is a decision or an omission.

Worth deciding: whether a share against a named ancestor level belongs in the
API, or whether the answer is a documented recipe for the bit-pattern join, so
that the correctness details (missing-safe keys, not filtering on labels) are
stated once rather than rediscovered.

sayuks:
Closes #375.

A share against a named ancestor level — what fraction of its region a channel
row is, in `rollup(region, store, channel)` — is refused as API and answered as
a recipe. ADR 0010 records the decision; the recipes guide owns the join.

**The `cube()` deferral does not reach it.** That deferral rests on a grouping
set having several strictly less detailed sets in the same plan. A rollup's
chain is a total order, so an ancestor two levels up is as unambiguous as the
immediate parent, and refusing it on parent ambiguity would copy a restriction
for a reason that does not apply — what ADR 0017 already refused in the other
direction.

**It is refused because the value is reachable by placement.** ADR 0017's rule
for pooling partitions, run the other way: `.by = region` with
`.grouping = rollup(store, channel)` gives every row its share of its region
through `share_of_total()`, with no join. What that costs is the levels above
the ancestor, and that residue is the recipe.

**The recipe owns two correctness points**, each shown failing:

- the join key carries each dimension's inclusion bit beside its value —
  without them, under `.margin_label = NULL`, East's region subtotal takes the
  source-`NA` store's revenue as its denominator and reports a share of 5.2375;
- `na_matches = "na"` is explicit because dbplyr's default is `"never"` while
  `dplyr::left_join()`'s is `"na"` — without it the same pipeline is right
  locally and drops the unnamed store's share on a database. The DuckDB chunk
  shows the fix holding there.

No gate is added: the section has no `must_error` chunk to write a
`verify-site.R` marker from, and the numbers are a caller's pipeline rather
than a package contract, so the chunks executing under `altdoc.yaml` is the
check. No `NEWS.md` entry, there being no API change.

## Checks

`lintr::lint_package()` and `spelling::spell_check_package()` clean,
`roxygen2::roxygenise()` produces no diff, `quarto render` of the guide
succeeds over all 75 chunks with DuckDB present, and `testthat::test_local()`
is 6263 passing / 0 failing / 0 skipped with `MARGINPLYR_REQUIRED_SUGGESTS=all`.

### DIFF
--- a/R/share.R
+++ b/R/share.R
@@ -39,6 +39,14 @@
 #' empty [grouping_set()], and a specification without one is rejected naming
 #' that fix.
 #'
+#' Neither helper reaches a level between the two. For a denominator at a
+#' named ancestor level, move that dimension into `.by` and use
+#' `share_of_total()`; the [recipes guide][recipes] writes that out, and the
+#' join for the case where the levels above the ancestor have to stay in the
+#' same result.
+#'
+#' [recipes]: https://sayuks.github.io/marginplyr/vignettes/recipes.html
+#'
 #' @section Direct shares:
 #' A direct call must be the complete right-hand side of an explicitly named
 #' summary. Its argument must be the bare name of one eligible ordinary summary
--- a/design/adr/0010-compute-parent-shares-as-a-contextual-summary.md
+++ b/design/adr/0010-compute-parent-shares-as-a-contextual-summary.md
@@ -473,6 +473,41 @@ why it does not reach a denominator that is not selected: `share_of_total()`
 accepts any plan containing a Grand total set, while this restriction on
 `share_of_parent()` still stands.
 
+A denominator at a *named ancestor level* — what fraction of its region a
+channel row is, in `rollup(region, store, channel)` — was rejected, and that
+deferral does not reach it either. ADR 0017's route is that a Grand total set
+is not selected at all; this one is selected, by the caller, and a rollup's
+parent chain is a total order, so the set two levels up is as unambiguous as
+the immediate parent. There is nothing for a selection model to decide, and
+refusing an ancestor denominator on parent ambiguity would copy a restriction
+for a reason that does not apply to it.
+
+It is refused because the value is already reachable, by the placement rule
+ADR 0017 states in the other direction. A caller who wants a cross-partition
+denominator moves that column out of `.by`; a caller who wants an ancestor
+denominator moves it in. `.by = region` with
+`.grouping = rollup(store, channel)` gives every row its share of its region
+through `share_of_total()`, with no join and no key of the caller's own.
+Placement is what chooses a denominator, and this is that lever run the other
+way rather than a workaround for a missing helper.
+
+What the placement costs is the rows above the ancestor: `region` becomes a
+fixed key, so no set of the plan pools regions and the result loses its
+all-regions row. A share against an ancestor level together with the rows
+above it is therefore the residue, and it is a documented join rather than an
+API: *Compare a row with an ancestor level* in the recipes guide owns the two
+things that join has to get right, which are the ones a caller does not
+discover — the ancestor level is selected by Grouping identity rather than by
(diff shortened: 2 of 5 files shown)


sayuks:
## Review record

(`2fab370...e7e3e01`), two axes in parallel. Standards sources: `AGENTS.md`
(*Documentation*, *Chunks that must fail*, *Site verification*, *Code
comments*, *Linting*), `design/architecture.md`'s *Answering a review*, and
`design/agents/code-review.md`. Spec input: #375 with its triage comment.
Generated files were outside the review.

Six findings. Four answered in `ba36fa9`, one rejected, one is the same
finding both axes raised.

|---|---|---|
| ADR 0010 misdescribed the recipe's "two things" — split one, dropped the other, asserted a count the named file contradicts | Standards (hard), Spec (a) | Fixed: cites the section, enumerates nothing |
| The new ADR paragraphs cited no ticket | Standards | Fixed: cites #375 |
| The guide re-derived the ADR's reason beside the citation | Standards | Fixed: citation alone |
| "dbplyr's default is `never`" was load-bearing and shown by nothing | Standards | Fixed: the same join without the argument is now a chunk — `1` in the first, `NA` in the second |
| The `grouping_bit()` block appears three times in the section | Standards (Fowler) | Rejected: the three differ by `.margin_label`, by backend, and by which produces the wrong number; folding them yields a chunk no reader copies |
| `na_matches` is not among the two details #375 names | Spec (b) | Rejected: the issue asks for "missing-safe keys", and the local/database divergence is what makes the key missing-safe |

The Spec axis verified the branch's central claim independently: `.by = region`
with `rollup(store, channel)` reproduces rows 1–16 of the issue's own output
row for row in `region`/`store`/`channel`/`revenue`, differing only in the
share column, which is now the region denominator the issue asked for. What it
loses is one row — the all-regions total — which is what the ADR and the guide
both say it loses. It also confirmed the `5.2375` figure, the parent-chain
claim about `rollup(region, store, product, channel)`, and that the triage
comment's note about a skipped row in the issue body reaches nothing the
branch decided.

PR #397, opened 2026-09-02
Name a way forward when a verb refuses a RecordBatchReader

### LINKED ISSUE #391: expand_with_margins() refuses a RecordBatchReader without naming a way forward

sayuks:
Closes #391.

`assert_lazy_table()` named the class it refuses and stopped. It now carries a
remedy bullet:

```
`.data` must not be an object of the following classes: `RecordBatchReader`.
i Convert it with `arrow::as_arrow_table()` first.
```

The guard is shared, so the bullet is read from `summarize_with_margins()`,
`expand_with_margins()`, and `inspect_grouping()`. `nest_with_margins()` never
reaches it and its diagnostic is unchanged.

`arrow::as_arrow_table()` and not `dplyr::collect()` was the decision the ticket
existed to take; the commit message holds the argument, and the triage comment
on #391 holds the measurements both routes were chosen between.

`test-utils.R` pins the bullet — the existing expectations match the main line,
which a refusal that lost its remedy still has.

Full suite green with `NOT_CRAN=true`, `lintr::lint_package()` clean,
`roxygen2::roxygenise()` a no-op.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -147,14 +147,23 @@ assert_margin_input <- function(x) {
 # alternatives to pick one of; a blacklist refuses every entry at once, so a
 # second name would be refused alongside the first and not instead of it.
 # Nothing renders today, the vector holding one name.
+#
+# `dplyr::collect()` is not named beside `arrow::as_arrow_table()`, under the
+# rule ADR 0012's amendment states. Both resolve the refusal for every verb
+# reaching this guard, but they leave the caller in different places -- one a
+# local tibble, the other the Arrow table the verb answers as an Arrow query --
+# and converting is the smaller of the two.
 assert_lazy_table <- function(x) {
   # Read only from the cli template below, which codetools cannot see.
   nm <- deparse(substitute(x)) # nolint: object_usage_linter.
   invalid_lazy_table_names <- "RecordBatchReader"
   if (inherits(x, invalid_lazy_table_names)) {
-    abort_marginplyr(paste0(
-      "{.arg {nm}} must not be an object of the following classes: ",
-      "{.code {invalid_lazy_table_names}}."
+    abort_marginplyr(c(
+      paste0(
+        "{.arg {nm}} must not be an object of the following classes: ",
+        "{.code {invalid_lazy_table_names}}."
+      ),
+      i = "Convert it with {.fun arrow::as_arrow_table} first."
     ))
   }
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -53,10 +53,10 @@ assert_string_scalar <- function(x) {
 # `toString()` comma both replace was neutral between the two readings.
 #
 # `expand_with_margins()` is not named beside `dplyr::collect()`, under the
-# rule ADR 0012's amendment states. It takes the same grouping specification
-# and stays lazy, but it returns rows rather than the list column this call
-# asked for, and `assert_lazy_table()` refuses a `RecordBatchReader` that
-# reaches this refusal.
+# rule ADR 0012's second amendment states. It takes the same grouping
+# specification and stays lazy, but it returns rows rather than the list column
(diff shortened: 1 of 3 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` over `git diff origin/main...HEAD` at `e55571c`,
two axes in parallel. Standards sources passed: `AGENTS.md` (*Code comments*,
*Linting*, *Documentation*, *Optional-dependency guards*, *Release matrix*,
*Dependency metadata*), `design/architecture.md`'s *Answering a review* for what
a finding about prose may be, and ADRs 0012, 0015, 0020, 0023, 0024. Spec: the
agent brief on #391.

Two findings, one on each axis, on the same hunk — both about the header
sentence, reached independently.

|---|---|---|
| the header's ADR 0012 citation is false: the rule it names separates a remedy from a non-remedy, and both routes here are remedies | Standards | fixed in prose, `2f2f942` |
| the header re-derives the decision's rationale rather than saying which way the site falls | Spec | same fix |

The Standards finding is the substantive one. What excluded `dplyr::collect()`
was a criterion nothing in the repository held, so the answer is ADR 0012's
third amendment rather than a rewording — the criterion now has a home, and
#374's has one too. Its *Considered options* holds what was weighed. The
amendment also separates `assert_margin_input()`'s `or` bullet, which looks like
a counter-example and is not: no object reaching that refusal has both routes,
measured rather than asserted.

`assert_nest_possible()`'s bare "ADR 0012's amendment" becomes the *second*
amendment, since adding a third is what made it ambiguous.

The brief's *Out of scope* said no ADR was owed. That rested on the reasoning
this round overturned, and is superseded.

**Non-findings recorded so a second round does not re-raise them.** ADR 0023 —
`{.fun arrow::as_arrow_table}` matches the style table, and the bullet
interpolates nothing, so the structural gate's refusal of a computed template is
respected; that is also why the near-identical `Convert it with ... first.` at
three sites in the file is not extractable duplication. *Dependency metadata* —
arrow is named only inside a message string, with no call site, is already in
`optional_suggest_spec()`, and the new expectation sits under the existing
`skip_if_suggest_absent("arrow")`. No baseline smell landed.

The Spec axis executed all six acceptance criteria rather than reading them: the
bullet on all three verbs, the `as_arrow_table()` route through each, the
nesting diagnostic unchanged, the pin failing on exactly one expectation when
only the bullet is removed, the condition class, and lint. No scope creep, no
requirement missing.

Full suite green with `NOT_CRAN=true` at `2f2f942`; `lintr::lint_package()`
clean; `roxygen2::roxygenise()` a no-op.

PR #398, opened 2026-09-02
Delete the vignette prose restating a diagnostic printed below it (#392)

### LINKED ISSUE #392: Vignette prose describing a diagnostic printed directly below it

sayuks:
Closes #392.

#374 deleted one clause of this shape from `database_backends.qmd` and left
open whether that was one editorial call or a form the vignettes repeat. It is
a form. All 19 `must_error` chunks under `vignettes/` were scanned with the
paragraph above each one; five carried a sentence asserting what the refusal
below it contains, and three of those are bare restatement:

| vignette | the clause |
| --- | --- |
| `database_backends.qmd` | `and tells you the two ways to get it computed` |
| `get_started.qmd` | `and names the positions that collide` |
| `get_started.qmd` | `is rejected naming that fix` |

Each sentence keeps the argument it was making and loses only the promise about
the diagnostic's content. The Arrow one also asserted a count, which is the
failure #374 already met once — the clause it deleted had gone incomplete when
the refusal below it gained a remedy bullet.

The two remaining clauses stay: `not a silently different answer` and `instead
of choosing a parent` say what the diagnostic does not, so a reader learns from
them something the rendered refusal never tells them.

`design/architecture.md`'s *Answering a review* gains the rule as a fifth kind
a prose finding may be, naming the contrast case as what it does not reach.
Not `AGENTS.md`: its always-loaded closure is at 24198 of a 24198-byte
baseline, and the reasons its *Code comments* section gives for auditing by
hand are `R/`-specific and do not transfer to a vignette.

No marker moves, though one was close — `verify-site.R` pins the backends
guide's absorbed-summary section through the prose `refuses it before a row is
read`, since the chunk is behind `has_arrow`, and this edit shortens the
sentence around that run while leaving it whole.

Checks: `lintr::lint_package()` clean; full suite 6263 passing with no failures
and no skips under `MARGINPLYR_REQUIRED_SUGGESTS=all`; `verify-must-error.R`,
`verify-verifier-invocation.R`, and `verify-context-budget.R` pass with the
baseline unmoved; the site renders and `verify-site.R` verifies it.

### DIFF
--- a/design/architecture.md
+++ b/design/architecture.md
@@ -711,11 +711,16 @@ rejection with evidence. Naming which one is what the disposition does, and a
 comment is not among them: a finding a comment would answer is one of the
 others.
 
-A finding about prose is one of four: the prose is false, it duplicates an
-argument another file owns and can drift from it, it breaks a rule a repository
-document states, or something a repository document is required to hold is
-missing from it. How prose reads is not one, and no record holds a finding that
-is.
+A finding about prose is one of five: the prose is false, it duplicates an
+argument another file owns and can drift from it, it restates what a diagnostic
+rendered beside it says, it breaks a rule a repository document states, or
+something a repository document is required to hold is missing from it. How
+prose reads is not one, and no record holds a finding that is.
+
+The third does not reach a sentence saying what that diagnostic does not say —
+a contrast, or the answer the caller would otherwise have got — which is the
+page's own argument and not a copy of anything. #392 measured the `must_error`
+chunks under `vignettes/` and found both forms.
 
 Two answers join the list above when a finding is about prose. A claim found
 false is deleted rather than restated, as *Code comments* in `AGENTS.md`
--- a/design/architecture.md
+++ b/design/architecture.md
@@ -717,10 +717,10 @@ rendered beside it says, it breaks a rule a repository document states, or
 something a repository document is required to hold is missing from it. How
 prose reads is not one, and no record holds a finding that is.
 
-The third does not reach a sentence saying what that diagnostic does not say —
-a contrast, or the answer the caller would otherwise have got — which is the
-page's own argument and not a copy of anything. #392 measured the `must_error`
-chunks under `vignettes/` and found both forms.
+The restatement case does not reach a sentence saying what the diagnostic does
+not say — a contrast, or the answer the caller would otherwise have got — which
+is the page's own argument and not a copy of anything. #392 measured the
+`must_error` chunks under `vignettes/` and found both forms.
 
 Two answers join the list above when a finding is about prose. A claim found
 false is deleted rather than restated, as *Code comments* in `AGENTS.md`
--- a/vignettes/database_backends.qmd
+++ b/vignettes/database_backends.qmd
@@ -510,8 +510,7 @@ reasons.
 
 The first is a summary Arrow's own engine cannot evaluate. Arrow answers one of
 those by reading the whole input — every column, not just the ones the summary
-names — and computing it in R, so marginplyr refuses it before a row is read
-and tells you the two ways to get it computed:
+names — and computing it in R, so marginplyr refuses it before a row is read:
 
 ```{r}
 #| label: arrow-absorbed-summary
--- a/vignettes/database_backends.qmd
+++ b/vignettes/database_backends.qmd
@@ -524,11 +524,10 @@ arrow::arrow_table(january_sales) |>
   )
 ```
 
-Collecting first is one rewrite; selecting the columns the summary needs before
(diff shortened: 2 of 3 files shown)


sayuks:
## Review record

with `main` (`git diff origin/main...HEAD`, fixed point `0efdc4a`) over
`4556be1`. Note the fixed point: local `main` was four merges behind
`origin/main`, and `design/agents/code-review.md` warns that reviewing against
a `main` the branch has fallen behind reports that gap as the branch's.

Standards raised three judgement calls and no hard violation. Spec found no
missing or partial requirement, and raised one scope question and one
fragility. Four dispositions:

| finding | disposition |
|---|---|
| The scan read the paragraph *above* each `must_error` chunk while the rule says *beside*, and the gap held a site: the paragraph *below* the Arrow absorbed-summary chunk enumerated both remedy bullets of `abort_absorbed_summary()` and asserted the count | fixed in prose — `c307b7f` |
| `The third` indexed by position into an unnumbered inline list one paragraph above (raised by both axes) | fixed in prose — `c307b7f` |
| `names the rewrite instead of choosing a parent` is a hybrid whose first half restates | rejected — `c307b7f`'s message states why: the rule exempts a sentence saying what the diagnostic does not say, and the refusal names no parent and offers no choice |
| the `#392` citation closing the new paragraph is provenance belonging in a commit message | rejected — `design/architecture.md` cites tickets this way throughout, most nearly at its *Repository placement* and *Answering a review* sections |

The first is the one worth naming here, because it found the branch under-applying
the rule it was writing. `c307b7f` answers it with the edit rather than by
narrowing "beside" to "above", and re-ran the scan in both directions over all
19 chunks; the Arrow site is the only one the second direction added.

Spec independently re-extracted all 19 chunks and confirmed no fourth
bare-restatement site was missed, that the two contrast clauses are
byte-identical to `origin/main`, that the rule is stated in exactly one place
repo-wide, and that the *Out of scope* list holds — no `AGENTS.md` change, no
baseline move, no ADR, no change to any diagnostic, marker, or the `must_error`
hook.

Re-checked after `c307b7f`: `lintr::lint_package()` clean,
`test-documentation.R` 21 passing, `verify-context-budget.R` at 24198 of 24198
with the baseline unmoved, and the site re-rendered with `verify-site.R`
verifying it.

PR #403, opened 2026-09-02
Catalog explain and remote_query_plan as execution entry points

### LINKED ISSUE #399: Catalog dplyr::explain and dbplyr::remote_query_plan as execution entry points
## Parent

#318

## What to build

`lazy_execution_entry_points()` catalogs `dplyr::explain` and
`dbplyr::remote_query_plan`, both with `subject_test = FALSE`.

This is what makes #382's rejection structural rather than a note in an ADR.
Both functions reach `DBI::dbGetQuery()` from inside dbplyr, where the static
scan over `R/` cannot follow, so one line of either in `R/` would leave the
gate silent — reporting a clean result, which reads exactly like a package
that calls neither.

#318 lists this as independent of the record it specifies, and it is: the
capture that feature uses is `dbplyr::sql_render()`, which stays absent from
this catalog along with `show_query()`. Nothing here blocks or is blocked by
the rest of #318.

Neither entry takes a positive control, and a comment says why: a positive
control for "marginplyr does not call this" would be marginplyr calling it.

## Acceptance criteria

- [ ] `lazy_execution_entry_points()` holds both new rows with `subject_test = FALSE`
- [ ] The snapshot of the scanned entry-point set moves by exactly two rows
- [ ] The snapshot of internal functions reaching an entry point does not move
- [ ] `traced_execution_entry_points()` needs no edit — it derives from the same catalog
- [ ] A comment records why neither entry takes a positive control
- [ ] `sql_render()` and `show_query()` remain absent from the catalog
- [ ] Local gates pass: `lintr::lint_package()` after `pkgload::load_all()`, the suite with `NOT_CRAN=true`, and `Rscript .github/scripts/verify-suite-coverage.R`

## Blocked by

None — can start immediately.

sayuks:
Closes #399.

`lazy_execution_entry_points()` in `tests/testthat/test-query-policy.R` now
holds `dplyr::explain` and `dbplyr::remote_query_plan`, both with
`subject_test = FALSE`. Each reaches `DBI::dbGetQuery()` from inside dbplyr,
where the walk over `R/` cannot follow, so one line of either there would
leave the gate reporting a clean result — which reads exactly like a package
calling neither. That is what makes ADR 0027's rejection structural rather
than prose.

`traced_execution_entry_points()` derives from the same catalog and needed no
edit. The snapshot of the scanned entry-point set moves by exactly these two
rows; the snapshot of internal functions reaching an entry point does not
move, which is the assertion. `sql_render()` and `show_query()` stay absent.

A comment records why neither entry takes a positive control: what they assert
is that marginplyr does not call them, so a control would be marginplyr
calling `explain()`.

## Local gates

- Full suite with `NOT_CRAN=true` — exit 0
- `Rscript .github/scripts/verify-suite-coverage.R` — 707 tests, all covered

### DIFF
--- a/tests/testthat/_snaps/query-policy.md
+++ b/tests/testthat/_snaps/query-policy.md
@@ -7,14 +7,14 @@
       1    dplyr           collect        FALSE
       2    dplyr           compute        FALSE
       3    dplyr              pull        FALSE
-      4     base     as.data.frame         TRUE
-      5   tibble         as_tibble         TRUE
-      6      DBI        dbGetQuery        FALSE
-      7      DBI       dbSendQuery        FALSE
-      8      DBI   dbSendStatement        FALSE
-      9      DBI           dbFetch        FALSE
-      10     DBI       dbReadTable        FALSE
-      11   dplyr           explain        FALSE
+      4    dplyr           explain        FALSE
+      5     base     as.data.frame         TRUE
+      6   tibble         as_tibble         TRUE
+      7      DBI        dbGetQuery        FALSE
+      8      DBI       dbSendQuery        FALSE
+      9      DBI   dbSendStatement        FALSE
+      10     DBI           dbFetch        FALSE
+      11     DBI       dbReadTable        FALSE
       12  dbplyr remote_query_plan        FALSE
 
 # marginplyr functions reaching an execution entry point
--- a/tests/testthat/_snaps/query-policy.md
+++ b/tests/testthat/_snaps/query-policy.md
@@ -3,17 +3,19 @@
     Code
       lazy_execution_entry_points()
     Output
-         package            name subject_test
-      1    dplyr         collect        FALSE
-      2    dplyr         compute        FALSE
-      3    dplyr            pull        FALSE
-      4     base   as.data.frame         TRUE
-      5   tibble       as_tibble         TRUE
-      6      DBI      dbGetQuery        FALSE
-      7      DBI     dbSendQuery        FALSE
-      8      DBI dbSendStatement        FALSE
-      9      DBI         dbFetch        FALSE
-      10     DBI     dbReadTable        FALSE
+         package              name subject_test
+      1    dplyr           collect        FALSE
+      2    dplyr           compute        FALSE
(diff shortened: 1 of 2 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, both axes, against `git diff origin/main...HEAD` at `0cfb840` (merge-base `9e382a0`). Standards sources: `AGENTS.md` *Code comments*, *Queries against a lazy input*, *Linting*, *Release matrix*, with `design/architecture.md` *Answering a review* governing what a finding about prose may be. Spec sources: #399 and ADR 0027 *Test strategy*.

Four findings. Three answered in `9dec068`, one rejected.

|---|---|---|
| The comment restates ADR 0027's *Test strategy* beside the citation to it — the second copy *Code comments* refuses | both | fixed in prose |
| "They take no subject test for the reason the `DBI::` entries take none" is false: that reason ends in the first argument being a connection, and both new entries take the caller's lazy object there | both | fixed in prose |
| The two rows append after the `DBI::` block, splitting `dplyr::explain` from the file's by-package grouping | Standards | fixed in prose |
| The comment omits ADR 0027's sentence on what the existing entries' controls discharge | Spec | rejected |

The rejection: both acceptance criteria the comment answers are met without that sentence, and adding it is the first finding run backwards — more of the ADR's argument copied to a site the citation already reaches.

Spec axis confirmed the rest of #399's criteria against the consumers rather than the catalog alone: `traced_execution_entry_points()` is unedited and handles both rows (dplyr and dbplyr are Imports, so its reachability guard holds, and `subject_test = FALSE` skips the formals branch); the reach snapshot is untouched; `sql_render()` and `show_query()` stay absent. It also confirmed the reachability claim the membership rests on: `explain.tbl_sql` → `remote_query_plan` → `dbplyr_explain` → `db_get_query` → `DBI::dbGetQuery`.

Gates re-run after the answers, all green: `lintr::lint_package()` after `pkgload::load_all()` — no lints; full suite with `NOT_CRAN=true` — exit 0; `Rscript .github/scripts/verify-suite-coverage.R` — 707 tests, all covered.

Issue #402, opened 2026-09-02
Document the Sent query record in ?marginplyr and the database-backends vignette


sayuks:
## Parent

#318

## What to build

The two reader-facing homes for the feature, written once against a settled
implementation. #318's step 6.

**`?marginplyr`** gains a section after *Errors and warnings*. It is **not**
called *Audit signal* — there is no signal. `@section Recording the SQL
marginplyr sends:` matches the vocabulary and the ADR 0027 title. It states: off
by default and how to switch it on; that `last_sent_queries()` is how a caller
reads it; the two columns; that only `"result"` is promised; that a query is
recorded before it is sent and so survives a failing call; the four answers; the
SQL-not-execution boundary; and that there is one capture, a client-side render
that sends nothing the call did not already send. A short example mirroring the
existing error example's register.

**`vignettes/database_backends.qmd`** gains a section after *Other lazy
backends* and before *Backend verification levels*, so that it follows the
backends it is about. It leads with `last_sent_queries()` against a live DuckDB
or RSQLite call, and states plainly that the `"result"` row is a render and not
an execution — the caller's own `collect()` is still what runs it. It carries the
four answers briefly and the write-out recipe; **a reader skimming only the
vignette should not have to reach the Rd to learn that marginplyr writes no
file.**

`@section When marginplyr queries your data:` is **unchanged**. Recording sends
nothing new, so what that section enumerates is still exactly what reaches the
connection unasked.

Blocked by #401 as well as #400 because the vignette's DuckDB example renders
both a `"result"` and a `"selection_proxy"` row; rendered before #401 lands it
would bake in output the finished package does not produce.

## Acceptance criteria

- [ ] `?marginplyr` carries the section, titled `Recording the SQL marginplyr sends`, placed after *Errors and warnings*
- [ ] It states every item listed above, including that the capture sends nothing the call did not already send
- [ ] The vignette section sits between *Other lazy backends* and *Backend verification levels*
- [ ] It runs `last_sent_queries()` against a live lazy backend and shows real output
- [ ] It says the `"result"` row is a render, not an execution
- [ ] It says marginplyr writes no file, without the reader having to reach the Rd
- [ ] `@section When marginplyr queries your data:` is unchanged
- [ ] No roxygen line carries a `# nolint` outside a markdown table row; a long URL becomes a reference-style link
- [ ] `roxygen2::roxygenise()` and `spelling::spell_check_package()` are clean, and `man/` and `NAMESPACE` are committed as generated
[...]

Issue #401, opened 2026-09-02
Record the four Sent queries marginplyr executes


sayuks:
## Parent

#318

## What to build

The record gains the four queries marginplyr actually sends to a connection, so
an audited call reports every one of them and not only the result it renders.
#318's step 4, minus the `"result"` row that #400 lands.

Four purposes at three sites:

- `"selection_proxy"` — the zero-row proxy, in the `collect_selection_proxy`
  branch only. The arrow branch reads `arrow::schema()` and the fall-through
  branch sends nothing.
- `"observed_label_collision"` — the label scan under `.check_margin_label`.
- `"share_dialect"` and `"share_dialect_control"` — the dialect probe's question
  and its control. The control is a second call to the same function, so that
  function takes the purpose as an argument and its caller passes the two values.
  The row is written after the `tryCatch` that builds the query: a query that
  could not be built was never sent, so it takes no row.

**Every row is written before the `collect()` it describes.** That is what makes
the record hold what was sent rather than what succeeded: the probe's own
`tryCatch` swallowing a failure, an unrelated backend error, and
`abort_margin_label_collision()` firing on a real collision all leave the row
already written.

`nest_by_with_margins()`'s two `collect()` calls are excluded **and no code says
so**. `assert_nest_possible()` restricts input before nesting runs, so a SQL
`tbl_lazy` never arrives and the `is_sql` gate already withholds those calls.
This is a reachability fact, not a special case to write. The Arrow schema read
takes no row for the same reason.

What the share-eligibility probe and the observed-label-collision check *do* is
unchanged. This ticket only records the SQL they already send.
`validate_margin_label()` and `check_observed_label_collision()` gain no
parameter about backends, which is not what either answers.

## Acceptance criteria

- [ ] A DuckDB call records a `"selection_proxy"` row; an Arrow input and a backend without `collect_selection_proxy` record none
- [ ] `.check_margin_label = TRUE` against a lazy input records an `"observed_label_collision"` row
- [ ] A Parent share on a SQL backend records a `"share_dialect"` row, with `share_dialect_verdicts` emptied first the way `test-share-backends.R` already does
- [ ] The dialect probe takes its purpose as an argument, and its caller passes the question and control values
- [ ] Every row is written before the `collect()` it describes, and after the `tryCatch` that builds the probe's query
[...]

Issue #318, opened 2026-08-29
Record the SQL marginplyr sends, and read it back with last_sent_queries()


sayuks:
## Status

The specification is settled. Every verdict below arrives from a closed ticket
on map #377, and the decision is recorded in
[ADR 0027](https://github.com/sayuks/marginplyr/blob/main/design/adr/0027-record-the-sql-marginplyr-sends.md).
Nothing here is open for the implementer to decide; what this issue holds is
what to build. Where a sentence below states a reason, ADR 0027 is where the
argument for it lives — this issue does not re-derive it.

The feature this issue originally proposed was four features. One survived, and
the other three are rejected in ADR 0027's *Considered options*: the
`marginplyr_query_sql` condition (#381), the `"explain"` capture method (#382),
and the on-disk log (#384). An `is_sql_lazy_query()` predicate is rejected with
them (#386). **What is not built** below lists them so that a reader of this
issue does not have to reconstruct which half of an older plan is live.

## Context

Corporate audit policy requires the SQL used to produce collected data to be
recorded. Nothing available today records it distinguishably:
`dbplyr::last_sql()` holds one slot, overwritten by every render that funnels
through `db_sql_render()`, marginplyr's internal ones included, so after a
Margin verb it reports the selection proxy on DuckDB, the label scan under
`.check_margin_label`, or on RSQLite a stale query about an unrelated table —
with nothing saying which (#381).

**Sent query** is the term (`CONTEXT.md`). A Margin verb applied to a lazy input
returns its result unexecuted, so the query that matters most to an auditor is
one marginplyr never runs; it is captured by rendering, which sends nothing.

### 1. The option

`marginplyr.audit_sql`, off unless the caller sets it. This is the package's
first option: `R/` and `tests/` contain no `getOption("marginplyr` and no
`options(marginplyr` today.

It is read **exactly once per call**, at the reset in step 2, as
`isTRUE(getOption("marginplyr.audit_sql"))`, and the answer is remembered with
that call's record. Reading it at the recording sites instead would let a caller
who sets the option *after* the call read zero rows under a live option.

Any value that is not `TRUE` means "not audited", and the recording side reports
nothing about it: a garbage option value must not turn a caller's
`summarize_with_margins()` into an error. `last_sent_queries()` is the surface
that reports it, by refusing (step 5). `CONTEXT.md`'s *Option argument* entry
does not reach this option — it governs a fixed set of strings, and this is a
logical.

### 2. The record

[...]

PR #404, opened 2026-09-02
Record the Sent queries of one call, read back with last_sent_queries()

### LINKED ISSUE #400: Record the Sent queries of one call, read back with last_sent_queries()
## Parent

#318

## What to build

A caller sets `options(marginplyr.audit_sql = TRUE)`, runs a Margin verb against
a SQL backend, and reads the SQL marginplyr sent with `last_sent_queries()`.
The record belongs to one call and is emptied at the start of the next.

This is the tracer bullet through the whole feature: the option, the record, one
recorded site, and the accessor. #318's *What is built* steps 1, 2, 3, 5 and the
`"result"` row of step 4. The argument for every verdict lives in ADR 0027; this
ticket does not re-derive it.

**The option** is read exactly once per call, at the reset, and the answer is
remembered with that call's record — reading it at the recording sites would let
a caller who sets it *after* the call read zero rows under a live option. Any
value that is not `TRUE` means "not audited", and the recording side reports
nothing about it: a garbage option value must not turn a caller's
`summarize_with_margins()` into an error. `last_sent_queries()` is the surface
that reports it, by refusing.

**The reset site is `prepare_grouping_plan()`**, not `prepare_margin_operation()`.
`inspect_grouping()` reaches `prepare_grouping_plan()` directly and reaches the
selection proxy from inside it, so resetting in `prepare_margin_operation()`
appends `inspect_grouping()`'s query to the previous Margin operation's record
with nothing saying the record now spans two calls — the `dbplyr::last_sql()`
defect this feature exists to remove, rebuilt inside the feature. What makes
`prepare_grouping_plan()` the site is that all five entry points reach it and it
runs before every recorded site on every path.

**`grouping_backend()` gains `is_sql`**, already computed as a local for the
`kind` cascade. Its header contract: whether this input is a SQL backend;
equivalent to `dialect` being non-`NULL`, but `dialect` is the field dispatch
reads and its nullity does not announce itself as the test. The record reads
this field and nothing else. `is_sql` is written into the record where the
backend is computed, inside the same function whose top holds the reset, so no
recorded site is ever reached with the backend unknown.

**`last_sent_queries()`** returns a `dplyr::tibble()` with columns `purpose` and
`sql`, both character, and no print method — ADR 0013 settled both. Four answers,
distinguished:

1. nothing recorded in this session — a Package condition;
2. the last recorded call was not audited — a Package condition naming `marginplyr.audit_sql`;
3. audited and sent nothing — a zero-row tibble, the only zero-row case, and a correct answer;
4. a statement had no SQL form on this backend — a row with `sql = NA_character_`.

Both refusals are Package conditions under ADR 0015: the caller avoids each by
rewriting a call they wrote.

Row order is the order the queries were sent, and it is the only ordering the
record carries. There is no `time` column and no backend column.

## Acceptance criteria

- [ ] `marginplyr.audit_sql` is off unless the caller sets it, and is read exactly once per call at the reset
- [ ] A value that is not `TRUE` means "not audited" and raises nothing at the recording sites
- [ ] The record is a package environment emptied at the top of `prepare_grouping_plan()`
- [ ] `grouping_backend()` gains `is_sql` with a header contract, computed from the existing local
- [ ] `record_sent_query(purpose, query)` takes no backend argument and no connection; it is a no-op unless the remembered option flag and the remembered `is_sql` are both `TRUE`
- [ ] It renders with `dbplyr::sql_render()`, and a render that raises appends `sql = NA_character_` rather than dropping the row or propagating the error
- [ ] The `"result"` row records the value `order_margin_result()` returns, before it is returned to the caller
- [ ] `last_sent_queries()` is exported, returns a two-column `dplyr::tibble()`, and has no print method
- [ ] Its Rd carries: the SQL-not-execution promise; what `NA` in `sql` means; that only `"result"` is a promised `purpose`; that a query is recorded before it is sent; that two identical calls need not produce identical records, the dialect verdict being cached per dialect for the session; concurrency in one sentence (a package environment is invisible across `fork`, PSOCK and callr, so a read after a parallel run reports the last serial call); and the write-out recipe for one call and for a batch
- [ ] Its example is guarded with `marginplyr_suggest_available()`, not `\dontrun{}`, and restores the option
- [ ] The Rd shows `writeLines(x$sql)` for reading a multi-line statement
- [ ] Tests: option off on every backend records nothing and `last_sent_queries()` refuses naming the option (the primary regression guard)
- [ ] Tests: a local data frame and a dtplyr input, option on, each give answer 3 — the SQL-not-execution boundary asserted rather than only documented
- [ ] Tests: a SQL backend gives a `"result"` row whose `sql` is matched with `grepl()`, never exact-matched
- [ ] Tests: two sequential calls — the second replaces the first's rows rather than appending
- [ ] Tests: `inspect_grouping()` after a Margin operation — the record holds only `inspect_grouping()`'s own rows. This is the direct test of the reset placement
- [ ] Tests: a summary whose translation dbplyr refuses at render gives one row with `sql = NA_character_`, and the call itself is unaffected. The expression must build and raise only at render, and must not add a dependency the suite lacks
- [ ] Tests: answer 1, asserted without making test order load-bearing; if that is not achievable, assert the branch against an emptied environment and say so in a comment
- [ ] Tests: the result is unaffected — still lazy, and collects to the values the existing backend tests assert
- [ ] The option is restored with `on.exit(options(old), add = TRUE)`; the suite uses no `withr`
- [ ] Optional backends are reached through `skip_if_backend_absent()` / `backend_available()`, and no test requires more than one member of `optional_backends()`
- [ ] The snapshot of internal functions reaching an execution entry point does not move
- [ ] Local gates pass: lint, `roxygen2::roxygenise()`, `spelling::spell_check_package()`, the suite with `NOT_CRAN=true`, `verify-suite-coverage.R`, and `R CMD check` on a built tarball

## Blocked by

None — can start immediately.

sayuks:
Closes #400.

The tracer bullet through ADR 0027: the `marginplyr.audit_sql` option, the per-call record emptied at the top of `prepare_grouping_plan()`, `grouping_backend()$is_sql`, the `"result"` site in `finalize_margin_operation()`, and the exported `last_sent_queries()` with its Rd.

Out of scope here, per the ticket: the four internal sites (#401) and the `?marginplyr` section plus the vignette section (#402).

Local gates run: lint, `roxygenise()`, `spell_check_package()`, the suite with `NOT_CRAN=true`, `verify-suite-coverage.R`, `verify-context-budget.R`, and `R CMD check` on a built tarball (one NOTE, the worktree's `.git` file).


sayuks:
## Review record

Reviewed `9c93315...158bf52` through `mattpocock-skills:code-review` against #400 and the standards `AGENTS.md` (*Code comments*, *Linting*, *Release matrix*, *Queries against a lazy input*), `design/architecture.md` (*Answering a review*), ADR 0027, 0015, 0023, 0013. Answered in d85b850.

**Standards** — 3 hard, 4 judgement.
- Rd, NEWS, and the `grouping-plan.R` comment claim rows the code does not yet record → NEWS and the comment now state what this commit records; the Rd sentences stay, because #400's Acceptance names each of them and #401 adds the sites they describe. The vignette is #402.
- Headers of `reset_sent_queries()`, `record_sent_query()`, and `grouping_backend()` re-derived ADR 0027's argument → the re-derivations are deleted; `grouping_backend()`'s keeps the wording ADR 0027 assigns to #318 and loses the sentence about another file. The test comment is cut to the assertion it names.
- `sqlite_sent_queries_table()` used on DuckDB → renamed `sent_queries_table()`. `purpose`/`sql` grown as two vectors → one tibble. Six RSQLite connections → left as they are; each test opening and closing its own connection is the suite's idiom (`test-branch-union.R`, `test-factor.R`). `remember_sent_query_backend()` as speculative → rejected; #400's Acceptance requires `is_sql` to be remembered where the backend is computed.

**Spec** — 0 missing, 2 minor.
- `expect_identical(record$sql, sql_render(query))` against "never exact-matched" → deleted.
- "collects to the values the existing backend tests assert" asserted as audited-vs-unaudited `collect()` equality → kept; the equality is the property the criterion guards.

PR #405, opened 2026-09-02
Record the queries the four internal sites send

### LINKED ISSUE #401: Record the four Sent queries marginplyr executes

sayuks:
Closes #401.

The three recording sites #400 left, with the `"result"` row already landed:

- `"selection_proxy"` — the zero-row proxy, in the `collect_selection_proxy`
  branch only. The arrow branch reads `arrow::schema()` and the fall-through
  branch sends nothing, so neither takes a row.
- `"observed_label_collision"` — the label scan under `.check_margin_label`.
- `"share_dialect"` and `"share_dialect_control"` — the dialect probe's
  question and its control. The probe is one function called twice, so it
  takes its purpose as an argument and `probe_share_dialect()` passes the two
  values. Its row is written after the `tryCatch` that builds the query: a
  query that could not be built was never sent.

Every row is written before the `collect()` it describes, which is what makes
the record hold what was sent rather than what succeeded. The test of that is
a share the eligible-type rule refuses: the probe's row is readable after the
`marginplyr_error`.

`nest_by_with_margins()`'s two `collect()` calls take no row and no code says
so — `assert_nest_possible()` restricts input before nesting runs, so the
`is_sql` gate already withholds them.

The dialect probe's two rows get no runtime test that they are SQL. The
snapshot of internal functions reaching an execution entry point does not move.

## Gates

`lintr::lint_package()`, `roxygen2::roxygenise()` (no diff),
`spelling::spell_check_package()`, the suite with `NOT_CRAN=true`,
`Rscript .github/scripts/verify-suite-coverage.R`, and `R CMD check` on a built
tarball all pass.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -298,9 +298,8 @@
   than answered with zero rows, and a statement dbplyr cannot translate for
   the backend is recorded as `NA` without failing the call. The record also
   holds the queries marginplyr sends for its own reasons on the way: the
-  zero-row selection proxy it reads a lazy input's columns from, which is what
-  an audited `inspect_grouping()` records; the scan under
-  `.check_margin_label` looking for an observed value equal to a Margin
+  zero-row selection proxy it reads a lazy input's columns from; the scan
+  under `.check_margin_label` looking for an observed value equal to a Margin
   label; and the question it asks a dialect before calculating a contextual
   share, with the control that tells a refusal from a question that could
   not be put. Each is recorded before it is sent, so a call refused by a
--- a/NEWS.md
+++ b/NEWS.md
@@ -296,5 +296,13 @@
   a data frame or a `dtplyr` table records nothing. Reading before any call
   has run, or after a call made while the option was off, is refused rather
   than answered with zero rows, and a statement dbplyr cannot translate for
-  the backend is recorded as `NA` without failing the call. The option is off
-  by default (#318, #400).
+  the backend is recorded as `NA` without failing the call. The record also
+  holds the queries marginplyr sends for its own reasons on the way: the
+  zero-row selection proxy it reads a lazy input's columns from, which is what
+  an audited `inspect_grouping()` records; the scan under
+  `.check_margin_label` looking for an observed value equal to a Margin
+  label; and the question it asks a dialect before calculating a contextual
+  share, with the control that tells a refusal from a question that could
+  not be put. Each is recorded before it is sent, so a call refused by a
+  collision or by an ineligible share source leaves readable every query it
+  had already sent. The option is off by default (#318, #400, #401).
--- a/R/backend-metadata.R
+++ b/R/backend-metadata.R
@@ -10,7 +10,9 @@ grouping_selection_proxy <- function(.data,
     return(as.data.frame(arrow::schema(.data)))
   }
   if (backend$collect_selection_proxy) {
-    return(dplyr::collect(utils::head(.data, n = 0L)))
+    proxy <- utils::head(.data, n = 0L)
+    record_sent_query("selection_proxy", proxy)
+    return(dplyr::collect(proxy))
   }
   .data
 }
--- a/R/margin-label.R
+++ b/R/margin-label.R
@@ -267,7 +267,9 @@ check_observed_label_collision <- function(data,
     }
   )
   names(checks) <- read_cols
-  found <- dplyr::collect(dplyr::summarize(data, !!!checks))
+  query <- dplyr::summarize(data, !!!checks)
+  record_sent_query("observed_label_collision", query)
+  found <- dplyr::collect(query)
   found <- vapply(
     read_cols,
     function(col) {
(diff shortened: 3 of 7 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` over `git diff d85b850...HEAD` — the merge-base with `main` — on both axes. Standards was passed AGENTS.md's *Code comments*, *Queries against a lazy input*, *Release matrix*, *Optional-dependency guards*, and *Linting*, plus `design/architecture.md`'s *Answering a review*; Spec was passed #401 with #318 and ADR 0027. Round one, at a8a092a; answers are a0b3b65.

Six findings. Four answered in code or prose, two rejected.

| finding | disposition |
|---|---|
| `test-sent-queries.R`'s header said the option was the only state leaking between its tests, which the two new share tests made false | fixed in prose (a0b3b65) |
| NEWS claimed the selection proxy is "what an audited `inspect_grouping()` records", which holds only where the backend grants `collect_selection_proxy` — and the suite's one audited `inspect_grouping()` runs on RSQLite, which does not | claim deleted (a0b3b65), as *Answering a review* requires of a claim found false |
| the pair emptying and restoring `share_dialect_verdicts` was copied into a second file, and named there for the file rather than for the cache | moved to `tests/testthat/helper-share-dialect-verdicts.R`; both files now reach one copy under its own name (a0b3b65) |
| the arrow test sat in the selection-proxy section, where it read as an assertion about the proxy's arrow branch that `record_sent_query()`'s `is_sql` gate reaches before | moved beside the dtplyr test, in the section for what an audited call on a non-SQL backend records (a0b3b65) |
| the `purpose` comment's second clause restated what `probe_share_dialect()` shows two screens up | cut (a0b3b65) |
| Primitive Obsession: the `purpose` vocabulary is bare strings at four sites with no shared list, where `share_dialect_verdict_names()` gets one | rejected. Every site's string is asserted verbatim by a test in `test-sent-queries.R`, so a typo fails the suite. The asymmetry `R/share.R`'s own comment draws is what the guard there is for: a wrong verdict is cached under the dialect and reused, where a purpose is a label written once into a record the caller reads |
| the vignette section ADR 0027 requires is still absent | rejected: it is #402, which this branch does not close |

## Gates

`lintr::lint_package()`, `roxygen2::roxygenise()` (no diff), `spelling::spell_check_package()`, the suite under `NOT_CRAN=true`, `verify-suite-coverage.R` (731 tests, all executing in at least one single-backend configuration), and `R CMD check` on a built tarball — all pass. The check's one NOTE is the `.git` a build from a worktree carries into the tarball.

PR #406, opened 2026-09-02
Document the Sent query record in ?marginplyr and the database-backends guide

### LINKED ISSUE #402: Document the Sent query record in ?marginplyr and the database-backends vignette

sayuks:
Closes #402. #318's step 6, written once against the implementation #400 and
#401 settled.

## What changed

**`?marginplyr`** gains `@section Recording the SQL marginplyr sends:` after
*Errors and warnings*. It is not called *Audit signal* — there is no signal,
and the title matches ADR 0027's. It states: the record is off until
`options(marginplyr.audit_sql = TRUE)`; `last_sent_queries()` is how a caller
reads it; the two character columns and their row order; that only `"result"`
is promised; that a row is written before its query is sent, so a failing call
leaves readable everything it had already sent; the SQL-not-execution boundary;
the four answers; and that the one capture is `dbplyr::sql_render()`, a
client-side render that sends nothing the call did not already send. A short
non-executed example in the register of the existing error example.

**`vignettes/database_backends.qmd`** gains *Record the SQL marginplyr sends*
between *Other lazy backends* and *Backend verification levels*, so that it
follows the backends it is about. It runs against a live DuckDB table and shows
the real `c("selection_proxy", "result")` record, renders the `"result"` SQL
with `writeLines()`, and then collects the query — which is the point stated
plainly beside it: that row is a render, not an execution, and the caller's own
`collect()` is still what runs it. The four answers are carried briefly, and
that **marginplyr writes no file** is said in the guide rather than only in the
Rd, with `utils::write.csv(last_sent_queries(), path)` beside it.

**`.github/scripts/verify-site.R`** gains the section heading as a marker for
that page. The section's chunks are all behind `has_duckdb`, so nothing they
produce can serve as one.

`@section When marginplyr queries your data:` is unchanged, which is the same
fact as the capture sentence above: recording sends nothing new, so what that
section enumerates still reaches the connection unasked and only that.

## Gates

- `roxygen2::roxygenise()` — `man/marginplyr-package.Rd` regenerated and committed
- `spelling::spell_check_package()` — no errors
- `Rscript .github/scripts/verify-context-budget.R` — within baseline, unmoved
- `Rscript .github/scripts/verify-verifier-invocation.R` — passes
- Full suite with `NOT_CRAN=true` — 6337 pass, 0 fail, 0 skip
- Site rendered from the working tree install; `Rscript .github/scripts/verify-site.R` passes over `docs/`
- `R CMD check` on a built tarball — one NOTE, the worktree's own `.git` file, environmental
- `README.md` not regenerated: no chunk of `README.Rmd` changed

### DIFF
--- a/.github/scripts/verify-site.R
+++ b/.github/scripts/verify-site.R
@@ -364,6 +364,9 @@ markers <- list(
     "Parent shares use a staged lazy query",
     "share_of_total",
     "Completion stays in the lazy input pipeline",
+    # The Sent-query section, by its heading rather than by anything it
+    # renders: every chunk in it is behind `has_duckdb`, so on a machine
+    # without DuckDB the section produces no output for a marker to quote.
     "Record the SQL marginplyr sends",
     "Only DuckDB and SQLite are exercised as live SQL databases",
     "DuckDB covers native SQL end to end",
--- a/.github/scripts/verify-site.R
+++ b/.github/scripts/verify-site.R
@@ -364,6 +364,7 @@ markers <- list(
     "Parent shares use a staged lazy query",
     "share_of_total",
     "Completion stays in the lazy input pipeline",
+    "Record the SQL marginplyr sends",
     "Only DuckDB and SQLite are exercised as live SQL databases",
     "DuckDB covers native SQL end to end",
     "dozen-plus fallback dialects",
--- a/R/marginplyr-package.R
+++ b/R/marginplyr-package.R
@@ -133,8 +133,9 @@
 #' [last_sent_queries()] gives four answers, and each of the three that could
 #' look like an empty record is told apart from the others:
 #'
-#' - Nothing has been recorded in this session -- no Margin verb has run --
-#'   and the call is refused with a `"marginplyr_error"`.
+#' - Nothing has been recorded in this session -- neither a Margin verb nor
+#'   [inspect_grouping()] has run -- and the call is refused with a
+#'   `"marginplyr_error"`.
 #' - The last call was not audited, the option being unset or holding a value
 #'   other than `TRUE` when the call began, and the call is refused with a
 #'   `"marginplyr_error"` naming `marginplyr.audit_sql`. The option is read
--- a/R/marginplyr-package.R
+++ b/R/marginplyr-package.R
@@ -87,6 +87,65 @@
 #' summary expressions run when you collect the result rather than while the
 #' verb runs, so what they raise is the collecting call's to report.
 #'
+#' @section Recording the SQL marginplyr sends:
+#' marginplyr keeps no record of the SQL it sends unless you ask for one.
+#' `options(marginplyr.audit_sql = TRUE)` switches the record on, and
+#' [last_sent_queries()] reads back what the most recent call sent:
+#'
+#' ```r
+#' # `sales_db` is a lazy table on a SQL connection.
+#' old <- options(marginplyr.audit_sql = TRUE)
+#'
+#' summarize_with_margins(
+#'   sales_db,
+#'   revenue = sum(revenue),
+#'   .grouping = rollup(region, store)
+#' )
+#'
+#' last_sent_queries()
+#' options(old)
+#' ```
+#'
+#' The record is a tibble of two character columns -- `purpose`, what the
+#' query was sent for, and `sql`, the statement as it was rendered -- with one
+#' row per query, in the order the queries were sent. It belongs to a single
+#' call, being emptied at the start of every one.
+#'
+#' `"result"` is the one `purpose` promised: the query a Margin verb returns
+#' unexecuted, which your own [dplyr::collect()] is still what runs. Every
+#' other row is a query marginplyr sends for its own reasons, and the names
+#' those rows carry are not fixed.
+#'
+#' A query is recorded before it is sent, so the record holds what was sent
+#' rather than what succeeded, and a call that fails leaves readable every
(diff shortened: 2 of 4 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel, over
`git diff origin/main...HEAD` at `85f7a15`. The fixed point is `origin/main`
and not the local `main`, which was two merges behind — reviewing against a
stale base reports that gap as the branch's (`design/agents/code-review.md`).

Standards sources passed: `AGENTS.md` (*Linting*, *Documentation*, *Site
verification*, *Code comments*, *Agent skills*), `CONTEXT.md`'s **Sent query**
entry, ADR 0027, and `design/architecture.md`'s *Answering a review* as what a
finding about prose may be. Spec: #402, #318 step 6, ADR 0027's *Documentation
consequences*. `man/marginplyr-package.Rd` was excluded as generated.

Four findings; both axes raised the first one independently.

| # | Finding | Disposition |
|---|---|---|
| 1 | Answer 1 is false as written in both new copies: `"no Margin verb has run"`, when `inspect_grouping()` reaches `reset_sent_queries()` too. The vignette's `"the next Margin verb empties it"` narrows the same way | Fixed in prose, `a8f64ef` |
| 2 | The vignette's gloss on `"selection_proxy"` is a second account of a query *When marginplyr queries your data* owns, and had already dropped its "which backends" qualifier | Fixed in prose, `a8f64ef` |
| 3 | The new `verify-site.R` marker is a heading, the weaker form, where every other weak marker in that vector carries a comment | Fixed in code, `a8f64ef` |
| 4 | `?marginplyr` omits the concurrency position and the cached-dialect caveat, which ADR 0027's *Documentation consequences* names | Rejected: #318 step 6 and #402's acceptance criteria both enumerate that section's contents and neither names them, and `last_sent_queries()`'s Rd carries both |

Finding 1 was measured before it was answered: with the option set, an
`inspect_grouping()` on a local frame and nothing else gives
`# A tibble: 0 × 2` from `last_sent_queries()`, not a refusal.

Both axes reported the marker addition as defensible but unasked-for. It is
kept — it is the only thing that fails when the section stops rendering, and
three entries already in that page's vector are headings — and finding 3 is
what records the choice at the site.

Everything else was checked and found clean: every other factual claim in the
new prose against `R/sent-queries.R`, `R/grouping-backend.R`,
`R/backend-metadata.R`, and ADR 0027; the `?marginplyr`/Rd overlap as required
by #318 step 6 rather than a second copy; `@section When marginplyr queries
your data:` untouched; no roxygen `# nolint`; no bare URL. No baseline smell
has a site in a documentation diff.

## Gates re-run after the answers

`roxygenise()`, `spell_check_package()`, `lint_package()`,
`verify-context-budget.R`, site rendered from a fresh install with
`verify-site.R` passing over `docs/`, and `R CMD check` on a rebuilt tarball —
one NOTE, this worktree's own `.git` file, environmental. `README.md`
unchanged.

Issue #411, opened 2026-09-03
Restate the caller's spelling in the native adapter's translation error


sayuks:
Implements the decision recorded on #410. ADR 0022 already carries the amendment
(*the native adapter's translation error is restated too*) and
`investigation/what-truncates-an-argument-label-on-the-native-adapter.md` holds
the measurements; what is left is the code and the tests.

## What a caller sees today

On duckdb or postgres, a summary expression dbplyr cannot translate:

```
ℹ In argument: `total = +...`
Caused by error:
! Object `no_such_column` not found.
```

The same call on a local input:

```
ℹ In argument: `total = sum(value) + grouping_bit(a) + no_such_column`.
ℹ In group 1: `a = "x"`.
Caused by error:
! object 'no_such_column' not found
```

The truncation is not length-driven: `rewrite_grouping_dots()` splices in a
`dbplyr` `sql` object whose own `deparse()` is 59 characters, against the width
of 60 at which `as_label()` deparses the whole expression, so anything spliced
beside it overflows and rlang replaces the call's arguments with `...`. Any dot
combining a helper with anything else labels as `+...`, and a dot that is the
helper alone labels as `<sql>`.

## Changes

**1. `restate_argument_bullet()` (`R/conditions.R:506-519`) — relax the trailing
period, and put it back as found.**

The pattern is `` "^([iℹ] )?In argument: `(.*)`\\.$" ``. dbplyr's bullet carries
no trailing period, so extraction returns `character(0)` and the message passes
through unchanged even where the map's key already equals the span. Capture the
period instead of requiring it, and rebuild with what was captured — rebuilding
with an unconditional `` `. `` would add a period to dbplyr's sentence, and ADR
0022 is explicit that only the span inside the backticks is replaced.

The greedy read to the last backtick is unchanged; its reason (a non-syntactic
name the caller wrote puts backticks inside the span) still stands.

**2. `summarize_margin_native()` (`R/grouping-adapter-native.R`) — restate the
error from its one `summarize()`.**

Take the caller's labels, and around the `dplyr::summarize()` at
`R/grouping-adapter-native.R:61-70` catch the error, apply
`restate_condition_arguments(cnd, branch_argument_map(dots, caller_labels))`, and
re-raise. `restate_condition_arguments()` (`R/conditions.R:412`) takes a condition
and a map and depends on no branch machinery, so nothing else is needed.

Not `with_branch_conditions()`: the native adapter issues one `summarize()` and
[...]

Issue #199, opened 2026-08-16
The argument context quotes marginplyr's rewrite rather than the caller's spelling


sayuks:
Split out of #141, which settles the Repeated-condition contract and fixes two
of the three parts of the context leak. This is the third part, deliberately
left out of that change.

## What is wrong

dplyr's argument bullet quotes marginplyr's rewrite of the caller's expression
rather than what the caller wrote:

```
i In argument: `dplyr::across(dplyr::all_of(c("units", "revenue")), sum, na.rm = TRUE)`.
```

The caller wrote `c(units, revenue)`. `all_of(...)` is the rewrite
`rewrite_grouping_dots()` performs before each branch summarize, so the bullet
describes a call the caller never typed.

## Why it was deferred rather than fixed

ADR 0021 records the reasoning. The two parts fixed in #141 are safe for
reasons this part cannot borrow:

- the grouping-value names are fixed by substituting a `..marginplyr_key_N`
  token that marginplyr itself planted, which is a search for a known literal
  rather than a parse of dplyr's output;
- the blamed call is fixed by assigning a structured `$call` field.

Restoring the caller's spelling has neither property. It means deparsing the
caller's original dots, deparsing dplyr's rendering of the rewritten dots, and
text-matching one against the other — a parse of dplyr's formatting, for the
part of the context least likely to mislead. The caller wrote the expression;
seeing it requoted is confusing rather than wrong about what ran.

## Constraint on any fix

It must degrade to current behaviour when the match fails, as the
deduplication key in #141 does. A rewrite that silently quotes the wrong
expression is worse than one that quotes marginplyr's.

## Severity

Desirable improvement, lower priority than #141. The diagnostic is accurate
about what executed; it is the spelling that is not the caller's.

sayuks:
Three facts from a design review of this ticket, one of which corrects the body.

**1. The `all_of()` rewrite is not `rewrite_grouping_dots()`.** It is
`summary_all_of_expr()`, reached through `resolve_summary_selections()` from
`plan_summary_expressions()` — the planning stage both adapters share, as the
comment at `R/grouping-adapter-native.R:103` says. `rewrite_grouping_dots()`
only substitutes the Grouping-helper constants. This matters for scope: ADR
0021 could confine itself to the `UNION ALL` adapter because the key names and
the repetition are that adapter's, and this part cannot borrow that reason. It
is still confined to that adapter, but because both backends holding
`native_grouping_sets` (duckdb, postgres) are lazy.

[...]

PR #412, opened 2026-09-03
Decide ADR 0022 covers the native adapter's translation error

### LINKED ISSUE #410: Revisit ADR 0022's "covering nothing today" rejection: native-adapter translation errors are unreadable today
This is not a bug report. It asks for a decision about one rejected option in ADR 0022,
because a measurement contradicts the reason recorded for rejecting it.

## The tension inside ADR 0022

*Considered options* rejects extending the restatement to the native adapter:

> **Extend the restatement to the native grouping-sets adapter.** Rejected as
> covering nothing today, since both backends holding the capability are lazy.
> It becomes worth revisiting if an eager backend ever gains
> `native_grouping_sets`.

*Scope, and why it is the same scope as ADR 0021 for a different reason* already records
the case that reason does not cover:

> What confines this instead is which backends hold `native_grouping_sets`:
> duckdb and postgres, both lazy. A branch `summarize()` there builds a query
> without evaluating the caller's expression, so there is no condition raised
> while the verb runs and nothing to restate — the same reason ADR 0021 gives for
> covering eager inputs only. The exception is an error raised while dbplyr
> translates the rewritten expression, which does arrive while the verb runs and
> is left quoting the rewrite.

The exception the Scope section names is a case that exists today, on a lazy backend, and
that the *Considered options* entry's "covering nothing today" does not account for.

## What the exception looks like today

On a native (GROUPING SETS) backend, a summary expression combining `grouping_bit()` with
a column dbplyr cannot resolve produces:

```
ℹ In argument: `total = +...`
Caused by error:
! Object `no_such_column` not found.
```

The same expression on the local (UNION) path, where the restatement applies, produces:

```
ℹ In argument: `total = sum(value) + grouping_bit(a) + no_such_column`.
ℹ In group 1: `a = "x"`.
Caused by error:
! object 'no_such_column' not found
```

Reproduced on both a live DuckDB connection and `dbplyr::simulate_postgres()`, at
`4f367b42bc506e491ad72e99edf0ecb3a3bc5c68`.

## What was measured

Measured by capturing the SQL-rewritten dots the native adapter hands dplyr and running
the package's own `summary_argument_labels()`, `branch_argument_map()`, and
`restate_condition_arguments()` (all `R/conditions.R`) over the real condition objects.
Nothing under `R/` was modified; the end-to-end result below is a simulation in the sense
that the native adapter does not call this machinery today, so the package's real
functions were applied to the real condition after the fact.

1. **The quoted span and marginplyr's own rewritten label agree byte for byte, and the
   match is unique.** dplyr's bullet holds the span `total = +...`; applying
   `summary_argument_labels()` to the captured rewritten dots also gives `total = +...`.
   `span == marginplyr's rewritten label: TRUE`, and `span matches exactly one label:
   TRUE`. So the uniqueness condition ADR 0022 requires before a substitution fires holds
   here. (Boundary case, also measured: two *unnamed* dots that both collapse to `+...`
   do collide, and `branch_argument_map()` then drops the entry and leaves dplyr's
   quotation standing, exactly as documented. Named dots carry the name into the label,
   so `summarize()`'s name uniqueness rules that collision out.)

2. **The caller's own label is not truncated.**
   `rlang::as_label()` of the caller's spelling gives
   `total = sum(value) + grouping_bit(a) + no_such_column`, with no `...` in it.

3. **Applying the mechanism end to end yields a readable bullet:**

   ```
   ℹ In argument: `total = sum(value) + grouping_bit(a) + no_such_column`
   Caused by error:
   ! Object `no_such_column` not found.
   ```

4. **The ADR's crux sentence is therefore false for this case.** *dplyr's rendering is
   reproduced only as far as it is observable* argues:

   > A label dplyr truncated renders the same whichever expression it came from, so
   > substituting the caller's own would print the same `+...`

   Measured directly: handing dbplyr's lazy `summarize()` the caller's own spelling
   produces an *untruncated* bullet
   (`` ℹ In argument: `total = sum(value) + 0L + no_such_column` ``). The truncation is
   not caused by expression length; it is caused by the rewrite splicing in a SQL literal
   whose deparse is multi-line (`sum(value) + <sql literal>` labels as `+...`, the literal
   alone as `<sql>`). And the restoration is a string-level substitution performed after
   the message exists, so there is no stage at which the substituted caller label could be
   truncated again. The sentence is true of a different operation — replacing the
   *expression* in dplyr's dots — which is not what ADR 0022's mechanism does.

5. **The truncation does not depend on expression length, and native-path translation
   errors are almost entirely unreadable.** A short `total = grouping_bit(a) +
   no_such_column` also collapses to `total = +...`; `grouping_bit(a) + 1` labels as
   `+...` as well. Effectively any expression combining a helper with anything else is
   affected, and the one case that escapes — a dot that is the helper alone — labels as
   `<sql>`, which is not readable either.

## Implementation note

`restate_argument_bullet()` (`R/conditions.R:506-519`) matches with
`"^([iℹ] )?In argument: `(.*)`\\.$"`, which requires a trailing period. dbplyr's
bullet does not carry one — its message is `` In argument: `total = +...` `` — so
programmatic extraction with the current regex returns `character(0)` and the current
mechanism passes the message through unchanged even though the map's key already equals
the span. Any change that connects this mechanism to the native path has to include
relaxing that one byte. (Eager dplyr's bullet does end in a period, which is what the
regex was written against.)

## What is being asked

A decision on ADR 0022's *Considered options* entry: either amend that entry's reason in
light of the measurement above, or change the decision itself. This is not a request to
change code first — the implementation note is here so that whichever way the decision
goes, the cost of the "extend it" branch is on the record.

sayuks:
Records the decision on #410 and lands the evidence. Implementation is #411; no `R/` file changes here.

## What was decided

ADR 0022 rejected extending the Condition-context restatement to the native grouping-sets adapter "as covering nothing today, since both backends holding the capability are lazy", while its own *Scope* section named the exception that reasoning misses. #410 measured it, and neither claim holding up the rejection survived:

- The stated reason is contradicted by the ADR's own Scope section — a dbplyr translation error does arrive while the verb runs.
- The crux sentence supporting it, "substituting the caller's own would print the same `+...`", is false for this case. The truncation comes from `rewrite_grouping_dots()` splicing in a `dbplyr` `sql` object whose own `deparse()` is 59 characters, against the width of 60 at which `as_label()` deparses the whole expression — so anything spliced beside it overflows and rlang replaces the call's arguments with `...`. The caller's expression length does not decide it, and the restoration is a string-level substitution with no second truncation stage.

The case is not a corner: every dot combining a helper with anything else renders `+...`, and a dot that is the helper alone renders `<sql>`. The same call on a local input quotes the caller correctly, so today the backend decides the quality of the diagnostic — the asymmetry ADR 0022 exists to remove. `CONTEXT.md`'s *Condition context* entry already promises the restatement without qualification.

## The part most easily lost

ADR 0022 rests #199's hard constraint on "`as_label()` emits neither `+...` nor `.data$x`, so dplyr's label of one argument can never equal marginplyr's label of another." On the native adapter marginplyr's own label **is** `+...`, so that asymmetry is gone. The amendment replaces the argument rather than dropping it: what holds instead is the rule the decision already specifies for a shared label — a label two dots share is restored only where the *callers'* own labels agree as well, and left as dplyr wrote it where they do not; and a named dot carries `name = ` into its label, so it shares a label with neither an unnamed dot nor a differently-named one. Extending the mechanism without writing this down would leave the constraint resting on a sentence that had become false.

## Files

| file | what changed |
|---|---|
| `investigation/what-truncates-an-argument-label-on-the-native-adapter.md` | new — the measurements, with the two simulated steps marked as such |
| `investigation/dplyr-condition-context-rendering.md` | `Revised:` line and a `Revisions (2026-09-04)` section: two of its claims were established on the eager path and do not hold on this one |
| `design/adr/0022-...` | *Amendment: the native adapter's translation error is restated too* |
| `design/adr/0021-...` | citation narrowed to the note's 2026-08-18 revisions section, there now being two |

The evidence goes to a dated note rather than into the ADR, per `investigation/README.md`'s *Authority*: the note is authoritative for what was measured, the ADR for the decision.

Closes #410.

### DIFF
--- a/design/adr/0021-report-a-repeated-execution-condition-once.md
+++ b/design/adr/0021-report-a-repeated-execution-condition-once.md
@@ -101,8 +101,8 @@ Undoing the styling is also not sufficient by itself for the pointer: the
 stripped link renders the call without the backticks dplyr writes around it
 otherwise, so that pattern admits both spellings.
 
-Evidence: `investigation/dplyr-condition-context-rendering.md`, whose revisions
-section holds the measured report counts, the rendered pointer line, and the
+Evidence: `investigation/dplyr-condition-context-rendering.md`, whose 2026-08-18
+revisions section holds the measured report counts, the rendered pointer line, and the
 two things measured beside them and found not to need changing — the wrap
 indent that survives styling, and the error path that carries no styling at
 all.
--- a/design/adr/0022-quote-the-callers-own-spelling-in-a-condition-context.md
+++ b/design/adr/0022-quote-the-callers-own-spelling-in-a-condition-context.md
@@ -212,3 +212,65 @@ synthetic condition standing in for a dplyr wording change asserts that a span
 matching nothing is left alone. A restoration that quietly stopped happening
 would otherwise read exactly like a package whose contexts were all still
 faithful.
+
+## Amendment: the native adapter's translation error is restated too
+
+*Considered Options* rejected extending the restatement to the native
+grouping-sets adapter "as covering nothing today, since both backends holding
+the capability are lazy". *Scope* names the exception in the same document — an
+error raised while dbplyr translates the rewritten expression does arrive while
+the verb runs — so the two disagreed about whether anything was there. The
+exception is what is there, and it was never measured. #410 measured it:
+`investigation/what-truncates-an-argument-label-on-the-native-adapter.md`.
+
+The scope therefore includes that error. A caller whose summary expression
+dbplyr cannot translate is quoted the expression they wrote, on duckdb and
+postgres as on a local input, and the sentence above confining the scope to
+`summarize_margin_union()` no longer holds.
+
+What is extended is the restatement and not `with_branch_conditions()`. The
+native adapter issues one `summarize()` and repeats nothing, so the
+deduplication and the grouping-value restatement have nothing to act on; the
+one sentence of the old reasoning that survives is that one. The error is
+caught around that single `summarize()`, `restate_condition_arguments()` is
+applied with the map `branch_argument_map()` builds from the rewritten dots and
+the caller's labels, and it is re-raised. Nothing else about the adapter moves.
+
+Errors only. A branch `summarize()` on a lazy backend still builds a query
+without evaluating the caller's expression, so everything the original
+reasoning covers is still covered by it; a translation *warning* was searched
+for and not produced, and writing the mechanism against an unmeasured shape is
+what the rest of this decision declines to do.
+
+`restate_argument_bullet()`'s pattern requires a trailing period, and dbplyr's
+bullet does not carry one, so the mechanism read the native bullet as matching
+nothing and passed it through even where the map's key already equalled the
+span. The period becomes optional **and is put back as it was found**: rebuilding
+with an unconditional period would add one to dbplyr's sentence, and the sentence
+around the span is dplyr's.
+
+One argument above stops carrying, and this is what replaces it. *dplyr's
+rendering is reproduced only as far as it is observable* rests the #199
+constraint on `as_label()` emitting neither `+...` nor `.data$x`, so that
+dplyr's label of one argument can never equal marginplyr's label of another. On
+the native adapter marginplyr's own label **is** `+...`, because the rewrite
+splices in a SQL literal whose deparse is multi-line — measured, and independent
+of the length of the caller's expression, which is also why the same section's
+"substituting the caller's own would print the same `+...`" is false here. What
+holds instead is the collision drop this decision already specifies: two unnamed
+dots that both collapse to `+...` produce no map entry and leave dplyr's
+quotation standing, and a named dot carries `name = ` into its label, so it
+collides with neither an unnamed dot nor a differently-named one. The constraint
+is met by the degradation rather than by the asymmetry.
+
+Tests assert on rendered messages through `dbplyr::simulate_postgres()`, which
+needs no optional backend, so the native path is asserted wherever the suite
+runs rather than only where a database is installed. Both directions are
+asserted, as above: the restoration, and the collision that leaves dplyr's
+quotation alone.
+
+`R/summarize_with_margins.R` carried this ADR's rejected reasoning beside its
+citation, as a comment saying no condition is raised while the verb runs. It was
+false on the day it was written, and it is the second copy ADR 0023 names: the
+citation would have stayed correct across this amendment and the re-derivation
+did not. It goes.
--- a/design/adr/0022-quote-the-callers-own-spelling-in-a-condition-context.md
+++ b/design/adr/0022-quote-the-callers-own-spelling-in-a-condition-context.md
@@ -224,53 +224,50 @@ exception is what is there, and it was never measured. #410 measured it:
 `investigation/what-truncates-an-argument-label-on-the-native-adapter.md`.
 
 The scope therefore includes that error. A caller whose summary expression
-dbplyr cannot translate is quoted the expression they wrote, on duckdb and
+dbplyr cannot translate will be quoted the expression they wrote, on duckdb and
 postgres as on a local input, and the sentence above confining the scope to
-`summarize_margin_union()` no longer holds.
+`summarize_margin_union()` no longer holds. #411 implements what follows.
 
 What is extended is the restatement and not `with_branch_conditions()`. The
 native adapter issues one `summarize()` and repeats nothing, so the
-deduplication and the grouping-value restatement have nothing to act on; the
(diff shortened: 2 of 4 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, one round, both axes, against the merge-base with `main` (`4f367b4`) over `a377e12`. Answers are in `fcf0334`.

Standards raised 5, Spec raised 4; three pairs were the same finding, so 6 distinct.

| # | axis | finding | disposition |
|---|---|---|---|
| 1 | Spec | The evidence note's `as_label()` table row for "`grouping_bit(a)`'s rewrite alone" reads `+...`; it labels `<sql>`, contradicting row 1 and the note's own prose fifteen lines below | fixed in prose — the run measured `<sql literal> + 1`, and the row is named for what was run |
| 2 | Spec | The replaced #199 argument states the collision drop as the rule, where `branch_argument_map()` drops a shared label only when the *callers'* labels differ | fixed in prose, in all four places it was written: the amendment, the note, the revisions entry, and #411; #410's decision comment carries a correction comment |
| 3 | both | The amendment describes an implementation and tests as present on a branch that touches no `R/`, and says of a comment still in the tree that it goes | fixed in prose — it names #411 for both |
| 4 | Standards | Kind 2: the overturning argument is written in three places that can drift — the note, the revisions entry, and the amendment's re-derivation of the measurement | fixed in prose — the amendment's re-derivation deleted (the citation at its head carries it), and the note's *What this overturns* section deleted, `investigation/README.md`'s *Authority* putting the consequence in the ADR and the revisions entry |
| 5 | both | Bare "today" in the note's intro, which `investigation/README.md`'s *Tense* forbids by name | fixed in prose |
| 6 | Standards | Kind 2, minor: "That note gains a revisions entry pointing here" names a consumer, and pointers run one way | fixed in prose — inside the section deleted by #4 |

Verified clean by the Standards axis and not re-checked here: the line citations, the quoted regex, every quoted ADR and note phrase, `dbplyr` being an Import, the `investigation/README.md` *Pointers* grep, *Graduation*, the supersession form and placement, unedited originals, and the context budget.

Spec raised no missing deliverable and no scope creep.

sayuks:
## Review record — round 2

Same fixed point (`4f367b4`), now over `a377e12` + `fcf0334`. Answers are in `6e6a136`.

Both axes verified round 1's six answers landed and none introduced a new error. Spec found nothing wrong on the branch: all four artifacts present, all seven required elements in the amendment, no scope creep. It re-ran the two substantive round-1 corrections in R — the `as_label()` table's three rows, and `branch_argument_map()` on two unnamed dots with the callers agreeing and differing — and both hold at all five sites they are written.

One new finding, and two sets of stale citations.

| # | axis | finding | disposition |
|---|---|---|---|
| 1 | Standards | Kind 1: "a SQL literal whose deparse is multi-line" is false in both notes. `grouping_sql_expr()` returns a `sql` object whose own `deparse()` is a **single** 59-character line; what truncates is that 59 nearly fills the width of 60 at which `rlang:::deparse_one()` deparses the whole expression, so anything spliced beside it overflows and rlang replaces the call's arguments with `...`. Not the infix path either — `expr_deparse(e, width = 60)` answers `sum(value) + <sql>` on one line | fixed in prose, reproduced before fixing. Neither conclusion moves: the splice is still the cause and the caller's length still does not decide it, now for the sharper reason that the literal alone almost exhausts the budget |
| 2 | — | Found while fixing #1: the revisions entry carried a second instance of the row round 1 corrected — "`as_label()` answers `+...` for `grouping_bit(a)`'s rewrite on its own", where that rewrite is the bare literal and labels `<sql>`. Round 1 fixed the table and missed this | fixed in prose |
| 3 | Spec | `R/summarize_with_margins.R:1019-1022` is the `tryCatch(`/`{`/`if` lines; the comment is at 1021-1023. `R/grouping-adapter-native.R:62-69` is the argument list; the call spans 61-70 | fixed — both were only ever in #411, which is edited |
| 4 | both | `R/conditions.R:390-397` holds the single-candidate test but stops before the `keep` filter at 398 that effects the drop | fixed — `389-398` in the note, #411, and #410's correction |

Round 1's two corrections had also not been carried into this PR's own body; it is edited to match.

Verified clean and not re-checked here: every other line citation and quoted ADR phrase, the native bullet and its `$parent` chain, the supersession form and placement, unedited originals, the *Pointers* grep, *Graduation*, the tense grep on both notes (the one remaining "today" is inside a quotation of ADR 0022), and `verify-context-budget.R` at 24198/24198.

PR #413, opened 2026-09-03
Keep a declared NA factor level across the dtplyr branch union

### LINKED ISSUE #408: dtplyr backend drops a declared NA factor level under `.margin_label = NULL`
## Summary

On a dtplyr input, `summarize_with_margins(.margin_label = NULL)` returns a factor
dimension whose declared `NA` level has been dropped. The same call on the same data
through the local `data.frame` backend keeps it, so the result's column domain differs
by backend with no error, warning, or other diagnostic.

## Reproduction

```r
# Repro BC-1: dtplyr backend loses a declared NA factor level when
# .margin_label = NULL (typed missing), diverging from the local backend.

data <- data.frame(
  group = structure(c(1L, 1L), levels = c("x", NA_character_), class = "factor"),
  value = 1:2
)
cat("input levels: "); print(levels(data$group))

local_res <- summarize_with_margins(
  data, n = dplyr::n(), bit = grouping_bit(group),
  .grouping = rollup(group), .margin_label = NULL
)
dt_res <- dplyr::collect(summarize_with_margins(
  dtplyr::lazy_dt(data), n = dplyr::n(), bit = grouping_bit(group),
  .grouping = rollup(group), .margin_label = NULL
))

cat("\n--- local backend ---\n")
print(local_res)
cat("levels(local_res$group): "); print(levels(local_res$group))

cat("\n--- dtplyr backend ---\n")
print(dt_res)
cat("levels(dt_res$group): "); print(levels(dt_res$group))

cat("\nidentical levels? ", identical(levels(local_res$group), levels(dt_res$group)), "\n")

# Control: a plain dtplyr round trip keeps the declared NA level.
rt <- dplyr::collect(dtplyr::lazy_dt(data))
cat("plain round-trip levels: "); print(levels(rt$group))

# Control: same call with a non-missing label keeps the NA level on dtplyr.
dt_lab <- dplyr::collect(summarize_with_margins(
  dtplyr::lazy_dt(data), n = dplyr::n(),
  .grouping = rollup(group), .margin_label = "Total"
))
cat("dtplyr with .margin_label = 'Total' levels: "); print(levels(dt_lab$group))
```

Reproduced at `4f367b42bc506e491ad72e99edf0ecb3a3bc5c68`.

## Observed / Expected

Observed, verbatim:

```
input levels: [1] "x" NA 

--- local backend ---
  group n bit
1     x 2   0
2  <NA> 2   1
levels(local_res$group): [1] "x" NA 

--- dtplyr backend ---
# A tibble: 2 × 3
  group     n   bit
  <fct> <int> <int>
1 x         2     0
2 <NA>      2     1
levels(dt_res$group): [1] "x"

identical levels?  FALSE 
plain round-trip levels: [1] "x" NA 
dtplyr with .margin_label = 'Total' levels: [1] "x"     NA      "Total"
```

Expected: `levels(dt_res$group)` is `c("x", NA)`, as `levels(local_res$group)` is, and
`identical levels?` is `TRUE`. Nothing else in the output changes.

The two controls are what confine the cause to marginplyr's typed-missing path rather
than to dtplyr: a plain `dtplyr::lazy_dt(data) |> collect()` round trip preserves the
declared NA level, and the same marginplyr call on the same dtplyr input preserves it
when `.margin_label = "Total"` selects the non-missing branch.

## Contract violated

ADR 0012, the eight-case table, the row for `NULL` / NA level = yes / missing value = no:

> | `NULL` | yes | no | Allowed; preserve the NA level and use typed missing for margins |

and its prose:

> `NULL` requests typed missing values and therefore bypasses Margin label
> collision checks. It preserves existing factor levels, including an NA
> level.

Neither is qualified by backend.

ADR 0016, *The promise boundary*, places the levels on the promised side rather than the
passed-through side:

> marginplyr constructs, and therefore promises:
>
> [...]
> - factor and ordered-factor levels, ordering, and Margin-label placement, per
>   ADR 0012.

So a level set that differs between the local and the dtplyr backend is a divergence in
something the package promises, not in something it describes.

The one clause considered as an exemption and rejected is ADR 0012's test consequence
"local and factor-preserving lazy adapters where supported": both controls above show
dtplyr *is* factor-preserving here, so the clause does not reach this case.

## Implementation site

`R/margin-label.R:371` (`label_margin_branch()`), specifically the `values <- lapply(...)`
block at `R/margin-label.R:437-450` and the comment above it at `R/margin-label.R:428-431`.

## Root cause: a partial correction to the obvious reading

The obvious reading of the site is that the `rep(!!label, dplyr::n())` treatment at
`R/margin-label.R:441-443` is applied only in the non-missing branch
(`!is_missing_margin_label(label)`) and should be carried into the typed-missing branch
too. **That reading is probably not the whole fix, and porting `rep()` alone may well not
repair this.**

Measured separately: a dtplyr scalar `mutate(group = NA)` on its own *does* preserve the
levels `c("x", NA)`, while data.table's `funion()` / `rbindlist()` drop a declared NA
level when they union two factor columns (nlev 2 -> 1). The point at which the level
disappears is therefore not the scalar assignment inside the branch, but the fact that
the typed-missing path reaches the branch union without putting the factor column on the
encode/restore route the non-missing path uses. A fix should be verified against the
union, not against the assignment.

This does not affect the finding itself, and `label_margin_branch()` remains where the
defect lives.

## Nearest existing test and why it did not catch this

`tests/testthat/test-margin-label.R:42`, `"factor NA levels and missing values obey the
eight-case contract"`. It builds its cases as local `data.frame`s and runs all eight rows
of the ADR 0012 table there only; the eight cases have never been run through the dtplyr
backend. The dtplyr tests in the same file cover a declared-label collision refusal, a
non-missing label, and ordered-factor restoration — the `.margin_label = NULL` (typed
missing) x declared NA level combination is not among them.

## Not verified

- The same case (`.margin_label = NULL` x declared NA level) on duckdb, postgres, arrow,
  or sqlite. Those backends do not go through the data.table union this defect involves,
  so whether the same defect exists there is unmeasured either way.
- Whether the other Margin verbs reach the same branch-union path with the same result.

sayuks:
Fixes #408.

## What was wrong

On a dtplyr input, `summarize_with_margins(.margin_label = NULL)` returned a
factor dimension whose declared `NA` level had been dropped; the same call on
the same data through the local backend kept it. ADR 0012's sixth row promises
the level is preserved, and ADR 0016 places factor levels on the side
marginplyr constructs. Neither is qualified by backend.

## Root cause

`data.table`'s rbind drops a declared `NA` level from the result and turns the
values that used it into missing codes — for a list of one table as much as for
a union of branches:

```r
f <- structure(c(1L, 1L), levels = c("x", NA_character_), class = "factor")
a <- data.table::data.table(group = f)
levels(data.table::rbindlist(list(a))$group)
#> [1] "x"
```

The non-missing-label path never met that. A label is not one of the column's
levels, so its columns already crossed the union as character and were rebuilt
after it. A missing label adds no level, so its columns crossed as factors, and
the level was gone before restoration could keep it.

## The fix

A factor dimension whose Margin label is missing now takes the same
encode-and-rebuild route where the union would drop such a level. That is the
new `drops_na_factor_level_on_union` capability in `backend_capabilities()`,
granted to `dtplyr` alone.

The encoding has to be the sentinel one rather than `as.character()`: a value
on the `NA` level and the typed missing a margin row carries are both `NA` in
character, and the union is where the two would stop being distinguishable.
`margin_column_info()` decides it once per column as `encode_missing_label`, so
`label_margin_branch()` and `restore_margin_factors()` read one answer instead
of deciding twice.

## Backends other than dtplyr

Measured, and untouched. A plain duckdb round trip already drops the level
(`levels()` is `"x"` before any Margin verb runs), and `arrow` refuses the
column outright — `Invalid: Cannot insert dictionary values containing nulls`.
Both lose it before any branch exists, so ADR 0012's *local and
factor-preserving lazy adapters where supported* clause reaches dtplyr alone.

## Tests

The eight cases of ADR 0012's table had only ever run on a local data frame.
The table moved to a `factor_contract_cases()` helper and now runs on dtplyr
too, each case compared against the local result rather than against a literal.

A second test covers a value that *uses* the `NA` level, which no case in the
table does: it asserts the source row stays on the level (`is.na()` false) while
the margin rows are typed missing (`is.na()` true), for both
`summarize_with_margins()` and `expand_with_margins()`. That is what an encoding
which merely restored the levels would lose.

Row order is not asserted across backends: the two place a missing value within
one grouping set differently, which ADR 0018 leaves to the backend.

## Local checks

- `testthat::test_local()` with `NOT_CRAN=true`: 6373 pass, 0 fail, 0 skip.
- `lintr::lint_package()` after `pkgload::load_all()`: no lints.
- `.github/scripts/verify-context-budget.R`: within baseline.
- `roxygen2::roxygenise()`: no generated-file diff (no roxygen changed).

### DIFF
--- a/R/backend-metadata.R
+++ b/R/backend-metadata.R
@@ -101,10 +101,10 @@ margin_column_info <- function(data_proxy,
           # is at risk, only a union that drops one puts it there, and only a
           # sentinel keeps a value on that level apart from the typed missing
           # a margin row carries, which `as.character()` spells the same way
-          # (ADR 0012). The backend is read here because this is the one place
-          # it is, and both sides of the route -- `label_margin_branch()` and
-          # `restore_margin_factors()` -- then read one answer rather than
-          # deciding twice.
+          # (ADR 0012). Settled here so that both sides of the route --
+          # `label_margin_branch()` and `restore_margin_factors()` -- read one
+          # answer rather than deciding twice, which is what they would do:
+          # neither is handed the backend.
           encode_missing_label = has_na_in_level &&
             backend$drops_na_factor_level_on_union &&
             backend$can_encode_factor_missing_values
--- a/R/backend-metadata.R
+++ b/R/backend-metadata.R
@@ -88,12 +88,26 @@ margin_column_info <- function(data_proxy,
       names(schema)[vapply(schema, is.factor, logical(1))],
       function(col) {
         x <- schema[[col]]
+        has_na_in_level <- anyNA(levels(x))
         list(
           col = col,
           levels = levels(x),
           ordered = is.ordered(x),
-          has_na_in_level = anyNA(levels(x)),
-          preserve_missing_value = backend$can_encode_factor_missing_values
+          has_na_in_level = has_na_in_level,
+          preserve_missing_value = backend$can_encode_factor_missing_values,
+          # Whether this column takes the encode-and-rebuild route even where
+          # its Margin label is missing and so adds no level. All three terms
+          # are necessary and none is implied by the others: only an NA level
+          # is at risk, only a union that drops one puts it there, and only a
+          # sentinel keeps a value on that level apart from the typed missing
+          # a margin row carries, which `as.character()` spells the same way
+          # (ADR 0012). The backend is read here because this is the one place
+          # it is, and both sides of the route -- `label_margin_branch()` and
+          # `restore_margin_factors()` -- then read one answer rather than
+          # deciding twice.
+          encode_missing_label = has_na_in_level &&
+            backend$drops_na_factor_level_on_union &&
+            backend$can_encode_factor_missing_values
         )
       }
     )
--- a/R/factor.R
+++ b/R/factor.R
@@ -19,22 +19,17 @@ encodes_missing_label_factor <- function(info, label) {
   is_missing_margin_label(label) && isTRUE(info$encode_missing_label)
 }
 
-# The sentinel each such column is carried as, keyed by column; empty where no
-# column takes the route. Built once per branch so that the value a branch
-# omitting the dimension writes is the one the branches including it encode to.
-missing_label_sentinels <- function(factor_info, margin_labels) {
-  encoded <- Filter(
-    function(info) {
-      encodes_missing_label_factor(info, margin_labels[[info$col]])
-    },
-    factor_info
-  )
+# The sentinel each factor dimension is carried as while it crosses the branch
+# union as character, keyed by column. Derived once so that the value a branch
+# omitting the dimension writes is the one the branches including it encode to,
+# rather than the two arms deriving it apart.
+margin_factor_sentinels <- function(factor_info, margin_labels) {
   stats::setNames(
     lapply(
-      encoded,
+      factor_info,
       function(info) factor_missing_sentinel(info, margin_labels[[info$col]])
     ),
-    vapply(encoded, function(info) info$col, character(1))
+    vapply(factor_info, function(info) info$col, character(1))
   )
 }
 
--- a/R/factor.R
+++ b/R/factor.R
@@ -11,6 +11,33 @@ factor_missing_sentinel <- function(info, .margin_name) {
   sentinel
 }
 
+# Whether a factor dimension crosses the branch union as character despite a
+# missing Margin label, which is what `encode_missing_label` in
+# `R/backend-metadata.R` records for the column. Both sides of the route read
+# this, so neither can encode what the other does not rebuild.
+encodes_missing_label_factor <- function(info, label) {
+  is_missing_margin_label(label) && isTRUE(info$encode_missing_label)
+}
+
+# The sentinel each such column is carried as, keyed by column; empty where no
+# column takes the route. Built once per branch so that the value a branch
+# omitting the dimension writes is the one the branches including it encode to.
+missing_label_sentinels <- function(factor_info, margin_labels) {
+  encoded <- Filter(
+    function(info) {
+      encodes_missing_label_factor(info, margin_labels[[info$col]])
+    },
+    factor_info
+  )
+  stats::setNames(
+    lapply(
+      encoded,
+      function(info) factor_missing_sentinel(info, margin_labels[[info$col]])
+    ),
+    vapply(encoded, function(info) info$col, character(1))
+  )
+}
+
 encode_factor_for_margin <- function(x,
                                      missing_sentinel,
                                      preserve_missing_value) {
@@ -72,8 +99,15 @@ restore_margin_factors <- function(.data,
   Reduce(
     function(data, info) {
       label <- margin_labels[[info$col]]
+      # A missing label adds no level, so a column the branches carried as
+      # their own class arrives with its levels already intact. The exception
+      # is one that crossed the union as character to keep them, which is
+      # rebuilt on its declared levels: `NULL` appends none.
       if (is_missing_margin_label(label)) {
-        return(data)
+        if (!encodes_missing_label_factor(info, label)) {
+          return(data)
+        }
+        return(reconstruct_factor(data, info, NULL, position = position))
       }
       reconstruct_factor(data, info, label, position = position)
     },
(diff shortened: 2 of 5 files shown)


sayuks:
## Review record

One round of `mattpocock-skills:code-review`, both axes, against
`git diff main...HEAD` at the merge-base `4f367b42bc506e491ad72e99edf0ecb3a3bc5c68`,
reading `dc5a921`. Answered in `1b83cf7`.

Standards sources passed: the sections of `AGENTS.md` the diff reaches (*Linting*,
*Documentation*, *Dependency metadata*, *Optional-dependency guards*, *Release matrix*,
*Queries against a lazy input*, *Code comments*), `CONTEXT.md`, ADRs 0012, 0015, 0016,
0018, 0020, and `design/architecture.md`'s *Answering a review* for what a finding about
prose may be.

### Standards — 3 findings, 2 smells

| Finding | Disposition |
|---|---|
| `R/factor.R` — `missing_label_sentinels()`'s comment claimed the list is built once so both arms of a branch write the same string; the including arm called `factor_missing_sentinel()` again instead of reading it. False comment. | Fixed in code: `margin_factor_sentinels()` is keyed over every factor dimension and both arms read it. This also answers the *Duplicated Code* smell below — the derivation dropped from three sites to two, and the remaining one is `restore_margin_factors()`, on the other side of the union. |
| `R/backend-metadata.R` — "The backend is read here because this is the one place it is" is false: the same function reads `can_read_schema`, `can_restore_factors`, and `can_encode_factor_missing_values`, and `expand_margin_union()` reads `invents_row_on_column_add`. | Fixed in prose: the comment now says why the decision is settled there — neither side of the route is handed the backend. |
| Residual divergence, measured: a factor column with a declared NA level still loses it on dtplyr as a fixed `.by` key or as a passed-through column. | Ticket: #415. This branch's Acceptance is a Margin dimension, and #415 asks a question this branch does not — whether ADR 0016's promise boundary reaches a column marginplyr does not label. |
| *Duplicated Code* — the sentinel derived at three sites. | Fixed with the first finding. |
| *Duplicated Code* — the two eight-case loops assert the refusals differently (`conditionCall` vs `marginplyr_error`), so the two error rows were the one part of the contract not compared across backends. | Fixed in code: both loops now assert the call the refusal names. |

Checked and confirmed sound by the round, needing no change: `setNames`/`Filter` on empty
input; `margin_labels` keys always covering `factor_info`; `margin_factor_levels(info, NULL, …)`
passing `stopifnot` on `logical(0)`; dtplyr having no `native_grouping_sets`, so no path
reaches `restore_margin_factors()` without the branch encoding; no new execution entry
point (ADR 0020); the new test needing one backend (*Release matrix*).

### Spec — 1 finding

| Finding | Disposition |
|---|---|
| The nest verbs pass but nothing pinned them; the issue's *Not verified* asks whether the other Margin verbs reach the same branch-union path with the same result. | Fixed in code: the used-NA-level test covers all four Margin verbs. |

The round confirmed the issue's reproduction now prints its *Expected* section with
nothing else in the output changed, and independently measured the *Not verified*
backends — duckdb drops the level on the round trip before any branch, arrow errors on
ingest, sqlite returns character — so restricting `drops_na_factor_level_on_union` to
dtplyr is what ADR 0012's *factor-preserving lazy adapters* clause reaches. It found no
requirement implemented wrongly, having checked ordered factors with an NA level, a
named-list `.margin_label` mixing `NULL` with a label, `cube()` over two NA-level
dimensions, `.margin_label_position = "first"`, `.by` present, a `data.table` input,
`.check_margin_label = FALSE`, `.duplicates = "keep"`, and a level literally equal to the
sentinel string. It reported no scope creep.

### Local checks at `1b83cf7`

- `testthat::test_local()` with `NOT_CRAN=true`: 6373 pass, 0 fail, 0 skip.
- `lintr::lint_package()` after `pkgload::load_all()`: no lints.
- `roxygen2::roxygenise()`: no generated-file diff.

sayuks:
Correction to the record above: the full-suite count at `1b83cf7` is **6381** pass, 0 fail, 0 skip. 6373 was the count before the used-NA-level test was widened to all four Margin verbs in that same commit. The lint and roxygen lines are unchanged.

PR #414, opened 2026-09-03
Empty the Sent query record at every entry point

### LINKED ISSUE #409: `last_sent_queries()` returns a stale prior-call record when a call fails before `reset_sent_queries()`
## Summary

When an audited Margin call is refused by argument validation that runs *before*
`prepare_grouping_plan()` — and therefore before `reset_sent_queries()` — the record is
never emptied, and `last_sent_queries()` then hands back the previous successful call's
rows verbatim as if they belonged to the failed call. The documented answer for a call
that sent nothing is a zero-row tibble.

## Reproduction

```r
# Repro FH-2: last_sent_queries() returns the previous successful call's
# record verbatim after an audited call fails in argument validation before
# prepare_grouping_plan() (and thus reset_sent_queries()) is reached.

old <- options(marginplyr.audit_sql = TRUE)
con <- DBI::dbConnect(duckdb::duckdb(shared_home = FALSE))
db <- dplyr::copy_to(
  con,
  tibble::tibble(region = c("east", "west"), revenue = c(10, 20)),
  name = "t1", temporary = TRUE
)

invisible(summarize_with_margins(
  db, revenue = sum(revenue, na.rm = TRUE), .grouping = rollup(region)
))
r1 <- last_sent_queries()
cat("--- after successful call 1: purposes ---\n")
print(r1$purpose)

err <- tryCatch(
  summarize_with_margins(
    db, revenue = sum(revenue, na.rm = TRUE),
    .grouping = rollup(region), .sort = "bogus_choice"
  ),
  error = function(e) e
)
cat("\n--- call 2 refused with ---\n")
cat(class(err)[1], ":", conditionMessage(err), "\n")

r2 <- last_sent_queries()
cat("\n--- last_sent_queries() after the failed call 2 ---\n")
print(r2$purpose)
cat("rows:", nrow(r2), "\n")
cat("identical(r1, r2):", identical(r1, r2), "\n")

DBI::dbDisconnect(con, shutdown = TRUE)
options(old)
```

Reproduced at `4f367b42bc506e491ad72e99edf0ecb3a3bc5c68`.

## Observed / Expected

Observed, verbatim:

```
--- after successful call 1: purposes ---
[1] "selection_proxy" "result"         

--- call 2 refused with ---
marginplyr_error : `.sort` must be one of "none", "last", or "first". 

--- last_sent_queries() after the failed call 2 ---
[1] "selection_proxy" "result"         
rows: 2 
identical(r1, r2): TRUE
```

Expected: call 2 sent no query, so the read after it reports the zero-row answer —
`character(0)` for the purposes, `rows: 0`, and `identical(r1, r2): FALSE`. The refusal
of call 2 itself is correct and unchanged.

A secondary consequence follows from the same missing reset: the `audited` flag is stale
too, so a caller who changes `marginplyr.audit_sql` between the two calls gets the
previous call's audit state deciding which of the four answers is given.

## Contract violated

`R/sent-queries.R`, the roxygen of `last_sent_queries()`:

> marginplyr keeps a record of the SQL of one call -- the most recent call to
> [summarize_with_margins()], [expand_with_margins()], [nest_with_margins()],
> [nest_by_with_margins()], or [inspect_grouping()] -- and this function reads it back.
> The record is emptied at the start of every call, so what it holds belongs to that call
> alone.

Call 2 is the most recent call to `summarize_with_margins()`; the record was not emptied
at its start, and what is readable belongs to call 1.

> A query is recorded before it is sent, so the record holds what was sent rather than
> what succeeded, and a call that fails leaves readable every query it had already sent.

Call 2 had already sent none. What is readable is two.

And the *Four answers* section names the correct answer explicitly:

> - The call was audited and sent nothing: a zero-row tibble, the only answer with zero
>   rows.

The observed two-row prior record is none of the four documented answers.

ADR 0027, *One call, and which call*, is where the reset site was chosen, and it rules
this shape out by name:

> Under the obvious placement, `inspect_grouping()` appends its proxy query to whatever
> the previous Margin operation left behind, and nothing says the record now spans two
> calls — the `dbplyr::last_sql()` defect, rebuilt inside the feature written to answer
> it.

The clause considered as an exemption and rejected is that ADR's second answer, "The last
**recorded** call was not audited": that word describes the mechanism of the unaudited
refusal, while the user-facing promise in the roxygen is unqualified, and validation runs
inside the body of `summarize_with_margins()`, so the failed call is not one that never
began.

## Implementation site

- `R/sent-queries.R:16-21` — `reset_sent_queries()`, the only site that empties the record.
- `R/grouping-plan.R:144-154` — `prepare_grouping_plan()`, which calls it at its top.
- `R/summarize_with_margins.R:838-867` — the validation block that runs *before*
  `prepare_margin_operation()` and therefore before that reset:
  `assert_margin_input()`, `assert_lazy_table()`, `normalize_margin_options()` (which is
  what refuses `.sort` here), `assert_logical_scalar()`,
  `check_option_named_summaries()`, `check_empty_summaries()`,
  `check_summary_context_helpers()`, and `preflight_shares()`.

Any refusal from that block leaves the record stale, so `.sort` is one entry point among
several rather than the defect itself.

## Nearest existing test and why it did not catch this

`tests/testthat/test-sent-queries.R:450`, `"a refused share leaves the probe's row
readable"`. It covers a refusal that happens *after* the reset has been reached — the
dialect probe row is already recorded — and correctly asserts the refused call's own
partial record. No test reads `last_sent_queries()` after an audited call that failed in
the early validation block, i.e. *before* the reset.

## Not verified

- The same sequence on other SQL backends (postgres, sqlite). The defect is in the call
  sequence rather than in a backend branch, so it is expected to hold there, but it was
  not measured.
- The same sequence through `expand_with_margins()`, `nest_with_margins()`,
  `nest_by_with_margins()`, and `inspect_grouping()`. Each has the same early-validation
  structure, but only `summarize_with_margins()` was measured.

sayuks:
Fixes #409.

## What was wrong

`reset_sent_queries()` sat at the top of `prepare_grouping_plan()`, which every
entry point reaches before any recorded site — and *after* the argument
validation each of them opens with. A call refused there never emptied the
record, so `last_sent_queries()` answered it with the previous successful
call's rows verbatim, and the audited flag deciding which of ADR 0027's four
answers to give was that previous call's too.

The issue's reproduction now prints what it says it should:

```
--- after successful call 1: purposes ---
[1] "selection_proxy" "result"

--- call 2 refused with ---
marginplyr_error : `.sort` must be one of "none", "last", or "first".

--- last_sent_queries() after the failed call 2 ---
character(0)
rows: 0
identical(r1, r2): FALSE
```

## What changed

The reset is the first statement of `summarize_with_margins()`,
`expand_with_margins()`, `nest_with_margins()`, `nest_by_with_margins()`, and
`inspect_grouping()`, ahead of everything that can raise. What made
`prepare_grouping_plan()` the site is preserved rather than traded away: the
five reach it still, and a reset before their validation is a reset before
every recorded site. Neither thing that function computes is read by the reset
— `.grouping` is unresolved and the backend unknown — so
`remember_sent_query_backend()` stays where the backend is computed.

Five sites are not five rules. `test-sent-queries.R` gains a structural gate
over the loaded namespace: the set of functions calling `reset_sent_queries()`
is exactly the set of exported functions taking `.grouping`, and each calls it
first. An entry point added without a reset, and a validation moved above one,
are each what it fires on — verified by removing the reset from
`expand_with_margins()`, which names that verb in all three assertions.

The behavioural half runs every entry point's early-validation refusal after an
audited RSQLite call that recorded rows, and asserts the record it leaves is
the zero-row one; a second test asserts the refusal re-reads
`marginplyr.audit_sql` for itself.

ADR 0027 carries the move as an amendment, superseding the reset site its
opening paragraph and *One call, and which call* named. `NEWS.md` gains the
sentence for it.

## Checks

`lintr::lint_package()` clean; full suite green with `NOT_CRAN=true`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -304,6 +304,7 @@
   share, with the control that tells a refusal from a question that could
   not be put. Each is recorded before it is sent, so a call refused by a
   collision or by an ineligible share source leaves readable every query it
-  had already sent. A call refused before it sent anything -- by an argument
-  its verb rejects -- reads back as the empty record it is, rather than as the
-  previous call's. The option is off by default (#318, #400, #401, #409).
+  had already sent. A call one of these verbs refused before sending anything
+  reads back as the empty record it is, rather than as the previous call's,
+  and is a call the session has recorded. The option is off by default
+  (#318, #400, #401, #409).
--- a/NEWS.md
+++ b/NEWS.md
@@ -304,4 +304,6 @@
   share, with the control that tells a refusal from a question that could
   not be put. Each is recorded before it is sent, so a call refused by a
   collision or by an ineligible share source leaves readable every query it
-  had already sent. The option is off by default (#318, #400, #401).
+  had already sent. A call refused before it sent anything -- by an argument
+  its verb rejects -- reads back as the empty record it is, rather than as the
+  previous call's. The option is off by default (#318, #400, #401, #409).
--- a/R/expand_with_margins.R
+++ b/R/expand_with_margins.R
@@ -92,6 +92,7 @@ expand_with_margins <- function(.data,
                                 .duplicates = c("error", "drop", "keep"),
                                 .id = NULL,
                                 .sort = c("none", "last", "first")) {
+  reset_sent_queries()
   call <- rlang::current_call()
   with_margin_error_call(
     {
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -149,9 +149,6 @@ prepare_grouping_plan <- function(.data,
                                   validate_grouping = NULL,
                                   validate_names = NULL,
                                   call = rlang::caller_call()) {
-  # First, before anything that can raise: every entry point reaches this
-  # function, and every recorded site runs after it (ADR 0027).
-  reset_sent_queries()
   stopifnot(rlang::is_quosure(by_quo), rlang::is_quosure(grouping_quo))
   stopifnot(is.null(validate_grouping) || is.function(validate_grouping))
   stopifnot(is.null(validate_names) || is.function(validate_names))
--- a/R/inspect-grouping.R
+++ b/R/inspect-grouping.R
@@ -136,6 +136,7 @@ inspect_grouping <- function(.data,
                              .grouping = NULL,
                              .duplicates = c("error", "drop", "keep"),
                              .format = c("text", "list")) {
+  reset_sent_queries()
   call <- rlang::current_call()
   grouping_quo <- rlang::enquo(.grouping)
   by_quo <- rlang::enquo(.by)
--- a/R/nest_by_with_margins.R
+++ b/R/nest_by_with_margins.R
@@ -119,6 +119,7 @@ nest_by_with_margins <- function(.data,
                                  .sort = c("none", "last", "first"),
                                  .key = "data",
                                  .keep = FALSE) {
+  reset_sent_queries()
   call <- rlang::current_call()
   grouping_quo <- rlang::enquo(.grouping)
   by_quo <- rlang::enquo(.by)
--- a/R/nest_with_margins.R
+++ b/R/nest_with_margins.R
@@ -165,6 +165,7 @@ nest_with_margins <- function(.data,
                               .sort = c("none", "last", "first"),
                               .key = "data",
                               .keep = FALSE) {
+  reset_sent_queries()
   call <- rlang::current_call()
   if (is.null(.key)) {
     .key <- "data"
--- a/R/sent-queries.R
+++ b/R/sent-queries.R
@@ -10,11 +10,9 @@ sent_queries <- new.env(parent = emptyenv())
 # `TRUE` is "not audited", and nothing here reports it.
 #
 # Called as the first statement of every entry point, ahead of the argument
-# validation each of them opens with: a call refused there is a call that
-# began, and what `last_sent_queries()` answers for it is its own empty record
-# rather than the previous call's (ADR 0027). The backend is not known yet, so
-# the SQL flag starts `FALSE` and `remember_sent_query_backend()` sets it where
-# the backend is computed.
+# validation each of them opens with (ADR 0027). The backend is not known yet,
+# so the SQL flag starts `FALSE` and `remember_sent_query_backend()` sets it
+# where the backend is computed.
 reset_sent_queries <- function() {
   sent_queries$recorded <- TRUE
   sent_queries$audited <- isTRUE(getOption("marginplyr.audit_sql"))
@@ -97,8 +95,10 @@ record_sent_query <- function(purpose, query) {
 #' call made in this process and not one made in a worker.
 #'
 #' @section Four answers:
-#' - Nothing has been recorded in this session -- no Margin verb has run --
-#'   and the call is refused with a `"marginplyr_error"`.
+#' - Nothing has been recorded in this session -- no Margin verb has begun --
+#'   and the call is refused with a `"marginplyr_error"`. A verb that began and
+#'   then refused your call has begun, so what is read after one is its own
+#'   record, empty, and not this refusal.
 #' - The last call was not audited, the option being unset or holding a value
 #'   other than `TRUE` when the call began, and the call is refused with a
 #'   `"marginplyr_error"` naming `marginplyr.audit_sql`. The option is read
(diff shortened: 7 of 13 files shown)


sayuks:
## Code review record

One round, `mattpocock-skills:code-review`, against the merge-base with `main`
(`4f367b4`) over `b09942a`. Standards raised four findings, Spec four. All
eight are answered in `004d213`; none was rejected.

### Standards

| # | Finding | Disposition |
|---|---|---|
| 1 | The ADR amendment re-made #409's case, which `design/architecture.md` calls a second copy of the ticket's | fixed in prose |
| 2 | `reset_sent_queries()`'s header re-derived the amendment's argument beside its citation, against *Code comments* in `AGENTS.md` | fixed in prose |
| 3 | Duplicated Code: the gate's entry-point derivation is the shape of `verbs_taking()` in `test-grouping-plan.R`, minus its `is.function()` guard | fixed in code — `verbs_taking()` moves to `tests/testthat/helper-margin-verbs.R`, read by both files |
| 4 | `first_statement()` indexes `fn_body[[2]]` unconditionally, so an empty braced body raises rather than reporting | fixed in code, with the case added to the self-witness |

### Spec

| # | Finding | Disposition |
|---|---|---|
| 1 | "Ahead of everything that can raise", in the ADR and in `NEWS.md`, is wider than the fix | fixed in prose — the ADR states the bound, `NEWS.md` says what the verbs refuse |
| 2 | The reset marks the session as having recorded a call, moving which of the four answers a session of refusals gets; the ADR said the four were untouched and the roxygen's first answer said no verb had run | fixed in prose and in code — both now say what happens, and a test asserts it |
| 3 | ADR *Corrections*' first entry still named `prepare_grouping_plan()` as the reset site | fixed in prose — the amendment supersedes that clause too |
| 4 | The audit option's off-to-on direction was untested | fixed in code |

Finding 1's premise, measured on this branch:

```r
summarize_with_margins(db, total = sum(v, na.rm = TRUE), .grouping = rollup(region))
nrow(last_sent_queries())
#> 2
expand_with_margins(db, .grouping = rollup(region), bogus = 1)
#> Error: unused argument (bogus = 1)
nrow(last_sent_queries())
#> 2
```

R matches arguments before the body runs, so such a call reaches no statement
of one. That is a call that never began, under the same reading by which one
refused inside the body did, and the amendment's second paragraph is where the
bound is now written.

Finding 2's, on a session with the record emptied:

- before: `Error: No Margin operation has run in this session.` — the first answer
- after: a zero-row tibble — the third

`test-sent-queries.R`'s *a call refused before its plan is a call the session
recorded* is what fails without it.

`lintr::lint_package()` clean; full suite green under `NOT_CRAN=true`.

sayuks:
Pushed `6b20253` to this branch — the coverage job was failing here, and it was the gate rather than the package.

`test-coverage.yaml` runs `covr::package_coverage()`, and covr measures a namespace by replacing each statement in it with `if (TRUE) { covr:::count(<key>); <statement> }`. `first_statement()` reads a statement by **position**, so what it read was covr's wrapper — hence all six entry points reported at once against a package where every one of them opens with `reset_sent_queries()`, and hence `R-CMD-check.yaml` and `release-matrix.yaml` passing on the same SHA. Every other structural gate in the suite goes through `visit_calls()`, which finds a call wherever the wrapper puts it, which is why this is the only gate that noticed.

`first_statement()` now takes the wrapper off on both of its paths — covr wraps an unbraced body whole and wraps each statement of a braced one.

Not a `skip()` under covr: a structural gate that stops asserting in one job reads exactly like a gate nothing violates. The wrapper's shape is covr's and undocumented, so a covr that changed it fails this gate rather than quieting it, which is the direction this reading has to fail in.

Verified two ways locally:
- `covr:::trace_calls()` over the six real entry points — all six answer the reset instrumented and uninstrumented alike.
- The workflow's own command, `covr::package_coverage(quiet = FALSE, clean = FALSE, install_path = ...)`, run to completion: no failures.

The self-witness gained the covr shape, written out rather than produced by calling covr — covr is supplied by the workflow's `extra-packages` and is in no dependency field, so a test that called it would put it in one. It also asserts the other direction: an `if (TRUE)` a caller wrote holds no counter and is not taken apart.

PR #416, opened 2026-09-04
Cite a location by name rather than by line number

sayuks:
A line number fails differently from every other reference this repository writes. A renamed function or a moved file makes a citation unfindable, and a grep reports it; an edit above a cited line leaves the citation resolving, to something else. Nothing reports that, so the claim stays in the closure reading as though it were still true.

## What the rule says

Added to *Code comments* in `AGENTS.md`, after the paragraph on what a citation does:

> This holds wherever a maintained document points, not only in a comment: a citation names its target — a function, a heading, a chunk, an ADR or issue number — and not a line number, which an edit above it leaves pointing at a different line rather than at nothing. A SHA, and a permalink pinned to one, do not move.

## The measurement

A scan over every tracked source, excluding generated files and `investigation/`, found two violations. Both were already false:

| site | cited | what that line now holds | drift |
|---|---|---|---|
| `test-static-expression-analysis.R` | `R/share.R:307` | roxygen about non-numeric sources, not the `.fns` bullet meant | 8 lines |
| `test-static-expression-analysis.R` | `R/share.R:786` | `[[` inside a `vapply()` over `ordinary_records`, not the guard | 71 lines |

Both are rewritten to name what they point at: the *Indirect `.fns`* bullet of `share_of_parent()`'s *Rejected forms and supported rewrites*, and `plan_share_expressions()`.

The same scan found five coordinates that are not references — the testthat skip output `verify-backend.R` parses, and the synthetic covr counter keys `test-sent-queries.R` builds. The rule governs what a citation points *with*, so a coordinate quoted as data is outside it without needing to be named. No exception clause was added for them.

`investigation/`, issues, pull requests, and commit messages are outside the rule. `investigation/README.md` says a note "is not maintained afterwards", which is what makes a coordinate in one a record of where something was rather than a claim about where it is.

## The budget

`AGENTS.md` was at 24198 of 24198, so the rule had to be paid for.

| | bytes |
|---|---|
| before | 24198 |
| delete the `must_error` implementation's three load-bearing properties, which `inst/vignette-hooks/must-error.R` already states at the three sites holding them | −1056 |
| the rule | +322 |
| delete the *Dependency metadata* conclusion *Code comments* re-derived beside its own citation to it | −80 |
| **baseline** | **23384** |

The 80-byte deletion is the new rule applied to the paragraph that states it — a citation to *Dependency metadata* followed by a re-derivation of its conclusion.

The first commit *moved* the three properties into `inst/vignette-hooks/must-error.R`'s header. The review found that all three were already there, as inline comments on the code each is about, so the move made a second copy inside one file. The third commit deletes the header copy; the header now says only that a rewrite reads those sites. What the closure held was a third copy, so deleting it was right and relocating it was not.

## Review

Ran `mattpocock-skills:code-review` against `main...HEAD`. Three findings answered, three rejected with evidence; the record is in the review comment on this PR.

## Checks

`lintr::lint_package()` — no lints. `testthat::test_local()` under `NOT_CRAN=true` — no failures, warnings, or skips. `verify-context-budget.R` — 23384 within 23384.

### DIFF
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -24,12 +24,14 @@ source(".github/scripts/ci-helpers.R")
 # What the closure measured when it was last set, and the commit that set it.
 # Moving this means saying in the same commit what was added and what paid for
 # it.
-baseline_bytes <- 24198L
+baseline_bytes <- 23384L
 baseline_note <- paste(
-  "#295 deleting the arguments three sections restated and moving the rest to",
-  "`test-documentation.R` and `altdoc.yaml`, over #293's baseline; the 5957",
-  "bytes that removed paid for restoring the rule no file held, that no",
-  "package is promoted to Imports merely to make a check pass"
+  "moving the `must_error` implementation's three load-bearing properties to",
+  "`inst/vignette-hooks/must-error.R`, whose header had deferred outward for",
+  "facts about itself, over #295's baseline; the 1056 bytes that freed paid",
+  "for the rule that a citation names its target rather than a line number,",
+  "and for deleting the *Dependency metadata* conclusion *Code comments*",
+  "re-derived beside its own citation to it"
 )
 
 entry <- "CLAUDE.md"
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -26,12 +26,11 @@ source(".github/scripts/ci-helpers.R")
 # it.
 baseline_bytes <- 23384L
 baseline_note <- paste(
-  "moving the `must_error` implementation's three load-bearing properties to",
-  "`inst/vignette-hooks/must-error.R`, whose header had deferred outward for",
-  "facts about itself, over #295's baseline; the 1056 bytes that freed paid",
-  "for the rule that a citation names its target rather than a line number,",
-  "and for deleting the *Dependency metadata* conclusion *Code comments*",
-  "re-derived beside its own citation to it"
+  "#416 deleting the `must_error` implementation's three load-bearing",
+  "properties, which `inst/vignette-hooks/must-error.R` already stated at the",
+  "three sites holding them, over #295's baseline; 322 of the 1056 bytes that",
+  "removed paid for the rule that a citation names its target rather than a",
+  "line number"
 )
 
 entry <- "CLAUDE.md"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -344,6 +344,12 @@ compares the two, so an amendment leaves it standing and wrong. ADR 0023
 settled this once already, for a rule that governed one file while two others
 re-derived it.
 
+This holds wherever a maintained document points, not only in a comment: a
+citation names its target — a function, a heading, a chunk, an ADR or issue
+number — and not a line number, which an edit above it leaves pointing at a
+different line rather than at nothing. A SHA, and a permalink pinned to one,
+do not move.
+
 A comment describes the code it sits beside, in the present tense. What a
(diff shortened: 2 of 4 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` over `git diff main...HEAD`, read at `5c47cd7` against merge-base `9a44af9`. Two axes, parallel sub-agents. Six findings; three answered, three rejected. Answers are in `0cc3b4c`.

### Answered

**The relocation was the wrong move.** All three `must_error` properties were already stated in `inst/vignette-hooks/must-error.R` as inline comments on the code each is about — the `evaluate` wrapper, the guarded-chunk case inside it, and the restoration beside `after.knit` — and in more detail than the moved paragraph, naming `eval: !expr has_duckdb` and the child-mode guard. Moving `AGENTS.md`'s copy into the header made a second copy inside one file. The header now says only that a rewrite reads those sites. What the closure held was a third copy, so deleting it was right and relocating it was not.

**The `document` hook lost its citation.** It was the alternative to `after.knit` and was named only in the deleted paragraph. The `after.knit` comment now carries the citation to `investigation/restoring-knitr-hooks-a-vignette-installs.md`.

**`baseline_note` stated a saving as a spend.** It read "the 1056 bytes that freed paid for ... and for deleting the *Dependency metadata* conclusion". A deletion frees bytes. The note now says what was added and what paid for it, which is what `verify-context-budget.R`'s own header requires of a baseline move.

Separately, the PR body claimed "the section's next paragraph already names the file". The next paragraph names `.github/scripts/verify-must-error.R`; `inst/vignette-hooks/must-error.R` is named by the two *preceding* paragraphs. The outcome held, the reason did not; the body is corrected.

### Rejected, with evidence

**That the rule belongs outside *Code comments*, whose first line names `R/`.** `AGENTS.md` is read in full on every task, so a heading there is not a navigation aid, and the sentence states its own scope. The alternative home, `design/architecture.md`, is reached only from citations in `AGENTS.md` and from `design/agents/code-review.md` — during a review, not while prose is written.

**That the permalink carve-out is speculative because no permalink exists here.** Two do, at `vignettes/recipes.qmd`'s `[adr10]` and `[adr13]` link definitions. Both are branch-pinned, so a line fragment on either would rot exactly as the rule describes and a SHA-pinned one would not. The carve-out separates two forms that are each one edit away.

**That the rule under-implements the decision by naming a line number and not a positional description.** The four sites a wider rule would reach — `R/share.R`'s "falls to the line below", `R/grouping-plan.R`'s "the next line already turns", `test-documentation.R`'s "the `source()` line above it", `CONTEXT.md`'s "the first paragraph" — all point at something adjacent, read together with the comment that points. The rule exists for a reference that resolves elsewhere while nobody is looking.

### Verified by the review, not changed

The trimmed *hand audit* sentence is still true; *Dependency metadata* holds the scanner-expressiveness argument. Both rewritten citations resolve: the *Indirect `.fns`* bullet at `R/share.R:315` inside `share_of_parent()`'s roxygen, and `share_dependency[[1L]]` at `R/share.R:857` inside `plan_share_expressions()`. The scan was reproduced over 158 tracked files: no `#L` fragment anywhere, and the only surviving `file:NNN` hits are the five the rule does not reach. `verify-context-budget.R` is exact, not approximate — `AGENTS.md` 24187 → 23373, plus 11 for `CLAUDE.md`.

### Checks at `0cc3b4c`

`lintr::lint_package()` after `pkgload::load_all()` — no lints. `testthat::test_local()` under `NOT_CRAN=true` — no failures, warnings, or skips. `Rscript .github/scripts/verify-context-budget.R` — 23384 within 23384.

PR #419, opened 2026-09-04
Restate the caller's spelling in the native adapter's translation error

### LINKED ISSUE #411: Restate the caller's spelling in the native adapter's translation error

sayuks:
Implements #411 (decision: #410, ADR 0022's *Amendment: the native adapter's translation error is restated too*).

A caller on duckdb or postgres whose summary expression dbplyr cannot translate was quoted `total = +...`. The label is not truncated by length: `rewrite_grouping_dots()` splices in a SQL literal whose own deparse overflows the width `as_label()` deparses at, so any dot combining a contextual helper with anything else was unreadable. It now quotes what the caller wrote, as a local input already did.

## Changes

- `restate_argument_bullet()` captures the bullet's trailing period rather than requiring it, and rebuilds with what it captured. dbplyr's sentence carries none, so the reader matched nothing there even where the map's key already equalled the span.
- `summarize_margin_native()` takes the caller's labels, builds the map from its rewritten dots, and re-raises a restated error. Not `with_branch_conditions()`: the adapter summarizes once and repeats nothing. Errors only.
- The catch wraps both calls that hand dplyr the caller's expressions. The adapter also summarizes them ungrouped to learn the output names, and that call runs first — measured here, it is the one that raises. `check_summary_output_names()` stays outside it, its argument forced before it, so a Package condition is never reached by the restatement. ADR 0022's amendment is corrected to say this; it had said the catch was around the grouped `summarize()` alone.
- The native call site's comment claiming no condition is raised while the verb runs is gone. It was false when written.

## Tests

Both directions, through `dbplyr::simulate_postgres()`, which needs no optional backend: the restoration, and the label two differently spelled unnamed dots collapse to, which `branch_argument_map()` drops so dplyr's own quotation stands. The seam test gains the periodless bullet.

Verified additionally against a live duckdb connection.

Closes #411.

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -503,10 +503,15 @@ restate_argument_lines <- function(text, restatements, rendered) {
 # the name of the message vector rather than part of its text. The span is read
 # to the last backtick on the line, because a non-syntactic name the caller
 # wrote puts backticks inside it.
+#
+# The trailing period is captured rather than required, and put back as it was
+# found: eager dplyr ends the sentence with one and dbplyr does not, and the
+# sentence around the span belongs to whichever raised the condition
+# (ADR 0022).
 restate_argument_bullet <- function(line, restatements) {
   found <- regmatches(
     line,
-    regexec("^([i\u2139] )?In argument: `(.*)`\\.$", line)
+    regexec("^([i\u2139] )?In argument: `(.*)`(\\.)?$", line)
   )[[1L]]
   if (length(found) == 0L) {
     return(line)
@@ -515,7 +520,7 @@ restate_argument_bullet <- function(line, restatements) {
   if (is.na(restored)) {
     return(line)
   }
-  paste0(found[[2L]], "In argument: `", restored, "`.")
+  paste0(found[[2L]], "In argument: `", restored, "`", found[[4L]])
 }
 
 # Rewrites the internal grouping-column names dplyr built the context from into
--- a/R/grouping-adapter-native.R
+++ b/R/grouping-adapter-native.R
@@ -1,8 +1,14 @@
+# `caller_labels` is `new_summary_arguments()`'s, parallel to the `dots` this
+# is handed, and is what the translation error below is restated with. It is
+# required rather than defaulted, because a default computed here would be
+# computed from dots this has already rewritten, and would restate a spelling
+# the caller never wrote (ADR 0024).
 summarize_margin_native <- function(.data,
                                     dots,
                                     plan,
                                     margin_labels,
                                     reserved_names,
+                                    caller_labels,
                                     set_id_name = NULL,
                                     set_id_is_internal = FALSE) {
   con <- dbplyr::remote_con(.data)
@@ -12,6 +18,9 @@ summarize_margin_native <- function(.data,
     sql = TRUE,
     con = con
   )
+  # After the rewrite, because the rewritten dots are the expressions dplyr
+  # will quote.
+  restatements <- branch_argument_map(dots, caller_labels)
   group_vars <- unique(c(plan$by, plan$dimensions))
   if (!is.null(set_id_name)) {
     set_id_quo <- rlang::new_quosure(
@@ -50,23 +59,34 @@ summarize_margin_native <- function(.data,
     flag_quos <- list()
   }
 
-  check_summary_output_names(
+  # Forced here rather than left to lazy evaluation inside the check, so that
+  # the translation error a summarize of the caller's expressions can raise is
+  # restated and the check's own Package conditions are not reached by the
+  # restatement at all.
+  output_names <- with_native_restatement(
     native_summary_output_names(.data, dots),
+    restatements
+  )
+  check_summary_output_names(
+    output_names,
     group_vars = group_vars,
     internal_names = flag_names,
     set_id_name = set_id_name,
     set_id_is_internal = set_id_is_internal
   )
 
-  result <- dplyr::summarize(
-    .data = dplyr::group_by(
-      .data,
-      dplyr::pick(dplyr::all_of(group_vars))
+  result <- with_native_restatement(
+    dplyr::summarize(
+      .data = dplyr::group_by(
+        .data,
+        dplyr::pick(dplyr::all_of(group_vars))
+      ),
+      !!!dots,
+      !!!set_id_quos,
+      !!!flag_quos,
+      .groups = "drop"
     ),
-    !!!dots,
-    !!!set_id_quos,
-    !!!flag_quos,
-    .groups = "drop"
+    restatements
   )
 
   result <- attach_grouping_sets_query(result, plan$sets)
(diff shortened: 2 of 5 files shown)


sayuks:
## Review record — `mattpocock-skills:code-review`, one round

Fixed point: merge-base with `main` (`14686c2`), snapshot reviewed `e280528`. Answered in `fe35653`.

### Standards — 7 findings

| finding | disposition |
|---|---|
| `with_native_restatement()`'s header re-derives ADR 0022's *Amendment* beside its citation — the second copy *Code comments* forbids, and the defect the commit removed at the call site | fixed in code: the helper is gone |
| Same, for the sentence about both calls being wrapped | fixed in code |
| ADR's "Nothing else about the adapter moves" survived unedited while `output_names` was lifted out of the check's lazy argument | fixed in prose |
| `caller_labels`'s header near-restates `summarize_margin_branch()`'s | fixed in code: `summaries` is taken whole, so there is nothing to restate |
| Both new tests inline their reproduction and carry the same `# nolint` rationale, where every other reproduction in the file is a named fixture above the tests | fixed in code |
| `with_native_restatement()` is condition machinery in an adapter file | fixed in code: gone |
| Its header describes its call sites rather than its contract | fixed in code: gone |

### Spec — 6 findings

| finding | disposition |
|---|---|
| The ticket names one catch site; two were wrapped. The grouped `summarize()` never raises, and no dot was found that translates ungrouped and fails grouped, so that site was reached by no test | fixed in code: the second catch goes. The gate the defensive argument wanted is the end-to-end test, which fails if dbplyr ever moves the translation |
| ADR asserts a measurement `what-truncates-an-argument-label-on-the-native-adapter.md` does not hold | fixed in prose |
| Same paragraph called the ungrouped summarize "a second time" while it runs first | fixed in prose |
| `caller_labels` split a pair `stage_margin_summaries()`'s header says travels as one value | fixed in code |
| ADR 0024 cited where ADR 0022 governs | fixed in code: the citation is gone with the header |
| The union-path period regression is asserted at the seam but not on the path the ticket named | fixed in code: `a branch error quotes the caller's own spelling` now matches through dplyr's trailing period |

Correct as specified and unchanged: the regex in both directions, the map built from the rewritten dots, errors only, nothing from *Not in scope*, and change 3's replacement comment.

Full suite and `lintr::lint_package()` green after the answers; the restoration re-verified against a live duckdb connection.

PR #420, opened 2026-09-04
Restore an NA factor level on every column crossing the union (#415)

### LINKED ISSUE #415: dtplyr backend drops a declared NA factor level from a column that is not a Margin dimension
## Summary

On a dtplyr input, a factor column carrying a declared `NA` level loses it whenever
it crosses the branch union without being a Margin dimension — as a fixed `.by` key,
or as a column the verb passes through. The same call on the same data through the
local `data.frame` backend keeps it, so the result's column domain differs by backend
with no error, warning, or other diagnostic.

This is the residual of #408, which fixed the dimension case only. It was found by
the review of that branch (#413) and left as a ticket rather than widened into it,
because #408's Summary and Implementation site are both about a Margin *dimension*.

## Reproduction

```r

data <- data.frame(
  key = structure(c(1L, 2L), levels = c("p", NA_character_), class = "factor"),
  passthrough = structure(c(1L, 2L), levels = c("a", NA_character_), class = "factor"),
  value = 1:2
)

fixed_key <- function(input) {
  summarize_with_margins(
    input, n = dplyr::n(), .by = key, .grouping = rollup(group)
  )
}
cat("fixed .by key -- local:  "); print(levels(fixed_key(data)$key))
cat("fixed .by key -- dtplyr: ")
print(levels(dplyr::collect(fixed_key(dtplyr::lazy_dt(data)))$key))

passthrough <- function(input) {
  expand_with_margins(input, .grouping = rollup(group))
}
cat("passthrough  -- local:  "); print(levels(passthrough(data)$passthrough))
cat("passthrough  -- dtplyr: ")
print(levels(dplyr::collect(passthrough(dtplyr::lazy_dt(data)))$passthrough))
```

## Observed / Expected

Observed, verbatim, at the head of the #408 branch:

```
fixed .by key -- local:  [1] "p" NA 
fixed .by key -- dtplyr: [1] "p"
passthrough  -- local:  [1] "a" NA 
passthrough  -- dtplyr: [1] "a"
```

Expected: each dtplyr line is `[1] "p" NA` and `[1] "a" NA` respectively, as its local
line is. Nothing else in the output changes.

## Contract violated

ADR 0016, *The promise boundary*:

> marginplyr constructs, and therefore promises:
>
> [...]
> - factor and ordered-factor levels, ordering, and Margin-label placement, per
>   ADR 0012.

That clause is not qualified to Margin dimensions, and neither is ADR 0012's sixth row.
Whether it should be is the first question this ticket has to answer — a `.by` key and a
passed-through column are columns marginplyr does not label, and ADR 0016 elsewhere
separates what it constructs from what it passes through. If the boundary places them on
the passed-through side, the answer is an amendment saying so rather than code.

## Root cause

The same one as #408: `data.table`'s rbind drops a declared `NA` level from the result
and turns the values that used it into missing codes, for a list of one table as much as
for a union of branches.

```r
f <- structure(c(1L, 1L), levels = c("x", NA_character_), class = "factor")
a <- data.table::data.table(group = f)
levels(data.table::rbindlist(list(a))$group)
#> [1] "x"
```

#408 answered it by routing a factor *dimension* whose Margin label is missing through
the same encode-and-rebuild path a labelled one already took — the
`drops_na_factor_level_on_union` capability in `backend_capabilities()`, read per column
as `encode_missing_label` in `margin_column_info()`. That route is reached from
`label_margin_branch()` and `restore_margin_factors()`, both of which iterate
`plan$dimensions` only.

## What is unmeasured

- Whether a passed-through column can be routed at all. `margin_column_info()` reads the
  selection proxy for `plan$dimensions`; a passthrough column's levels are not acquired
  today, and ADR 0002 (*acquire typed metadata once*) is where widening that read would
  have to be argued.
- The cost of acquiring and restoring levels for every factor column of a wide input,
  against acquiring them for the dimensions alone.
- Whether the ordered-factor and Margin-label-position rules have anything to say about a
  column that carries no Margin label.

sayuks:
Closes #415.

A factor column carrying a declared `NA` level lost it on a dtplyr input whenever it crossed the branch union without being a Margin dimension — as a fixed `.by` key, or as a column the verb passes through — and the values that used the level became missing. #408 fixed the dimension case; `label_margin_branch()` and `restore_margin_factors()` both reached its encode-and-rebuild route from the plan's dimensions alone.

`margin_column_info()` now takes `carried`, the input columns the verb's result holds beside the dimensions, and returns `factor_info` for those too. `prototypes` stays keyed to the dimensions, which are the only columns a branch omits. No query is added: the selection proxy already holds every column for each kind that can restore factors, so this is a second read of the one snapshot ADR 0002 acquired.

Each verb states its own set rather than a word naming its kind, because only a column the result holds can be rebuilt after the union: `expand_with_margins()` carries every column, while the summary and nest verbs carry the fixed `.by` keys — a summary output may reuse an input column's name, and a nested payload column is inside a cell by the time the finalizer runs.

`margin_label_of()` is what lets the widened set be read against `margin_labels`, which stays keyed to the dimensions: `[[` on an absent name raises a subscript error rather than answering that the column is unlabelled.

## Residual

A dtplyr `nest_with_margins()` payload column still loses a declared `NA` level inside its cell. That is the same union, but the column is inside `.key` by the time the finalizer runs and nothing rebuilds it, so it is outside this ticket's Reproduction and Acceptance. Filed as #421 rather than widened into here.

## Verification

- The issue's Reproduction prints `[1] "p" NA` and `[1] "a" NA` on both backends, with `is.na()` agreeing.
- `testthat::test_local()` green with dtplyr, duckdb, arrow, data.table, DBI and tidyr installed; `test-query-policy.R`'s snapshots unchanged.
- `lintr::lint_package()` clean.

### DIFF
--- a/R/backend-metadata.R
+++ b/R/backend-metadata.R
@@ -49,8 +49,8 @@ grouping_selection_proxy <- function(.data,
 # rather than anything a caller can rewrite their way out of -- either a proxy
 # that does not answer for its own columns, or a subclass whose `[[` is not
 # column extraction. It is still worth stopping on, and stopping bare: silence
-# here would report the dimension as an absent prototype and label its margin
-# rows `NA` instead.
+# here would report a column with no levels and no prototype as one the input
+# declared that way.
 proxy_columns <- function(data_proxy, cols) {
   columns <- stats::setNames(
     lapply(cols, function(col) data_proxy[[col]]),
@@ -72,9 +72,8 @@ proxy_columns <- function(data_proxy, cols) {
 # `carried` names the input columns that reach the result beside the Margin
 # dimensions -- a fixed `.by` key, or a column the verb passes through. They
 # cross the same branch union and lose a declared NA level in the same place,
-# so `factors` covers them too (#415). Each verb states its own set, because
-# what a result carries is a property of the verb: only a column present in
-# the result can be rebuilt after the union.
+# so `factors` covers them too (#415). Which columns those are is settled by
+# `prepare_margin_operation()`'s `carried_columns`.
 #
 # `prototypes` stays keyed to the dimensions alone. It stands for the value a
 # branch omitting a dimension writes, and only a dimension is ever omitted.
--- a/R/backend-metadata.R
+++ b/R/backend-metadata.R
@@ -51,12 +51,12 @@ grouping_selection_proxy <- function(.data,
 # column extraction. It is still worth stopping on, and stopping bare: silence
 # here would report the dimension as an absent prototype and label its margin
 # rows `NA` instead.
-proxy_columns <- function(data_proxy, dimensions) {
+proxy_columns <- function(data_proxy, cols) {
   columns <- stats::setNames(
-    lapply(dimensions, function(col) data_proxy[[col]]),
-    dimensions
+    lapply(cols, function(col) data_proxy[[col]]),
+    cols
   )
-  absent <- dimensions[vapply(columns, is.null, logical(1))]
+  absent <- cols[vapply(columns, is.null, logical(1))]
   if (length(absent) > 0L) {
     stop(
       "The selection proxy has no column",
@@ -69,10 +69,25 @@ proxy_columns <- function(data_proxy, dimensions) {
   columns
 }
 
+# `carried` names the input columns that reach the result beside the Margin
+# dimensions -- a fixed `.by` key, or a column the verb passes through. They
+# cross the same branch union and lose a declared NA level in the same place,
+# so `factors` covers them too (#415). Each verb states its own set, because
+# what a result carries is a property of the verb: only a column present in
+# the result can be rebuilt after the union.
+#
+# `prototypes` stays keyed to the dimensions alone. It stands for the value a
+# branch omitting a dimension writes, and only a dimension is ever omitted.
+#
+# No query is added. The selection proxy already holds every column for each
+# kind that can restore factors, so reading more of it is a second read of the
+# one snapshot ADR 0002 acquired rather than a second acquisition.
 margin_column_info <- function(data_proxy,
                                dimensions,
-                               backend) {
-  if (length(dimensions) == 0L) {
+                               backend,
+                               carried = character()) {
+  read <- unique(c(dimensions, carried))
+  if (length(read) == 0L) {
     return(list(factors = list(), prototypes = list()))
   }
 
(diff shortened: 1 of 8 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel, against the merge-base with `main` (`842f156`). Two rounds.

### Round 1 — over `842f156...7e26f4e`

**Spec:** eight acceptance criteria, all satisfied; no scope creep; no wrong implementation. One qualification, which the Standards axis raised independently as finding 3 below: criterion 4 (*"A factor column with no `NA` level takes no new route"*) held in code but had no such column in the new test's fixture, so the criterion rested on a manual check. The axis also read #421 and judged the nest deferral defensible.

**Standards:** three hard findings, three judgement calls.

| finding | disposition |
|---|---|
| `proxy_columns()`'s header names a dimension's consequence — an absent prototype, margin rows labelled `NA` — for a function that now reads carried columns, which have neither | fixed in prose, `62a606b` |
| `margin_column_info()`'s header re-derives why each verb states its own carried set, an argument `prepare_margin_operation()` owns — the second copy *Code comments* refuses | fixed in prose, `62a606b`; it cites `carried_columns` by name |
| the test named for a carried factor with no `NA` level asserts `encode_missing_label` false only for the Margin dimension | fixed in code, `62a606b`; `plain` is that column and `expect_false(encode[["plain"]])` fails without the `has_na_in_level` term |
| judgement: `expect_carried_factor_agrees(result, expected, col)` takes a column argument while hardcoding one fixture's levels | fixed in code, `62a606b`; it is `expect_passthrough_agrees(result, expected)` |
| judgement: `margin_column_info()`'s header is a 13-line block carrying three arguments | answered by the second finding's fix, which removed one of them |
| judgement: `carried_columns(data_vars, plan)` — `data_vars` is read by one of three implementations, and two are identical | rejected. `expand_with_margins()` cannot compute `data_vars` itself: the names are resolved inside `prepare_margin_operation()`, after `normalize_grouping_input()`. The two identical bodies carry different reasons, which their comments state at each site |

### Round 2 — over `7e26f4e..62a606b`

Round 1's answers were prose and a fixture, so this round read them rather than re-reviewing the branch, per *Running it a second time* in `design/agents/code-review.md`. It confirmed all four answers land and that `plain` made no existing assertion vacuous — `expand_with_margins()` still returns four rows and the `key`, `passthrough`, and `group` assertions are untouched.

One finding, introduced by the answer: the comment beside the new assertions claimed the dimension was there to say that being a dimension is not what decides the route. It cannot — `group` has no `NA` level, so its assertion is the same case as `plain`'s. Deleted rather than restated (`40697b1`).

PR #422, opened 2026-09-04
Verify that a path a maintained document names exists (#417)

### LINKED ISSUE #417: Verify that a path a maintained document names exists
`AGENTS.md`'s *Code comments* now requires a citation to name its target rather than a coordinate (#416). A name fails loudly — a grep for it returns nothing — but nothing in this repository runs that grep, so a path that a document names and that no longer exists stays in place reading as though it resolved.

## What it would check

For every tracked source that is not generated, every repository-relative path written in prose exists.

## What it costs

Measured on `main` at the time of writing:

- Dead references: **0**. The gate is green on the day it lands.
- The only apparent hits are generated or external: `vignettes/*.html` and `man/*.html`, which are the site URLs `R/*.R` roxygen links to and the page names `verify-site.R` derives. Those are the exclusions.
- `design/agents/code-review.md` names `design/review-dispositions.md`, which #288 retired. That reference is in the past tense and is a correct historical statement, so a scan has to read it as one — either by scoping to present-tense claims or by accepting that one site needs a suppression with a stated reason.
- `AGENTS.md` bytes: **0**. A verifier's own header states its mechanism, and `verify-verifier-invocation.R` derives the set from the directory, so nothing has to be added to the always-loaded closure.

## Why it is not the line-number rule

The rule added in #416 governs what a citation points *with*. This gate checks whether one of the permitted forms still resolves. A scan for line numbers is a separate and narrower question — `\.R:[0-9]+` over the same set — and is not part of this.

## Where it goes

`.github/scripts/verify-doc-references.R`, reached by a step in `lint.yaml`.

sayuks:
Closes #417.

`AGENTS.md`'s *Code comments* requires a citation to name its target rather than a coordinate, on the argument that a name fails loudly: a grep for it returns nothing. Nothing here ran that grep, so a path a document named and that was since renamed or deleted stayed in place reading as though it resolved.

`.github/scripts/verify-doc-references.R` runs it, from a `doc-references` job in `lint.yaml` — beside `context-budget` and for its reason, since it needs no package installed.

## What it reads, and what it does not

- **Prose, not code.** Markdown throughout; comment text in `.R` and YAML. A path in a string literal is resolved by the interpreter and fails at run time, and `verify-verifier-invocation.R` already reports the `source()` case. A path in a comment resolves for nobody.
- **Maintained documents.** `investigation/` notes are outside the set: they are not maintained after their date, and they describe other packages' source trees, so an `R/`-rooted path in one is rlang's or arrow's and no scan can tell that from ours. Their own file names stay covered from the side of every document that cites one. `investigation/README.md` states the convention rather than recording a note, so it is read.
- **Generated files.** `README.md` is read through `README.Rmd`; `man/` and `NAMESPACE` carry no extension this reads.
- **Paths, not bare names.** A candidate has to contain a `/`. Matching a bare `CHANGELOG.md` or `cnd-last.R` against this repository's base names reports prose about another repository as a dead reference — measured, it returns thirteen such hits and no real one.
- **Files, not directories.** Prose spells an alternative `a/b` the way a directory path is spelled, so a trailing-slash match returns `and/or`, `Config/Needs/`, and `filter/summarise/` before it returns a directory.

A candidate's extension has to be one some tracked file carries. That set is derived rather than listed, which is what excludes the `.html` paths the ticket named as the only apparent hits on `main`: no tracked file is `.html`, so an `.html` path never names one. Resolution is against the tracked set rather than the working tree — which carries `docs/` and other build output that varies by what was last run — either repository-relative or relative to the document naming it, since a markdown link resolves the second way. A `*` is matched as a glob.

## The two exemptions

Both are in `design/agents/code-review.md` and both are correct as written, so each sits in `exempt` with the reason:

- `design/review-dispositions.md` — #288 retired it, and the sentence records that it has no successor.
- `docs/agents/issue-tracker.md` — the left column of a table of what the skill looks for, so the row exists to say this repository has no such file.

The check fails in both directions when an exemption goes stale — the path came back, or the document stopped naming it — so it cannot outlive what it excuses.

## Verification

Green on this branch: 172 paths across 150 maintained documents. Three failure modes were exercised locally and each reported:

- a dead reference injected into `CONTEXT.md`,
- an exemption pointed at a path that resolves,
- the candidate reader stubbed out, which the probe catches before any verdict.

`lintr::lint_dir(".github", ...)` clean, `verify-verifier-invocation.R` lists the new script as reached by a `lint.yaml` step, `verify-context-budget.R` unchanged at 23384 bytes, and the full suite passes.

### DIFF
--- a/.github/scripts/verify-doc-references.R
+++ b/.github/scripts/verify-doc-references.R
@@ -1,0 +1,280 @@
+# Fails when a maintained document names a repository path no file answers.
+#
+# `AGENTS.md`'s *Code comments* requires a citation to name its target rather
+# than a coordinate, because a name fails loudly: a grep for it returns
+# nothing. Nothing in this repository ran that grep, so a path a document names
+# and that was since renamed or deleted stayed in place reading as though it
+# resolved -- which is the failure every gate here exists to refuse, a
+# reference that asserts nothing and reports nothing (#417).
+#
+# What is read, and why each bound is where it is:
+#
+# * Prose, not code. A path in a string literal is resolved by the interpreter
+#   and fails at run time when it is wrong; `verify-verifier-invocation.R`
+#   already reports the `source()` case. A path in a comment resolves for
+#   nobody, so nothing but this fails on it. Markdown is prose throughout; in
+#   `.R` and YAML it is the comment text.
+# * Maintained documents. `investigation/` notes are not maintained after their
+#   date (`investigation/README.md`), and they describe other packages' source
+#   trees, so an `R/`-rooted path in a note is rlang's or arrow's and no scan
+#   can tell that from ours. Their own file names are still covered, from the
+#   side of every document that cites one. `investigation/README.md` states
+#   the convention rather than recording a note, so it is read.
+# * Generated files. `README.md` is rendered from `README.Rmd`, which is read
+#   in its place; `man/` and `NAMESPACE` are roxygen output and carry no
+#   extension this reads.
+# * Paths, not bare names. A candidate has to contain a `/`. A bare
+#   `CHANGELOG.md` or `cnd-last.R` names a file in whatever tree the sentence
+#   is about, and matching it against this repository's base names reports
+#   prose about another repository as a dead reference.
+# * Files, not directories. Prose spells an alternative `a/b` the way a
+#   directory path is spelled, so a trailing-slash match is noise rather than a
+#   weaker signal: on this repository it returns `and/or`, `Config/Needs/`, and
+#   `filter/summarise/` before it returns a directory. A renamed directory is
+#   caught through the files under it that anything cites.
+#
+# A candidate's extension has to be one some tracked file carries. That is
+# derived rather than listed, and it is what excludes the `.html` paths, which
+# are the site URLs `R/*.R` roxygen links to and the rendered vignettes
+# `cran-comments.md` says the tarball ships: no tracked file is `.html`, so an
+# `.html` path never names one.
+#
+# A candidate resolves against the tracked set -- not the working tree, which
+# carries `docs/` and other build output that varies by what was last run --
+# either repository-relative or relative to the document naming it, since a
+# markdown link resolves the second way. A `*` is matched as a glob.
+#
+# Run it locally with:
+#
+#     Rscript .github/scripts/verify-doc-references.R
+
+source(".github/scripts/ci-helpers.R")
+
+# Paths a document names deliberately and that no file answers. Each entry
+# states why the reference is correct as written; the check below fails when
+# one goes stale, in either direction, so this cannot outlive what it excuses.
+exempt <- data.frame(
+  file = c(
+    "design/agents/code-review.md",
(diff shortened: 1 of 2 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` against `git diff main...HEAD`, base `842f156`, subject `afd1bbf`. Two axes, run as parallel sub-agents. Answered in `a448ee1`.

### Standards — 5 findings, all answered

| finding | disposition |
|---|---|
| The probe fed strings straight to `candidates()` and never ran the comment extraction, which supplies 66 of the 173 paths read; and nothing floored the document set, so a broken `prose_kinds` reported "0 across 0" and passed | fixed in code: `prose_of()` takes lines, three probes run it over a markdown, an R, and a YAML document, and the count is floored at two documents and at one path found |
| `_quarto/` was a third, unargued filter excluding zero files, under a comment claiming the exclusions are the two the header argues for | fixed in code: deleted, comment corrected |
| "the derived extension set … excludes the `.html` paths, which are the site URLs `R/*.R` roxygen links to" — those are stripped as URLs before any extension is matched | prose false, deleted; only the `cran-comments.md` half remains |
| "`verify-verifier-invocation.R` already reports the `source()` case" — it reports a `source()` under `.github/scripts/`, not the general case | prose false, deleted |
| the trailing-slash claim: `and/or` carries no trailing slash, and real directories come out at the top | prose false, deleted. Re-measured on `842f156`: 56 references answering to no tracked directory, none of them dead. That is the reason now stated |

Correctness, same round: `flatten()` returned a path for `../../outside/x.md` by dropping the `..` that walks past the root, and now returns `NA`; `utils::glob2rx()` rendered `*` as `.*`, so `man/*.Rd` answered to a file under a directory of `man/`, replaced by `glob_pattern()`; R comments now come from `getParseData()` rather than the first `#` on the line, which is what the header's "Prose, not code" bound claims and what `grepl("#", y)` broke; `strip_noise()`'s comment no longer calls its second substitution a format conversion alone, since it eats `%in%` too.

Smells (judgement, both accepted): *Primitive Obsession / Data Clumps* and *Duplicated Code* around `dead` as unnamed 2-vectors — `dead` is now a data frame with `exempt`'s two columns, so the staleness check and the filter share one key and the summary and diagnostic read columns.

### Spec — 1 finding, answered

The three narrowings past the spec's own exclusion set were measured. `investigation/` (41 hits, all other packages' trees) and bare names (17 unresolved, all external) hold as stated. The directory argument did not, and is the fifth Standards row above. No scope creep found; the line-number scan is correctly absent, per "*is not part of this*". The second exemption was confirmed legitimate and its both-direction staleness check verified to fire.

### What now asserts it

Ten mutations were run against the rewrite, each reported: a dead reference injected into `CONTEXT.md`; an exemption pointed at a path that resolves; `candidates()` stubbed; the R comment reader stubbed; the YAML comment reader stubbed; the R reader widened to read code; URL stripping removed; `prose_kinds` narrowed; the extension set emptied — and the restored file green at 173 paths across 150 documents.

PR #423, opened 2026-09-04
Decide which file owns why must_error exists (#418)

### LINKED ISSUE #418: Decide which file owns why must_error exists
The first sentence of *Chunks that must fail* in `AGENTS.md` is present verbatim in `inst/vignette-hooks/must-error.R`'s header:

> A vignette showing a rejected call executes it rather than quoting its error, so the reader sees the diagnostic their own session would produce.

Both then argue that Quarto's `error: true` is the wrong option on its own, in different words.

#416 moved the implementation's three load-bearing properties out of `AGENTS.md` into that header, on the grounds that the paragraph named an audience that is reading the file rather than every task. This duplication was measured in the same pass and left alone, because it is not the same case:

- `AGENTS.md` has to say **when to mark a chunk**, which fires for anyone writing a vignette, and stating that needs some of the why.
- `AGENTS.md`'s version carries two things the header does not: that knitr and Quarto offer no option for the other half and no package supplies one, and the citation to `investigation/requiring-a-documentation-chunk-to-fail.md`.

So the question is not whether to delete one copy but which file owns the argument and what the other keeps. The always-loaded closure is at 23384 bytes and the paragraph is 507, so the answer is worth having either way.

sayuks:
Closes #418.

`AGENTS.md` owns why the option exists. Everything about how a file works stays in that file.

## What the ticket named, and what measuring found

#418 named one duplication: the first sentence of *Chunks that must fail*, verbatim in `inst/vignette-hooks/must-error.R`'s header. Measuring found more.

The sentence is in **five** places, not two — `AGENTS.md`, the hook, and the `include: false` setup chunk of all three vignettes. Four of the five already name `AGENTS.md` as authoritative, and the two that disagree disagree about scope: the vignettes say `AGENTS.md` owns "what the option does and why it exists", while the hook says it owns only "when to mark a chunk" and then states the why itself.

The same section holds two further pairs the ticket did not name:

| `AGENTS.md` | already held by |
|---|---|
| the gate paragraph and the fixture paragraph | `verify-must-error.R`'s header and `:275-281`; the `--ignore-vignettes` clause by `release-matrix.yaml`'s header |
| the `parent`-chain reason and the malformed-value reason | `must-error.R:36-38`, `:59-64`, and `verify-must-error.R:253-254` |

A seven-word shingle scan of the whole section against the six files that touch the option confirms the inventory is complete.

## The rule applied

An author-facing rule stays in `AGENTS.md`, because it fires for anyone writing a vignette. Reasoning about how a file works stays in that file, because its audience is reading it.

That points both ways, and the split is where the two directions land:

- **Why the option exists** is what a vignette author needs before choosing to mark a chunk. `AGENTS.md` keeps it, with the citation to `investigation/requiring-a-documentation-chunk-to-fail.md`. `must-error.R` drops its copy.
- **Why the gate is a script and not a testthat test** is read by whoever opens the verifier. `verify-must-error.R` keeps it. `AGENTS.md` keeps the pointer and the local command. The general rule this derives from is already in *Release matrix*, so the derivation was the second copy.

Two sites were checked against the rule and **kept**:

- `altdoc.yaml:54-56` borrows the "reports nothing" premise, but its own decision is that the step runs *before* the render, and "failing here says which half broke" is unsupported without it. The verifier's header says nothing about step ordering.
- `AGENTS.md`'s placement paragraph cites *Dependency metadata* and then says the quarto → rmarkdown → knitr chain is which way this site falls under it. `investigation/README.md` says a note "is not maintained afterwards" and "never describes the present", so a live claim cannot be delegated to one — unlike a verifier header.

## What changed

- `AGENTS.md` — gate and fixture paragraphs reduced to the pointer; two reasons in the two-forms paragraph reduced to the facts.
- `inst/vignette-hooks/must-error.R` — header drops the why and the `inst/` placement argument, whose vignette list named two of the three documents that hold the call. Delegation line widened to "what the option is for and when to mark a chunk".
- `vignettes/{recipes,get_started,database_backends}.qmd` — setup comments state what the `source()` call installs rather than restating the argument they delegate.
- `tests/testthat/test-documentation.R` — cited the malformed-header reason and then restated it with "field" substituted for "chunk header". The restatement is deleted; the citation stays.
- `.github/scripts/verify-context-budget.R` — baseline ratcheted.

No rule was added: *Code comments* already forbids re-deriving a cited decision's argument, and #416 settled that it reaches `AGENTS.md`'s own prose.

## The budget

| | bytes |
|---|---|
| before | 23384 |
| the gate's argument and its fixture paragraph | -1076 |
| the two reasons stated beside the code each is about | -228 |
| a claim the review found false, and one more second copy | -75 |
| **baseline** | **22005** |

Nothing was added, so the ratchet is the whole of it. Note that #418 as scoped saves nothing: the 507-byte paragraph it named stays, and the savings all come from the sites measuring found.

## Review

Ran `mattpocock-skills:code-review` against `origin/main...e494d29`. Two findings, both answered by deleting in `adfbba9`; none rejected. The record is in the review comment on this pull request.

## Checks

| | |
|---|---|
| `verify-context-budget.R` | 22005 of 22005 |
| `verify-must-error.R` | 8 fixtures |
| `verify-doc-references.R` | 171 paths across 150 documents |
| `verify-verifier-invocation.R` | 9 verifiers |
| `lintr::lint_package()` | no lints |
| `test-documentation.R` | 21 pass |

`docs/` and `README.md` are unaffected: every vignette edit is a comment inside an `include: false` chunk.

### DIFF
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -24,13 +24,12 @@ source(".github/scripts/ci-helpers.R")
 # What the closure measured when it was last set, and the commit that set it.
 # Moving this means saying in the same commit what was added and what paid for
 # it.
-baseline_bytes <- 23384L
+baseline_bytes <- 22080L
 baseline_note <- paste(
-  "#416 deleting the `must_error` implementation's three load-bearing",
-  "properties, which `inst/vignette-hooks/must-error.R` already stated at the",
-  "three sites holding them, over #295's baseline; 322 of the 1056 bytes that",
-  "removed paid for the rule that a citation names its target rather than a",
-  "line number"
+  "#418 deleting the gate's argument from *Chunks that must fail*, which",
+  "`verify-must-error.R`'s header and `release-matrix.yaml`'s already held,",
+  "together with the two reasons `inst/vignette-hooks/must-error.R` states",
+  "beside the code each is about, over #416's baseline; nothing was added"
 )
 
 entry <- "CLAUDE.md"
--- a/.github/scripts/verify-context-budget.R
+++ b/.github/scripts/verify-context-budget.R
@@ -24,12 +24,13 @@ source(".github/scripts/ci-helpers.R")
 # What the closure measured when it was last set, and the commit that set it.
 # Moving this means saying in the same commit what was added and what paid for
 # it.
-baseline_bytes <- 22080L
+baseline_bytes <- 22005L
 baseline_note <- paste(
   "#418 deleting the gate's argument from *Chunks that must fail*, which",
   "`verify-must-error.R`'s header and `release-matrix.yaml`'s already held,",
-  "together with the two reasons `inst/vignette-hooks/must-error.R` states",
-  "beside the code each is about, over #416's baseline; nothing was added"
+  "together with the reasons `inst/vignette-hooks/must-error.R` states beside",
+  "the code each is about and a claim its review found false, over #416's",
+  "baseline; nothing was added"
 )
 
 entry <- "CLAUDE.md"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -94,10 +94,10 @@ offer no option for the other half, and no package supplies one
 (`investigation/requiring-a-documentation-chunk-to-fail.md`).
 
 `inst/vignette-hooks/must-error.R` therefore defines a `must_error` chunk
-option. It implies `error: true`, so the two are never set inconsistently, and
-it halts the render naming the chunk when a chunk marked with it completes
-without raising an error. Mark a chunk with it instead of `error: true`
-whenever the surrounding prose asserts that the call fails.
+option. It implies `error: true`, and it halts the render naming the chunk when
+a chunk marked with it completes without raising an error. Mark a chunk with it
+instead of `error: true` whenever the surrounding prose asserts that the call
+fails.
 
 It takes two forms. `must_error: true` accepts any error, which is what the
 option has always meant. `must_error: marginplyr_error` additionally requires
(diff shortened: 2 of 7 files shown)


sayuks:
## Review

Ran `mattpocock-skills:code-review` against `origin/main...e494d29`. Local `main` was two merges behind, so the fixed point is `origin/main` (`e653f31`); reviewing against the stale ref would have reported #415 and #417 as this branch's.

Two findings, both answered in `adfbba9` by deleting.

### 1. `AGENTS.md`: "Any other value halts the render." is false

`expected_class()` maps `FALSE` to `NULL` alongside an absent option, so `must_error: false` reads as unmarked. Knitted a fixture with that header on a chunk raising no error: the render completes.

The sentence generalised over a paragraph naming two forms, so `false` fell under "any other value". `must-error.R`'s own comment enumerates the not-marked case first and is right. Deleted rather than corrected — *Answering a review* has a claim found false deleted rather than restated, and the safety property it gestured at is stated correctly beside the code.

The sentence is not new wording of a true claim: the same generalisation was in `main`. This branch rewrote the sentence, so it is this branch's to answer.

### 2. `AGENTS.md`: "so the two are never set inconsistently" is a second copy

The same claim and the same reason sit on `must-error.R`'s `opts_hooks` entry that does the implying. This is the shape the branch removes everywhere else. It had been kept on the grounds that a reader could not otherwise tell whether the implication is desirable — which is how the prose reads, and *Answering a review* says that is not a finding either way, so it was not a reason to keep it. Deleted; "It implies `error: true`" stays.

### Rejected: none

The Spec axis raised no missing or partial requirement and no incorrect implementation. Its scope observation — that five of the seven files touched are not named by #418, and that the 507-byte paragraph the issue named was kept — is what the PR body already states: the named duplication saves nothing, and every byte comes from sites measuring found. The Standards axis flagged the identical setup comment in all three vignettes as possible Duplicated Code and suppressed it itself: nothing shares a comment across setup chunks, and the repo already accepts the identical `source()` line beside it.

### Budget

The baseline follows the two deletions to **22005**, from 22080.

### Checks after the answers

| | |
|---|---|
| `verify-context-budget.R` | 22005 of 22005 |
| `verify-must-error.R` | 8 fixtures |
| `verify-doc-references.R` | 171 paths across 150 documents |
| `lintr::lint_package()` | no lints |

Issue #424, opened 2026-09-04
dtplyr nesting fails on a column named key, which data.table() reads as its own argument


sayuks:
## Summary

`nest_with_margins()` and `nest_by_with_margins()` fail on a `dtplyr` input
whose data carries a column named `key`, with a diagnostic naming
`data.table()` rather than anything the caller wrote. The same call on the same
data through the local `data.frame` backend succeeds.

The name is not marginplyr's: nothing here reserves it, and
`?nest_with_margins`'s *No input column name is reserved for internal
bookkeeping* says so.

## Reproduction

```r
pkgload::load_all(".", quiet = TRUE)
suppressPackageStartupMessages(library(dplyr))

data <- data.frame(
  group = factor(c("g1", "g2")),
  key = c("p", "q"),
  value = 1:2
)

nest_with_margins(data, .grouping = rollup(group))                    # works
nest_with_margins(dtplyr::lazy_dt(data), .grouping = rollup(group)) |>
  collect()                                                           # fails
```

```
Error in `data.table(key = key, value = value)`:
key argument of data.table() must be character
```

## Root cause

`nest_expanded_margins()` builds a cell with `dplyr::pick(dplyr::everything())`,
which dtplyr translates to a `data.table()` call with one argument per column.
`data.table()` has a `key` formal, so a column named `key` arrives as that
argument instead of as a column.

Nothing about marginplyr is required to reach it — the same expression on a bare
`lazy_dt()` fails the same way:

```r
d <- data.frame(g = c("a", "a", "b"), key = c("p", "q", "p"), v = 1:3)
dtplyr::lazy_dt(d) |>
  summarize(data = list(dplyr::pick(dplyr::everything())), .by = g) |>
  collect()
#> Error: some columns are not in the data.table: [p, q]
```

Found while writing the tests for #421, whose fixture held a `key` column. The
fixture was changed and this was split out; `na_level_nested_data()` in
`tests/testthat/test-margin-label.R` carries a comment pointing here.

## What is unmeasured

- Which other `data.table()` formals collide. `key` is the one observed;
  `check.names`, `stringsAsFactors`, and `.name_repair` are the candidates, and
  whether each is reachable through dtplyr's translation has not been tested.
- Whether `dplyr::tibble(.rows = dplyr::n())`, the other cell expression, has a
  collision of its own.
- Whether the fix belongs here or upstream in dtplyr. A cell built by naming
  columns positionally, or through a translation that does not spread them into
[...]

PR #425, opened 2026-09-04
Restore an NA factor level inside a nested cell (#421)

### LINKED ISSUE #421: dtplyr nest payload drops a declared NA factor level inside its cell
## Summary

On a dtplyr input, a factor column carrying a declared `NA` level loses it when
`nest_with_margins()` folds it into `.key`. The same call on the same data
through the local `data.frame` backend keeps it, so the nested column's domain
differs by backend with no error, warning, or other diagnostic.

This is the residual of #415, which widened the encode-and-rebuild route to the
columns a verb's result carries *as columns of its own* — a fixed `.by` key, or a
column `expand_with_margins()` passes through. A nested payload column is not one
of those: it crosses the branch union like every other column, but by the time
`finalize_margin_operation()` runs it is inside a cell, where
`restore_margin_factors()` has nothing to rebuild.

## Reproduction

```r

data <- data.frame(
  passthrough = structure(
    c(1L, 2L),
    levels = c("a", NA_character_),
    class = "factor"
  ),
  value = 1:2
)

nested <- function(input) nest_with_margins(input, .grouping = rollup(group))

cat("nest payload -- local:  ")
print(levels(nested(data)$data[[1]]$passthrough))
cat("nest payload -- dtplyr: ")
print(levels(as.data.frame(collect(nested(dtplyr::lazy_dt(data)))$data[[1]])$passthrough))
```

## Observed / Expected

Observed, verbatim, at `7e26f4e` (the head of #415's branch):

```
nest payload -- local:  [1] "a" NA
nest payload -- dtplyr: [1] "a"
```

Expected: the dtplyr line prints `[1] "a" NA`, as its local line does, with
`is.na()` on the cell's column agreeing between the two.

## Root cause

The same one as #408 and #415: `data.table`'s rbind drops a declared `NA` level
from the result and turns the values that used it into missing codes.

`nest_margin_pipeline()` hands `carried_columns = function(data_vars, plan) plan$by`
to `prepare_margin_operation()`, and the comment there says why: the encode arm of
`label_margin_branch()` and the rebuild in `restore_margin_factors()` are two halves
of one route, and encoding a column the finalizer cannot reach would leave it as
character — strictly worse than the level it loses today.

## What was unmeasured

All three items are measured; the results are in the comment below, and the
prototype there is green under the full suite. In summary: the rebuild goes
before `nest_expanded_margins()` folds the cell, because the dtplyr union turns
the values on the declared level into genuine missing before any fold, so an
in-cell rebuild has nothing to rebuild from; `.keep = TRUE`'s copies take the
same loss, and so does the nested copy of a *grouping* column carrying an NA
level, whose outer column #415 restores while its cell copy keeps the character
domain; and `nest_by_with_margins()` does not differ.

The comment's *Acceptance* section is what this issue is done against.

sayuks:
Closes #421.

A factor column carrying a declared `NA` level lost it on a dtplyr input
whenever a nesting verb folded it into `.key`, so the nested column's domain
differed by backend with no diagnostic. `.keep = TRUE` carried a second case the
issue did not name: an NA-level *grouping* column was restored on the outer
column and lost inside the cell of the same result.

The measurements that settled the fix's shape are in
https://github.com/sayuks/marginplyr/issues/421#issuecomment-5536247881. The
decisive one is that dtplyr's union turns the values on the level into genuine
missing before anything folds a cell, so the rebuild has to happen while the
payload is still a column — an in-cell rebuild would restore the level with
nothing left on it.

`#424` was filed from this branch's test work: a column named `key` breaks
nesting under dtplyr, because `pick(everything())` translates to a
`data.table()` call whose `key` formal swallows it. Not marginplyr's, and the
fixture here avoids the name and says why.

Full suite green, `lintr::lint_package()` clean. No generated file changed.

### DIFF
--- a/R/backend-metadata.R
+++ b/R/backend-metadata.R
@@ -70,10 +70,11 @@ proxy_columns <- function(data_proxy, cols) {
 }
 
 # `carried` names the input columns that reach the result beside the Margin
-# dimensions -- a fixed `.by` key, or a column the verb passes through. They
-# cross the same branch union and lose a declared NA level in the same place,
-# so `factors` covers them too (#415). Which columns those are is settled by
-# `prepare_margin_operation()`'s `carried_columns`.
+# dimensions -- a fixed `.by` key, a column the verb passes through, or one it
+# folds into a cell. They cross the same branch union and lose a declared NA
+# level in the same place, so `factors` covers them too (#415, #421). Which
+# columns those are is settled by `prepare_margin_operation()`'s
+# `carried_columns`.
 #
 # `prototypes` stays keyed to the dimensions alone. It stands for the value a
 # branch omitting a dimension writes, and only a dimension is ever omitted.
--- a/R/margin-label.R
+++ b/R/margin-label.R
@@ -412,10 +412,11 @@ label_margin_branch <- function(.data,
     function(info) info$col,
     character(1)
   )
-  # A column `factor_info` names that is not a dimension is a fixed `.by` key
-  # or one the verb passes through, so it is in every branch rather than in
-  # the ones that include it, and it is never labelled: only the encoded arm
-  # above can select it (#415).
+  # A column `factor_info` names that is not a dimension is one the verb
+  # carries -- a fixed `.by` key, one it passes through, or one it folds into a
+  # cell -- so it is in every branch rather than in the ones that include it,
+  # and it is never labelled: only the encoded arm above can select it (#415,
+  # #421).
   encoded_factors <- Filter(
     function(info) {
       present <- !(info$col %in% plan$dimensions) || info$col %in% included
--- a/R/margin-operation.R
+++ b/R/margin-operation.R
@@ -43,9 +43,14 @@ new_margin_operation <- function(data,
 # resolves in the aggregate query alone, and dbplyr discards the ordering of
 # any query it goes on to wrap in a subquery — which labelling the omitted
 # dimensions and placing the grouping columns both do.
-new_margin_execution <- function(result, sort_id = NULL) {
+#
+# `factor_info` is the columns the finalizer can still reach, and `NULL` says
+# every column the operation named. Only an executor that puts a column
+# somewhere the finalizer cannot `mutate()` reports one: nesting folds its
+# payload into a cell and rebuilds it there (#421).
+new_margin_execution <- function(result, sort_id = NULL, factor_info = NULL) {
   structure(
-    list(result = result, sort_id = sort_id),
+    list(result = result, sort_id = sort_id, factor_info = factor_info),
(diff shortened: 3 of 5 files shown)


sayuks:
## Review round 1

`mattpocock-skills:code-review`, both axes, against the merge-base with `main`
(`e653f31`) at `637f8e1`. Answered in `232dda7`.

### Standards

No documented-standard violation. Two baseline smells, which were one site.

| finding | disposition |
|---|---|
| **Duplicated Code** — `execute_margin_nest()` derived the payload columns and `nest_margin_pipeline()` derived their complement, in different functions over separately built sets, with nothing tying them | fixed in code: `new_margin_execution()` carries `factor_info`, so the split is derived once beside the fold |
| **Mysterious Name** — `outer_cols` for `by + dimensions`, while the same file already used it for `group_cols + set_col` | fixed in code: the site is gone with the finding above |

The axis also answered three correctness questions the invocation put to it —
whether the finalizer can be handed an entry naming a column the result does not
hold, whether every field of a `.keep` copy's synthesized entry is correct under
the copy's name, and whether the Grouping set identifier can fall through the
payload filter — and found none of them reachable. The last has a reason worth
keeping: `check_margin_id_collision()` refuses a `.id` colliding with a source
column and internal names come from `new_margin_internal_names()`, so the
identifier is never in `factor_info` at all.

### Spec

No defect, against the Acceptance in
https://github.com/sayuks/marginplyr/issues/421#issuecomment-5536247881. Six
break attempts were run against `load_all()` — an NA level and a genuine missing
in one payload column, an NA-level fixed `.by` key under `.keep = TRUE`,
`cube()`, `.margin_label_position = "first"`, an ordered payload factor with the
NA level in the middle, and a payload column named `..marginplyr_nest_1` — and
dtplyr matched local in every one.

| finding | disposition |
|---|---|
| `nest_by_with_margins()` covered at `.keep = FALSE` alone, so "with `.keep` either value" held for one verb of two | fixed in code: `nest_by_with_margins keeps a used NA level with .keep = TRUE` |
| `expect_cell_column_agrees()` sorted the codes, fixing the multiset and not the `is.na()` pattern the Acceptance names | fixed in code: cells are put in `value` order and the codes compared position by position |
| uncovered: a fixed `.by` key under nesting, `cube()`, `.margin_label_position = "first"`, `.id`, `.sort` | split: the `.by` key is a route of its own — a grouping column, so the payload rebuild passes it over and it reaches the cell as a `.keep` copy — and has a case. The rest are rejected: none changes which columns are folded or how they are encoded, and a label position on a carried column is asserted by `a Margin label position adds no level to a carried column` |

Full suite green and `lintr::lint_package()` clean at `232dda7`.

PR #426, opened 2026-09-04
Build a nested cell without spreading columns into a callee's formals (#424)

### LINKED ISSUE #424: dtplyr nesting fails on a column named key, which data.table() reads as its own argument

sayuks:
Closes #424.

## What was wrong

The nesting verbs built their cell with `pick(everything())`. dtplyr
translates a `pick()` standing where a value stands into a literal
`data.table()` call carrying one named argument per column, so a payload
column named for one of that function's formals arrived as that argument:

| Payload column name | Before |
|---|---|
| `key` | error |
| `check.names` | error, with a different diagnostic |
| `keep.rownames` | no error; the column was absent from every cell |
| `stringsAsFactors` | no error; the column was absent from every cell |

The two silent cases are the worse half and were not in the issue as filed;
triage measured them.

## What changed

`nest_cell_expr()` takes the payload columns and the names they carry, and
names them into `list()`, whose only formal is `...`. Converting afterwards
rather than naming into `tibble()` directly, because that function has formals
(`.rows`, `.name_repair`) a column could be named for too.

Naming the cell's columns subsumes the `.keep = TRUE` branch's `rename()` and
`relocate()`, so the two branches of `nest_expanded_margins()` are now one and
it no longer takes `.keep`.

## Behaviour change

Cells under dtplyr are tibbles rather than data tables. The element class is
described rather than promised; the reference prose that described it per
backend is updated, and no test asserted it. The README's nesting output is
unchanged.

## Coverage

`test-nest-operation.R` gains a test over six column names — the four
`data.table()` formals and the two `tibble()` formals the obvious alternative
fix would have broken — across both verbs and both `.keep` settings, comparing
the dtplyr result against the local one and reading the cell names back
separately. It requires `dtplyr` and no other member of `optional_backends()`.

`na_level_nested_data()` in `test-margin-label.R` existed only because
`na_level_carried_data()` holds a column named `key`, which this bug refused.
The two are merged back into one fixture, so the #421 tests now read a nested
cell that carries that name.

## Checks

Full suite green, `lintr::lint_package()` clean, `roxygen2::roxygenise()`
committed, `spelling::spell_check_package()` clean,
`verify-context-budget.R` within baseline.

### DIFF
--- a/R/nest_by_with_margins.R
+++ b/R/nest_by_with_margins.R
@@ -151,7 +151,7 @@ nest_by_with_margins <- function(.data,
     result <- dplyr::summarize(
       empty_data,
       "{.key}" := list(
-        !!nest_cell_expr(colnames(empty_data), colnames(empty_data))
+        !!nest_cell_expr(rlang::set_names(colnames(empty_data)))
       )
     )
     if (!is.null(.id)) {
--- a/R/nest_by_with_margins.R
+++ b/R/nest_by_with_margins.R
@@ -151,7 +151,13 @@ nest_by_with_margins <- function(.data,
     result <- dplyr::summarize(
       empty_data,
       "{.key}" := list(
-        !!nest_cell_expr(rlang::set_names(colnames(empty_data)), "local")
+        !!nest_cell_expr(
+          rlang::set_names(colnames(empty_data)),
+          # The kind of the input, not of `empty_data`: this rebuilds what the
+          # pipeline returned for it, so its cell is the one every other cell
+          # from this input is.
+          grouping_backend(.data)$kind
+        )
       )
     )
     if (!is.null(.id)) {
--- a/R/nest_by_with_margins.R
+++ b/R/nest_by_with_margins.R
@@ -151,7 +151,7 @@ nest_by_with_margins <- function(.data,
     result <- dplyr::summarize(
       empty_data,
       "{.key}" := list(
-        !!nest_cell_expr(rlang::set_names(colnames(empty_data)))
+        !!nest_cell_expr(rlang::set_names(colnames(empty_data)), "local")
       )
     )
     if (!is.null(.id)) {
--- a/R/nest_by_with_margins.R
+++ b/R/nest_by_with_margins.R
@@ -151,7 +151,7 @@ nest_by_with_margins <- function(.data,
     result <- dplyr::summarize(
       empty_data,
       "{.key}" := list(
-        !!nest_cell_expr(empty_payload = ncol(empty_data) == 0L)
+        !!nest_cell_expr(colnames(empty_data), colnames(empty_data))
       )
     )
     if (!is.null(.id)) {
--- a/R/nest_with_margins.R
+++ b/R/nest_with_margins.R
@@ -51,10 +51,11 @@
 #'
 #' The list column is a regular list of data frames; its exact `vctrs_list_of`
 #' subclass is not part of the API. Neither is the class of its elements,
-#' which follows what the backend made of the cell expression and is tibbles
-#' on both today. Only their being data frames holding the input's non-key
-#' columns is promised, so `lapply(result$data, tibble::as_tibble)` is what a
-#' caller who needs the class itself writes. [nest_with_margins()] follows
+#' which follows whichever backend produced them: tibbles for a local input,
+#' and data tables under `dtplyr`, which is what [tidyr::nest()] on the same
+#' input gives. Only their being data frames holding the input's non-key
+#' columns is promised. Call `lapply(result$data, tibble::as_tibble)` when one
+#' element class is needed across backends. [nest_with_margins()] follows
 #' [tidyr::nest()] for an empty ungrouped input and returns zero outer rows.
 #' [nest_by_with_margins()] follows [dplyr::nest_by()] and returns one row
 #' containing the empty input when there are no grouping keys.
@@ -62,18 +63,19 @@
 #' When nesting leaves no payload column — every input column is a fixed key or
 #' a grouping dimension, and `.keep` does not put them back — each nested data
 #' frame still has one row per source row it stands for, as [dplyr::nest_by()]
-#' does. That row count is promised; the class of such a cell is described
-#' rather than promised, as every element class is.
+#' does. That row count is promised. The class of such a cell is described
+#' rather than promised, as every element class is: on both backends it is what
+#' [dplyr::tibble()] produced, because a `data.table` cannot hold rows without
+#' columns at all.
 #'
-#' A `data.table` cannot hold rows without columns at all, which is a limit on
-#' what an input can carry into `dtplyr` rather than anything nesting does: a
-#' data frame with rows and no columns loses its rows on the way in, so
-#' `dtplyr::lazy_dt(data.frame(row.names = 1:3))` is already empty before
-#' marginplyr reads it and no behavior here can restore the three rows. Nest
-#' an input that has rows and no columns locally when its row count matters.
-#' This is the whole of the difference: a column-less input that reaches the
-#' backend with the rows it had -- one with no rows either -- nests to the
-#' same result on both backends.
+#' That last point is also a limit on what an input can carry into `dtplyr`,
+#' rather than anything nesting does: a data frame with rows and no columns
+#' loses its rows on the way in, so `dtplyr::lazy_dt(data.frame(row.names =
+#' 1:3))` is already empty before marginplyr reads it and no behavior here can
+#' restore the three rows. Nest an input that has rows and no columns locally
+#' when its row count matters. This is the whole of the difference: a
+#' column-less input that reaches the backend with the rows it had -- one with
+#' no rows either -- nests to the same result on both backends.
 #'
 #' No input column name is reserved for internal bookkeeping. Temporary
 #' grouping-set and `.keep` columns are generated collision-free and removed
@@ -378,6 +380,7 @@ execute_margin_nest <- function(operation, .key, .keep) {
           set_col = set_col,
           keep_cols = keep_cols,
           .key = .key,
+          kind = operation$backend$kind,
           drop_set_col = is.null(operation$set_id_name) && !sorting
         ),
         sort_id = if (sorting) set_col else NULL,
@@ -388,11 +391,12 @@ execute_margin_nest <- function(operation, .key, .keep) {
   )
 }
 
-# The expression building one cell, given the columns it is to hold: a named
-# character vector whose names are the names the cell's columns take and whose
-# values are the columns of the expanded step they read. An empty one is a
-# nesting that has no payload column left.
-nest_cell_expr <- function(cell_cols) {
+# The expression building one cell, given the columns it is to hold and the
+# backend kind that will evaluate it. `cell_cols` is a named character vector
+# whose names are the names the cell's columns take and whose values are the
+# columns of the expanded step they read; an empty one is a nesting that has
+# no payload column left.
+nest_cell_expr <- function(cell_cols, kind) {
   if (length(cell_cols) == 0L) {
     # A nesting that removes every payload column still stands for a known
     # number of source rows per cell, and once the columns are gone the count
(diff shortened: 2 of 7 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` against the merge-base with `main` (`3358720`), over `165556d`. Two axes, six findings, all answered in `8ff2371`.

### Standards

| Finding | Disposition |
|---|---|
| ADR 0016 asserts "data tables under dtplyr" in two passages this branch made false, and it is the file that owns the element-class decision | fixed in prose — `## Amendment` on ADR 0016 naming both passages |
| `nest_cell_expr()`'s header described dtplyr's translation and never stated the function's contract (*Code comments*) | fixed in code — header states the contract, each branch's argument moved inline |
| The new test's header re-derived the argument `nest_cell_expr()`'s comment owns (*Code comments*, second copy) | fixed in prose — it cites that function and #424 |
| Data Clumps: `payload_cols` and `out_names` travel together; `nest_cell_expr(colnames(x), colnames(x))` at one call site | fixed in code — one named character vector, in `keep_cols`'s direction |
| Duplicated Code: `formal_shadow_cells()` is `sales_cells_as_tibble()` with a different key | fixed in code — both helpers take the keys, the duplicate is gone |

### Spec

Verified holding: all six column names agree local vs. dtplyr for both verbs and both `.keep` settings; the empty-payload branch is unchanged; `.keep` cell names and order match `main`, non-syntactic and reserved-word names included; `verify-suite-coverage.R` exits 0, so the new test satisfies the one-backend policy. No scope creep, nothing implemented wrong.

| Finding | Disposition |
|---|---|
| The brief asked for the `data.table()`-formal capture as an *inline* comment; it was written as a header paragraph | fixed in code — the same edit the Standards axis asked for |
| Only half the reference sentence changed: the `lapply()` advice stood beside a passage now saying both backends give tibbles | fixed in prose — it says why a caller writes it, the class being described and not promised |

Checks after the answers: full suite green, `lintr::lint_package()` clean, `spelling::spell_check_package()` clean, `roxygen2::roxygenise()` committed.

sayuks:
## The nested element class is back to what each backend's own nesting gives (`91a1b22`)

The collision-safe cell built a tibble on both backends, and the argument for letting it stand was that no test asserted the class and the reference did not promise it. That is what nobody having considered the question looks like, not a decision — and the measurement settles it:

```r
tidyr::nest(dtplyr::lazy_dt(d), data = c(x, v))
#> Call:   `_DT3`[, .(data = .(.SD)), by = .(g)]
#>   g     data
#> 1 a     <dt [2 x 2]>
```

dtplyr translates `tidyr::nest()` itself and yields a `data.table`. These verbs are documented as following that verb, so a caller nesting the same lazy step with and without marginplyr should not get different element classes.

`nest_cell_expr()` takes the backend kind and names the converter for it: `dplyr::as_tibble()` locally, `data.table::as.data.table()` under dtplyr. Both take the columns through `list()`, so what #424 fixed is untouched — no column name reaches a formal of either converter. Verified over all six names, both verbs, both `.keep` settings.

The `data.table::` call sits behind a backend-kind guard, as `arrow::schema()` does; dtplyr imports data.table, so it is never reached without it, and `DESCRIPTION` needs no change.

Element classes match `main` again, so the reference paragraph and ADR 0016's are restored. Each keeps one change, because one clause in each was a reason rather than an observation:

- the reference gave `dplyr::pick()`'s return value as the reason, and the cell no longer calls it;
- the ADR said marginplyr chooses neither class, and it now does — choosing per backend what that backend's own nesting verb answers. The amendment names that sentence and says a backend changing its nesting verb is a change to the ADR.

**Coverage for the gap this exposed:** "dtplyr nesting agrees with the local result and stays lazy" now asserts the element class per backend, which nothing did before. A payload-free cell is a tibble on either backend — a `data.table` has no form that holds rows without columns — so the assertion reads the scenario's cell names to know which it expects.

The result is still a lazy `dtplyr_step`; only the expression inside the `j` changed.

Checks: full suite green, `lintr::lint_package()` clean, `spelling::spell_check_package()` clean, `roxygen2::roxygenise()` committed, `verify-suite-coverage.R` exits 0 (753 tests).

sayuks:
## Review record, round 2

`mattpocock-skills:code-review` against the merge-base with `main` (`3358720`), over `91a1b22`. Round 1 ran over `165556d` and was answered in `8ff2371`; `91a1b22` — the backend-kind branch and the `data.table::` call site — is what no round had seen. Three findings, answered in `ade2975`.

### Spec

**A regression this branch introduced.** dtplyr's squash folds a symbol named `T` or `F` to the logical constant *before* it consults the step's columns, so naming the cell's columns as bare symbols put `TRUE` in the cell in place of a column named `T` — both verbs, both `.keep` settings, silently:

```r
d <- data.frame(region = c("East","East","West"), T = c("p","q","r"), units = 1:3,
                check.names = FALSE)
collect(nest_with_margins(dtplyr::lazy_dt(d), .grouping = rollup(region)))$data[[1]]$T
#> 91a1b22: TRUE TRUE      main: "p" "q"
```

`main` escaped it because `pick()` reached a translation that rebuilds the call from column names. Fixed by naming the columns through `margin_column_pronoun()` — the repo's existing answer to this hazard, whose header already gives the reason for the `.data[["name"]]` form, and which reaches the column on either backend. Verified: `T`, `F`, `.data` and the original six all correct on both backends, both verbs, both `.keep` settings.

`formal_shadow_names` gains `T`, `F`, `.data`, and the test now reads the shadowed column's **values** back, not only its name — a folded name leaves a cell that has the column and the right row count, so a name-only assertion passes. Confirmed red without the fix.

(The data.table pronouns `.N`, `.I`, `.SD`, `.GRP`, `.BY`, `.NGRP` error as column names on both `main` and here, loudly and before the cell is built. Pre-existing and out of this branch's scope.)

**The reference's reason clause was false for the payload-free cell** — `tidyr::nest()` there gives a 0-column `data.table`, and marginplyr gives a tibble on purpose. The clause is gone; the paragraph that owns that case already says what it is.

### Standards

**A rule the code did not keep.** `nest_by_with_margins()`'s reconstruction of an empty ungrouped input passed the kind `"local"` — honest about the collected frame it evaluates against, wrong about the input — so an empty `dtplyr` nesting gave a tibble where every other cell from that input is a `data.table`. It passes the input's kind now, with a test. `main` had the same divergence; what is new is ADR 0016's amendment stating the rule it breaks.

The payload-free cell is the one real exception, and the amendment says so rather than leaving the reference to carry it alone.

**Three copies of one argument.** The amendment's reasoning was re-derived in `nest_cell_expr()`'s inline comment and again in the test's. Both cite ADR 0016 and stop.

### Verified clean by the review

`data.table::as.data.table()` needs no `marginplyr_suggest_available()` guard: `packageDescription("dtplyr")$Imports` carries `data.table (>= 1.13.0)`, and the `identical(kind, "dtplyr")` guard is the shape *Dependency metadata* sanctions for `arrow::schema()`. No `DESCRIPTION` edit. `list()` wrapping is load-bearing — `as.data.table.list` has formals `keep.rownames, key, check.names, .named`, none reachable through a positional `list()`. Element classes, cell column order, and `dtplyr_step_subset` laziness match `main` for non-colliding data.

### Behaviour change beyond `main`

One, and it is the finding above: `nest_by_with_margins()` on an empty **dtplyr** input now returns a `data.table` cell rather than a tibble. Everything else matches `main`.

Checks: full suite green (754 tests), `lintr::lint_package()` clean, `spelling::spell_check_package()` clean, `roxygen2::roxygenise()` committed, `verify-suite-coverage.R` exits 0.

sayuks:
## Review record, round 3

`mattpocock-skills:code-review` against the merge-base with `main` (`3358720`), over `ade2975` — what round 2's answers introduced and no round had seen. Four findings, all prose, answered in `5f0c683`.

### Spec — no regression found

The axis's primary job this round was to hunt for another silent regression of the `T`/`F` family. It ran four differential batteries HEAD vs `main` in side-by-side worktrees, comparing cell class, names, `nrow`, per-column class, per-column `attributes()`, formatted values, and the outer frame, across both verbs × both `.keep` settings × local/dtplyr:

- **65 column names** — all of `dtplyr:::dt_symbols`, `T F TRUE`, `.data .env . .. ..x ..N ..1 ..2 ...`, `if let by V1 N class levels row.names structure`, non-syntactic (`a b`, `a-b`, ``a`b``, `日本語`, embedded newline), and the six from the brief. Zero unintended diffs; `T`/`F` byte-identical to `main`.
- **19 column types** — factor, ordered, `Date`, `POSIXct` with `tzone`, `difftime`, `hms`, list-column, packed df-column, matrix column, `integer64`, all-`NA`, custom attributes, custom S3, `POSIXlt`, `vctrs_rcrd`, raw, complex, UTF-8. Zero diffs, attributes included.
- **208 shape and option cases** — cell column order, row order within cells, row counts, zero-row inputs, empty payload, `.id`, all three `.sort` values, `.margin_label_position`, `.key`, `.duplicates="drop"`, grouped inputs, `.by` fixed columns, `cube()`, factor dimensions, NA factor levels. Zero diffs.

All acceptance criteria met, including the `#421` fixture question and the regenerated reference.

### Findings answered

|---|---|---|
| `nest_cell_expr()`'s header says `kind` is the backend that *evaluates* the expression; `nest_by_with_margins()` passes the input's kind into an expression a local `summarize()` evaluates, as its own call-site comment says | Standards | fixed in prose — `kind` names the converter |
| "the grouping columns lead it, which is what `.keep` promises" — `.keep` is documented as keeping the *values*; nothing promises the cell's column order | both | fixed in prose — clause deleted; the order stands on its own and the dtplyr test pins it |
| "a `data.table` having no form that holds rows without columns" is in five places, two added by rounds 1–2's answers | Standards | fixed in prose — the test comment goes (`lazy_cell_class()` beside it carries the fact), and the ADR amendment points at the reference paragraph rather than re-deriving why |

### Checked and clean

`grouping_backend(.data)` called a second time is not an ADR 0002 breach — that ADR governs the typed *schema* snapshot and says explicitly it is not a rule that an operation reads a backend once. The function is class inspection only; its one query path is unreachable here because nesting refuses SQL inputs upstream, and `check_backend_version()` is idempotent. No new execution entry point, so ADR 0020 and `test-query-policy.R` are untouched. Every claim in the ADR 0016 amendment was verified by running code, and it follows ADRs 0001/0009's amendment form. The `margin_column_pronoun()` call satisfies that function's stated contract.

### Still outside #424

`nest_by_with_margins()` on an empty **dtplyr** input returns a `data.table` cell rather than a tibble. It answers round 2's finding that the ADR amendment stated a rule this path broke; the alternative answer was to scope the prose and leave `main`'s behaviour. Flagged for the maintainer rather than settled here.

Checks: full suite green (754 tests), `lintr::lint_package()` clean, `spelling::spell_check_package()` clean, `roxygen2::roxygenise()` committed.

sayuks:
## Decision: the empty-dtplyr cell stays a `data.table`

The maintainer's call on the one item round 3 left open. Verified that what the branch gives is what dtplyr gives:

```
tidyr::nest() on an empty lazy_dt   -> 1 outer row, data.table cell, cols x,y, 0 rows
DT[, .(data = list(.SD))]           -> 1 outer row, data.table cell, cols x,y, 0 rows
nest_by_with_margins(lazy_dt(empty))-> 1 outer row, data.table cell, cols x,y, 0 rows
```

The local path is unchanged and still matches `dplyr::nest_by()` (tibble cell). So the round-2 answer — passing the input's backend kind rather than the literal `"local"` — is what makes the two agree, and it stays.

This is the branch's only behaviour change against `main` outside #424, and it is now a decision rather than a side effect. No code change; recorded here.

PR #433, opened 2026-09-04
Write a dtplyr margin label full-size on every dimension type (#427)

### LINKED ISSUE #427: dtplyr expand and nest fail on every non-character, non-factor dimension carrying a margin label
## Summary

`expand_with_margins()`, `nest_with_margins()`, and `nest_by_with_margins()`
fail on a `dtplyr` input whenever a Margin dimension is `integer`, `double`,
`Date`, or `logical` and the margin label is a non-missing string. The same
call on the same data through the local `data.frame` backend succeeds.

`?expand_with_margins` says the verb "works with local data frames and `dtplyr`
steps", and *Display labels and grouping identity* says a non-missing label
converts an ordinary dimension to character. Neither happens here: the call
raises from data.table, naming nothing the caller wrote.

`summarize_with_margins()` is unaffected — its omitted dimension does not reach
the branch as a column of the input.

## Reproduction

```r

df <- data.frame(i = c(1L, 1L, 2L), v = 1:3)

expand_with_margins(df, .grouping = rollup(i))                     # works
expand_with_margins(dtplyr::lazy_dt(df), .grouping = rollup(i)) |>
  collect()                                                        # fails
```

```
Warning in `[.data.table`(copy(`_DT1`), , `:=`(i = "Total")) :
  Coercing 'character' RHS to 'integer' to match the type of column 1 named 'i'
Warning: NAs introduced by coercion
Error in .set_ops_arg_check(x, y, all, block_list = !all) :
  Item 1 of x is 'character' but the corresponding item of y is 'integer'.
```

Measured over one dimension of each type, on `dtplyr::lazy_dt()`:

| Dimension type | `expand` | `nest` | `nest_by` | `summarize` |
|---|---|---|---|---|
| `integer` | fails | fails | fails | ok |
| `double` | fails | — | — | ok |
| `Date` | fails | — | — | ok |
| `logical` | fails | — | — | ok |
| `character` | ok | ok | ok | ok |
| `factor` | ok | ok | ok | ok |

The `double`, `Date`, and `logical` rows raise the same error with
`'numeric'`, `'Date'`, and `'logical'` in place of `'integer'`.
`.margin_label = NULL` and `.margin_label = NA_character_` are unaffected on
every type.

## Root cause

`label_margin_branch()` (`R/margin-label.R:383`) writes the label into an
omitted dimension as a scalar in `dplyr::mutate(.data, !!!values)`. dtplyr
renders a scalar overwrite as data.table `:=`, which coerces the RHS to the
column's *existing* type rather than replacing the column. The omitted branch
therefore keeps `integer` where the included branches carry `character`, and
`funion()` refuses the pair.

The factor arm at `R/margin-label.R:467-481` already carries the fix for its
own case, and its comment states the mechanism:

> dtplyr renders a scalar overwrite as data.table `:=`. On a multi-row factor
> column, data.table recycles it without replacing the column class [...] A
> full-size value replaces the column so factor restoration remains after the
> union.

The same is true of every other non-character type; only factors were given
`rep(!!label, dplyr::n())`.

## Fix direction

Widen the `dtplyr_factor_cols` arm to any column that is not already
`character` on a `dtplyr_step`, so the labelled value is written full-size
there too. `labelled_included` above it converts *included* columns with
`as.character()`; the omitted ones have no equivalent.

## What is unmeasured

- Whether any other backend renders a scalar overwrite the same way. The union
  adapter's SQL backends were not tested for this, since the label there is
  built inside the branch query rather than by a `mutate()` on a step.
- Whether a list-column or a `difftime` dimension behaves as the four types
  above.

sayuks:
Fixes #427.

`expand_with_margins()`, `nest_with_margins()`, and `nest_by_with_margins()`
failed on a `dtplyr` input whenever a Margin dimension was `integer`, `double`,
`Date`, or `logical` and the margin label was a non-missing string. The same
call through the local `data.frame` backend succeeded, and the error came from
data.table naming nothing the caller wrote.

## Cause

`label_margin_branch()` wrote the label into an omitted dimension as a scalar
in `dplyr::mutate()`. dtplyr renders a scalar overwrite as data.table `:=`,
which recycles the value and coerces it to the column's *existing* type rather
than replacing the column. The omitted branch therefore kept `integer` --
holding `NA` where the label was -- while the included branches carried
`character`, and `funion()` refused the pair.

The factor arm already carried the fix for its own case with
`rep(label, n())`; the mechanism its comment names is the same one every other
non-character type meets.

## Change

The full-size form is now selected by the backend rather than by the column
set: on a `dtplyr_step`, every labelled omitted dimension is written
`rep(label, n())`. A full-size value replaces the column, which is what makes
the labelled branch match the `as.character()` the included branches took. A
character dimension is unaffected either way, so no schema read is needed to
decide.

## Test

`tests/testthat/test-margin-label.R` gains a test generalising the existing
ordered-factor case over the four types that failed -- `integer`, `double`,
`Date`, `logical` -- across all three expansion verbs. It requires only
`dtplyr`, so the one-backend-per-test policy holds.

Full suite green with `NOT_CRAN` set; `lintr::lint_package()` reports no lints.

### DIFF
--- a/R/margin-label.R
+++ b/R/margin-label.R
@@ -460,13 +460,14 @@ label_margin_branch <- function(.data,
       dplyr::across(dplyr::all_of(labelled_as_character), as.character)
     )
   }
-  # dtplyr renders a scalar overwrite as data.table `:=`. On a multi-row
-  # column, data.table recycles the scalar and coerces it to the column's own
-  # type rather than replacing the column, so the omitted branch reaches
-  # `funion()` in the type it had: an ordered factor opposite the character
-  # branch, and an integer holding `NA` where the label was (#427). A full-size
-  # value replaces the column, which is what makes the labelled branch match
-  # the `as.character()` the included ones took above.
+  # dtplyr renders a scalar overwrite as data.table `:=`, which recycles the
+  # value and coerces it to the column's own type rather than replacing the
+  # column. The omitted branch then reaches `funion()` as the ordered factor or
+  # the integer it started as, holding the `NA` a character label coerced to,
+  # while the included branches are character (#427). A full-size value
+  # replaces the column. Every labelled dimension takes it rather than only the
+  # ones that are not character already, for which it replaces the column with
+  # itself.
   dtplyr_full_size_label <- inherits(.data, "dtplyr_step")
   values <- lapply(
     omitted,
@@ -483,7 +484,9 @@ label_margin_branch <- function(.data,
         # branches spell one. `as.character()` of the prototype cannot: it is
         # `NA`, which is also what a value on the NA level becomes, and the
         # union is where the two would stop being distinguishable (ADR 0012).
-        # Full size for the reason the labelled arm above is.
+        # Full size for the reason the labelled arm above is, and with no
+        # backend test because `drops_na_factor_level_on_union`, which
+        # `missing_label_encoded` requires, is dtplyr's alone.
         return(rlang::expr(rep(!!sentinels[[col]], dplyr::n())))
       }
       value <- prototypes[[col]]
--- a/R/margin-label.R
+++ b/R/margin-label.R
@@ -460,21 +460,20 @@ label_margin_branch <- function(.data,
       dplyr::across(dplyr::all_of(labelled_as_character), as.character)
     )
   }
-  # dtplyr renders a scalar overwrite as data.table `:=`. On a multi-row factor
-  # column, data.table recycles it without replacing the column class, leaving
-  # an ordered factor opposite the character branch at `funion()`. A full-size
-  # value replaces the column so factor restoration remains after the union.
-  dtplyr_factor_cols <- if (inherits(.data, "dtplyr_step")) {
-    vapply(factor_info, function(info) info$col, character(1))
-  } else {
-    character()
-  }
+  # dtplyr renders a scalar overwrite as data.table `:=`. On a multi-row
+  # column, data.table recycles the scalar and coerces it to the column's own
+  # type rather than replacing the column, so the omitted branch reaches
+  # `funion()` in the type it had: an ordered factor opposite the character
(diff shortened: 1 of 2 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review` ran once on `7524c48`, against the merge-base
with `main` (`a1cfe56`). Standards sources: `AGENTS.md`, `CONTEXT.md`,
`design/architecture.md`, `design/adr/`, `design/agents/code-review.md`. Spec
source: #427. Answered in `a4e81dd`.

**Standards** -- no hard violation; five judgement calls.

| Finding | Disposition |
|---|---|
| The labelled arm's comment closes on a clause that is false for a dimension already `character` | fixed in prose |
| "Full size for the reason the labelled arm above is" leaves the sentinel arm's missing backend test unexplained | fixed in prose |
| `inherits(.data, "dtplyr_step")` classifies a backend outside `grouping_backend()` | rejected: `label_margin_branch()` is not handed the backend, and `R/backend-metadata.R`'s comment on `encode_missing_label` says why it is not |
| `factor_info` may have become partly dead | rejected: `margin_factor_sentinels()`, `missing_label_encoded`, and `encoded_factors` all still read it |
| The test is weaker than its neighbour -- no laziness assertion, no failure context, title overclaims | fixed in code |

**Spec** -- nothing missing, nothing wrong; two observations.

| Finding | Disposition |
|---|---|
| The gate is the backend, where the issue's *Fix direction* said "any column that is not already `character`" | site left alone; the comment at it now says why |
| `difftime` and list-column dimensions, from the issue's *What is unmeasured*, are still untested | site left alone: the reviewer measured both as correct on dtplyr, and the test's title no longer claims them |

The reviewer verified every cell of the issue's table on `dtplyr::lazy_dt()`,
plus named mixed labels, `cube()`, and a composite `rollup(c(a, b))`, against
the local `data.frame` result.

PR #434, opened 2026-09-04
Give a summary branch with nothing to select a column (#428)

### LINKED ISSUE #428: Union adapter builds an unrenderable query when a lazy input is asked for no summaries
## Summary

`summarize_with_margins()` with no `...` summaries and no `.by` raises a dbplyr
internal error on every backend that takes the `UNION ALL` path. The local
`data.frame` path and the native `GROUPING SETS` path both return the distinct
keys for the same input.

The no-summary call is a documented shape — `?summarize_with_margins` describes
what the verb returns when "asked for no summaries and given no fixed key or
grouping dimension", and `tests/testthat/test-branch-union.R` asserts it for
dtplyr — and the docs promise that "Other lazy backends use a portable
`UNION ALL` adapter with the same semantics."

## Reproduction

```r

df <- data.frame(a = c("x", "x", "y", NA), v = 1:4)
DBI::dbWriteTable(con, "t", df)

summarize_with_margins(df, .grouping = rollup(a))                   # 4 rows
summarize_with_margins(tbl(con, "t"), .grouping = rollup(a)) |>
  collect()                                                          # fails
```

```
Error in `purrr::map()`:
i In index: 1.
Caused by error in `vars_prev[[1]]`:
! subscript out of bounds
```

The error comes out of `dbplyr:::select_use_star()` -> `vctrs::vec_match()`,
so it names nothing the caller wrote and no marginplyr function.

Also observed with `.margin_label = NULL`, with `.id = "id"`, and with
`rollup(a, b)`. On DuckDB the same call takes the native path and collects
fine (`SELECT a FROM t GROUP BY GROUPING SETS ((a), ())`); it fails there once
the call is routed off the native path, e.g.
`grouping_sets(rollup(a), grouping_set())` with `.duplicates = "keep"` and
`.id`. Adding a `.by` key makes the union path work, because the branch then
has a column.

## Root cause

`summarize_margin_union()` (`R/grouping-adapter-union.R:367`) builds each
branch with `summarize_margin_branch()` (`:278`). For the grand-total grouping
set with no summaries and no fixed key, that branch is a `summarize()` selecting
no columns at all, and the margin-label / `.id` `mutate()` layered on top of a
zero-column lazy query produces a `lazy_select_query` dbplyr cannot render.

## Fix direction

Give the empty grouping set a column-bearing branch: build the label and `.id`
columns inside the branch `summarize()` rather than in a `mutate()` over it, or
seed the branch with a constant that the final `select()` drops.

## What is unmeasured

- Which other zero-column shapes reach the same query. Only the no-summary,
  no-`.by` grand-total branch was traced.
- Whether the dtplyr path, which `test-branch-union.R` already covers for this
  shape, reaches the branch differently or only survives because data.table
  tolerates the zero-column step.

sayuks:
Fixes #428.

## What was wrong

`summarize_with_margins()` asked for no summaries and given no fixed key
built a Grand total branch that summarizes nothing and groups by nothing,
so the branch query selected no columns at all. The margin label and the
Grouping set identifier are layered on with `mutate()`, and over a
zero-column lazy query that produces a `lazy_select_query` dbplyr cannot
render: collecting failed inside dbplyr's own star expansion, naming
neither the caller's call nor any marginplyr function.

The database backends on the `UNION ALL` path were the ones affected. The
other two lazy backends on it were not: an Arrow input already collected
the local answer, and a `dtplyr` one already tolerated the zero-column
step. The local path and the native `GROUPING SETS` path both returned
the distinct keys for the same input.

## The fix

Such a branch now summarizes `n()` into a placeholder column of its own,
dropped again once the columns the result keeps are in place -- last, so
the identifier `mutate()` is not handed the zero-column query the label
one was. `n()` rather than a literal, because the placeholder has to
aggregate: a constant in a `summarize()` over no groups renders
`SELECT 1 AS x FROM t`, one row per source row, where `COUNT(*)` is the
single row the branch stands for.

## The shape left over

The remaining zero-column shape is the one whose whole *result* has no
columns -- no summaries, no key, no dimension, no `.id` -- which no
adapter can fix, since a SQL table of no columns cannot be written. That
is what `dplyr::summarize()` answers for the same input, as the `dtplyr`
half of the limit already was, so the documented limit on
`summarize_with_margins()` now names both inputs rather than dtplyr
alone, and says what an Arrow input answers.

## Verification

- `tests/testthat/test-branch-union.R` gains three tests, each on one
  optional backend: the reported call and three named neighbours held to
  the local answer, the reproduction #428 names for a backend routed off
  the native path, and the leftover shape held to `dplyr::summarize()`'s
  own refusal.
- Full suite, `NOT_CRAN=true`: FAIL 0 | SKIP 0 | PASS 6690.
- `lintr::lint_package()`: no lints.
- `Rscript .github/scripts/verify-context-budget.R`: within budget.

### DIFF
--- a/R/grouping-adapter-union.R
+++ b/R/grouping-adapter-union.R
@@ -276,23 +276,23 @@ abort_absorbed_summary <- function(labels) {
 # maintainer's signal is `test-query-policy.R`, which fails when a read
 # happens at all.
 #
-# `seed_name` names one column the branch summarizes for itself. Only a caller
-# whose branch would otherwise select nothing at all passes one, and
+# `placeholder_name` names one column the branch summarizes for itself. Only a
+# caller whose branch would otherwise select nothing at all passes one, and
 # `summarize_margin_union()` is where that is decided and where the column is
 # dropped again.
 summarize_margin_branch <- function(.data,
                                     ...,
                                     .by,
                                     caller_labels,
-                                    seed_name = NULL) {
+                                    placeholder_name = NULL) {
   dots <- rlang::enquos(...)
 
-  # The seed joins what dplyr is handed and not `dots`, which stays the
+  # The placeholder joins what dplyr is handed and not `dots`, which stays the
   # caller's: the handler below places a blamed expression in it by position,
   # and a column the caller did not write has no label to be blamed under.
   summarize_dots <- dots
-  if (!is.null(seed_name)) {
-    summarize_dots[[seed_name]] <- rlang::quo(dplyr::n())
+  if (!is.null(placeholder_name)) {
+    summarize_dots[[placeholder_name]] <- rlang::quo(dplyr::n())
   }
 
   result <- withCallingHandlers(
@@ -405,17 +405,17 @@ summarize_margin_union <- function(.data,
   # A branch that summarizes nothing and groups by nothing selects no columns
   # at all, and the label and identifier `mutate()`s layered on that query give
   # dbplyr a `lazy_select_query` whose star expansion has no column to read
-  # (#428). Such a branch carries a column of its own instead, dropped once the
-  # columns the result keeps are in place.
+  # (#428). Such a branch summarizes a column of its own to stand in the
+  # result's place, dropped once the columns the result keeps are there.
   #
-  # `n()` and not a literal, because the seed has to aggregate: a constant in a
-  # `summarize()` over no groups renders as `SELECT 1 AS seed FROM t`, one row
-  # per source row, where `COUNT(*)` is the single Grand total row the branch
-  # stands for.
-  seed_name <- new_margin_internal_names(
+  # `n()` and not a literal, because the placeholder has to aggregate: a
+  # constant in a `summarize()` over no groups renders as `SELECT 1 AS x FROM
+  # t`, one row per source row, where `COUNT(*)` is the single Grand total row
+  # the branch stands for.
+  placeholder_name <- new_margin_internal_names(
     1L,
     used_names = c(reserved_names, unname(key_names)),
-    prefix = "..marginplyr_seed_"
+    prefix = "..marginplyr_placeholder_"
   )
 
   conditions <- new_branch_conditions(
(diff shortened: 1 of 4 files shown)


sayuks:
## Review record

One round, `mattpocock-skills:code-review`, against the merge-base with `main`
(`git diff main...HEAD`, merge-base a1cfe56) over bb90523. Eight findings; all
answered in 1da6173 except the one that asked for a measurement.

### Standards

| finding | disposition |
|---|---|
| The widened limit in the roxygen — "One result **no lazy backend** can carry back" — is false: an Arrow input collects the one row holding nothing that a local data frame returns, on this branch and before it | fixed in prose |
| The comment above `check_summary_output_names()` claimed to speak for "every question below", where only that one call drops the placeholder | fixed in prose |
| The SQL-table-of-no-columns argument was in a third place, beside the identifier attachment, where `R/grouping-backend.R` and the roxygen already hold it | fixed in prose — the copy is now a citation |
| Mysterious Name: `seed_name` says nothing about why the column exists | fixed in code — `placeholder_name`, `..marginplyr_placeholder_` |
| The four cases were an unnamed list, so a failure named none of them | fixed in code — named, with `info = label` |

### Spec

| finding | disposition |
|---|---|
| Same false widening, reached as scope creep: #428 asked to fix the union branch, not to generalize the limit | fixed in prose |
| `grouping_sets(rollup(a), grouping_set())` with `.duplicates = "keep"` and `.id` — the reproduction #428 names for a backend routed off the native path — worked but had no test | fixed in code |
| #428's *What is unmeasured*: "Whether the dtplyr path ... reaches the branch differently or only survives because data.table tolerates the zero-column step" | measured, no change: `main` and this branch give identical dtplyr results for the no-summary shape, with `.id` and without, so data.table tolerated the zero-column step and the placeholder changes nothing there |

The measured answer is recorded here rather than in a comment, because it is a
fact about what `main` did and not about the code beside it.

Issue #435, opened 2026-09-04
A rewritten data-frame-valued summary is named, so dplyr packs what it used to expand


sayuks:
## Summary

#430 names an unnamed summary marginplyr rewrites, from the expression the
caller wrote. dplyr expands a data-frame-valued summary's columns into the
result only while that summary is unnamed, and packs them into one column under
any name. So a data-frame-valued summary a rewrite reaches now comes back
packed where it used to expand.

`name_rewritten_summary_dots()` (`R/summary-selections.R`) excludes the two
data-frame-valued shapes a static reading recognizes — a summary written as
`across()` or `pick()` itself, and the `tibble()`/`data.frame()` family. What is
left is a data-frame-valued expression the reading does not recognize.

## Reproduction

```r
pkgload::load_all(".", quiet = TRUE)
suppressPackageStartupMessages(library(dplyr))

data <- data.frame(group = c("a", "a", "b"), x = c(1, 3, 5), y = c(2, 4, 6))
range_frame <- function(columns) data.frame(lo = min(columns[[1L]]), hi = max(columns[[1L]]))
totals <- function(value, bit) data.frame(sum = sum(value), margin = bit)

names(summarize(data, range_frame(pick(x)), .by = group))
#> [1] "group" "lo"    "hi"

names(summarize_with_margins(data, range_frame(pick(x)), .grouping = rollup(group)))
#> [1] "group"                       "range_frame(dplyr::pick(x))"

names(summarize_with_margins(data, totals(x, grouping_bit(group)), .grouping = rollup(group)))
#> [1] "group"                          "totals(x, grouping_bit(group))"
```

A summary no rewrite reaches — `range_frame(x)` — expands as before, and so
does `across(c(x, y), mean)`.

`tests/testthat/test-grouping-interface.R`'s *a rewritten data-frame-valued
summary is named and packed* asserts both shapes above, so this is pinned
rather than latent.

## Why #430 did not fix it

Which of two expressions is data-frame-valued is a question about the value's
type, and the two are one call apart: `nrow(pick(v, w))` has to be named, which
is #430's second reproduction, while `range_frame(pick(x))` has to not be.
ADR 0019 reads spellings statically, and no static reading separates them.

#430's own *Fix direction* — `rlang::quos_auto_name()` applied before any
rewriting — packs strictly more than this, `across()` and `range_frame(x)`
included. What shipped is the narrower rule.

## What a fix would need

Deciding after the summarize rather than before: name the dot, and where a
[...]

PR #436, opened 2026-09-04
Name an unnamed summary from the caller's own expression

### LINKED ISSUE #430: Unnamed summaries are auto-named from the rewritten expression, splitting union branches
## Summary

An unnamed summary is handed to `dplyr::summarize()` after marginplyr has
rewritten it, so dplyr auto-names the column from the *rewritten* expression
rather than from what the caller wrote. This has two consequences:

1. An unnamed summary containing `grouping_bit()` or `grouping_id()` gets a
   different column name in each union branch, and the call dies inside an
   internal invariant `stop()` on local, dtplyr, and arrow while the SQL path
   returns a result.
2. An unnamed summary containing `pick()` or an `across()`-family helper gets a
   column name carrying marginplyr's qualified rewrite instead of dplyr's.

Both come from the same site, so they are filed together.

`?summarize_with_margins` describes `...` as "Name-value pairs as used in
`dplyr::summarize()`", which fixes dplyr's naming as the contract.

### 1. Crash on a contextual helper

```r

df <- data.frame(a = c("x", "y"), v = 1:2)

summarize_with_margins(df, sum(v), .grouping = rollup(a))                # ok
summarize_with_margins(df, sum(v) + grouping_bit(a), .grouping = rollup(a))
```

```
Error : Union branch 2 does not have the columns of branch 1.
Only in branch 2: sum(v) + 1L
Only in branch 1: sum(v) + 0L
```

A bare `grouping_bit(a)` gives the same error with `1L` and `0L`. dtplyr raises
`x and y must have the same column names`; arrow raises
`UnionNode input schemas must all match`. DuckDB succeeds and returns a column
named `+...`.

### 2. Wrong column name

```r
df <- data.frame(g = c("a", "a", "b"), v = 1:3, w = c(2.5, 3.5, 4.5))

names(summarize(df, nrow(pick(v, w)), .by = g))
#> [1] "g" "nrow(pick(v, w))"

names(summarize_with_margins(df, nrow(pick(v, w)), .grouping = rollup(g)))
#> [1] "g" "nrow(dplyr::pick(dplyr::all_of(c(\"v\", \"w\"))))"
```

## Root cause

`rewrite_grouping_dots()` (`R/grouping-context.R:128`) substitutes a different
literal per branch — `0L` or `1L` for `grouping_bit()` — and
`resolve_summary_selections()` (`R/summary-selections.R:384`) rewrites helper
calls into qualified `all_of()` form. Both return the quosures with their
original, empty names, and `execute_margin_summary()`
(`R/summarize_with_margins.R:896`) passes them to the adapters as they are, so
dplyr names each branch's column by deparsing whatever that branch received.

For case 1 the branch names then differ, and the union invariant at
`R/grouping-adapter-union.R:64` fires. That is a bare `stop()`, which ADR 0015
reserves for a marginplyr defect no call rewrite avoids — but it is reached
here by valid caller input.

`plan_summary_expressions()` already captures `caller_labels` before rewriting;
they are currently used only for condition restatement.

## Fix direction

Name the unnamed dots from the pre-rewrite expressions — `caller_labels`, or
`rlang::quos_auto_name()` applied before any rewriting — so every branch agrees
on the column name and the name is the one dplyr would have produced.

Note that this changes the column name a caller sees on the SQL path, where
such a summary today comes back as `+...` or `<sql>` (the deparse truncation
#411 investigated). That is the point of the fix, but it is a visible change.

## What is unmeasured

- Whether any other rewrite in `resolve_summary_selections()` reaches a
  column name, beyond `pick()` and the `across()` family.
- What DuckDB names the case-1 summary once the branches agree.

sayuks:
Closes #430.

## What changed

An unnamed summary was handed to `dplyr::summarize()` after marginplyr had
rewritten it, so dplyr named the column by deparsing the rewrite. A Grouping
helper became the branch's own `0L` or `1L`, so `sum(v) + grouping_bit(a)`
named a different column in each Grouping-set branch: the union adapter's
column invariant refused it — a bare `stop()` ADR 0015 reserves for a defect,
reached by valid caller input — and the native adapter named the column after a
SQL literal's deparse instead. A selection helper became a qualified `all_of()`
literal, so `nrow(pick(v, w))` named its column after the rewrite.

`plan_summary_expressions()` now names an unnamed dot from the expression the
caller wrote, after the labels a Condition context reads and before share
planning moves one.

Only a summary marginplyr rewrites is named. #430's *Fix direction* proposed
`rlang::quos_auto_name()` over every dot before rewriting; that also names the
data-frame-valued summaries dplyr expands only while they are unnamed, so
`across(c(x, y), mean)` and a caller's own one-row-frame function would have
come back packed into one column. The bound keeps both expanding.

`known_data_frame_output_names()`'s recognition moves into
`data_frame_valued_summary_kind()`, which both readers share.

## The one behaviour change that is not the fix

A data-frame-valued summary a rewrite *does* reach is named, and is then
packed: `range_frame(pick(x))` returned `lo` and `hi` and now returns one
data-frame column. Separating it from `nrow(pick(v, w))`, which #430 requires
be named, is a question about the value's type, which ADR 0019's static reading
does not answer. #435 holds it, `NEWS.md` states it, and
*a rewritten data-frame-valued summary is named and packed* asserts it.

## Verification

`testthat::test_local()` green with every optional backend installed;
`lintr::lint_package()` reports no lints; `roxygen2::roxygenise()` produces no
diff. The two reproductions in #430 were checked on local, dtplyr, Arrow,
DuckDB, and a simulated Postgres connection, which name the column identically.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -148,6 +148,20 @@
   evaluated to know what it calls — `get("pick")(units)` — is an ordinary call
   as it was.
   See *Relationship to dplyr summaries* in `?summarize_with_margins`.
+* An unnamed summary now takes its column name from the expression you wrote
+  rather than from the one marginplyr rewrote it into. `...` is documented as
+  `dplyr::summarize()`'s name-value pairs, and the rewrite ran first: a
+  `grouping_bit()` or `grouping_id()` became the branch's own `0L` or `1L`, so
+  `sum(v) + grouping_bit(a)` named a different column in each Grouping-set
+  branch — which stopped the operation inside an internal invariant on every
+  backend taking the portable `UNION ALL` path, and named the column after a
+  SQL literal on the backends running `GROUP BY GROUPING SETS`. A selection
+  helper became a qualified `all_of()` literal, so `nrow(pick(v, w))` named its
+  column `nrow(dplyr::pick(dplyr::all_of(c("v", "w"))))`. Both are now named
+  after the caller's own expression, identically on every backend. A summary no
+  rewrite reaches is named by dplyr exactly as before, and a data-frame-valued
+  summary still expands its columns into the result rather than packing them
+  into one column under a name (#430).
 * Dynamically named data-frame summaries now reserve collision-free internal
   grouping names, and opaque collisions fail with a targeted diagnostic.
   Lazy margin-label checks use portable numeric `CASE` aggregates across
--- a/NEWS.md
+++ b/NEWS.md
@@ -159,9 +159,15 @@
   helper became a qualified `all_of()` literal, so `nrow(pick(v, w))` named its
   column `nrow(dplyr::pick(dplyr::all_of(c("v", "w"))))`. Both are now named
   after the caller's own expression, identically on every backend. A summary no
-  rewrite reaches is named by dplyr exactly as before, and a data-frame-valued
-  summary still expands its columns into the result rather than packing them
-  into one column under a name (#430).
+  rewrite reaches is named by dplyr exactly as before, and so is a summary
+  written as `across()` or `pick()` itself, so both go on expanding a
+  data-frame value's columns into the result. A data-frame-valued summary a
+  rewrite *does* reach is now named, and dplyr packs a named one:
+  `range_frame(pick(x))` returned `lo` and `hi` and now returns one
+  data-frame column named after the call. Telling that summary from
+  `nrow(pick(v, w))` is a question about the value's type rather than about how
+  it is spelled, so give either one a name of your own where you want the
+  columns expanded or packed regardless (#430).
 * Dynamically named data-frame summaries now reserve collision-free internal
   grouping names, and opaque collisions fail with a targeted diagnostic.
   Lazy margin-label checks use portable numeric `CASE` aggregates across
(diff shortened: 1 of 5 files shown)


sayuks:
## Review record

One round of `mattpocock-skills:code-review` against the merge-base with `main`
(`a1cfe56`), over `52584d9`, on both axes.

**Spec** — the two reproductions in #430 were re-run on local, dtplyr, Arrow,
DuckDB, and a simulated Postgres connection, and both "What is unmeasured"
items were answered: the `identical()` comparison covers every rewrite in
`resolve_summary_selections()` rather than the enumerated two, and DuckDB names
the case-1 summary `sum(v) + grouping_bit(a)` like every other backend. No
scope creep. One finding, the same one Standards found independently.

**Standards** — three findings.

| Finding | Disposition |
|---|---|
| The header of `data_frame_valued_summary_kind()` cited `name_unnamed_summary_dots()`, which no commit ever held | Fixed in `5875ecd` |
| `name_rewritten_summary_dots()`'s header and the `NEWS.md` entry both claimed the bound reaches every data-frame-valued summary; a caller's own frame-returning function is rewritten whenever a helper sits inside it | Claims deleted and replaced by the bound in `5875ecd`; behaviour asserted by *a rewritten data-frame-valued summary is named and packed*; what is left is #435, cited at the site in `d18947a` |
| `known_data_frame_output_names()`'s last branch ran the `across` reader for whatever kind reached it, so a fourth kind would have been read as an `across()` | Fixed in `5875ecd`: the branch names its kind, with a backstop `stop()` (ADR 0015) |

PR #437, opened 2026-09-04
Read a named data-frame summary for its name alone

### LINKED ISSUE #431: A named data-frame-valued summary is refused for names it does not produce
## Summary

`summarize_with_margins()` refuses a *named* data-frame-valued summary — a
`tibble()`, `data.frame()`, or `across()` call given its own name — for
collisions with the inner argument names, which such a summary never produces
as top-level columns. dplyr packs them into the one named column.

The package already documents the rule it is violating: `?share_of_parent`
says a named `across()` "packs both results into one data-frame-valued column".

## Reproduction

```r

df <- data.frame(g = c("a", "a", "b"), v = 1:3)

str(summarize(df, s = tibble(g = sum(v)), .by = g))
#> 'data.frame': 2 obs. of  2 variables:
#>  $ g: chr  "a" "b"
#>  $ s: tibble [2 x 1]
#>   ..$ g: int  3 3

summarize_with_margins(df, s = tibble(g = sum(v)), .grouping = rollup(g))
#> Error: Summary results cannot overwrite grouping column:
#> i `g`.

summarize_with_margins(df, s = across(v, mean, .names = "g"), .grouping = rollup(g))
#> Error: Summary results cannot overwrite grouping column:
#> i `g`.

summarize_with_margins(df, s = tibble(a = sum(v)), .grouping = rollup(g), .id = "a")
#> Error: `.id` (`a`) conflicts with a summary output.
```

dplyr returns `g` and a packed `s` for the first; marginplyr should return the
same with the Margin rows added.

The *unnamed* forms are correctly refused, and must stay refused — dplyr
unpacks those, so `tibble(g = sum(v))` with no name really does produce a
top-level `g`.

## Root cause

`known_summary_output_names()` (`R/summary-selections.R:630`) delegates to
`known_data_frame_output_names()` (`:644`), which reads the argument names of
every `tibble()` / `data.frame()` / `across()` call in `...` regardless of
whether the dot itself carries a name. Those predicted names then feed the
grouping-column and `.id` collision checks in `execute_margin_summary()`
(`R/summarize_with_margins.R:914-928`), which run before execution.

The runtime check `check_summary_output_names()` on the actual branch result
passes for these calls; only the static prediction is wrong.

## Fix direction

In `known_summary_output_names()`, skip any dot whose own name is non-empty:
its only output is that name, which `names(dots)` already contributes. Keep the
unnamed path exactly as it is.

## What is unmeasured

- Whether `.unpack` on a named `across()` reintroduces top-level names that the
  skip would then miss. `?dplyr::across` documents `.unpack`, and it was not
  exercised here.
- Whether `pick()` under a name behaves the same as `tibble()` under one.

sayuks:
Closes #431.

`known_summary_output_names()` read the argument names of every `tibble()`,
`data.frame()`, `across()`, and `pick()` call in `...` regardless of whether
the dot itself carried a name. dplyr packs a named data-frame result into the
one column its name gives, so those argument names never reach the result's
top level, and the pre-execution checks in `execute_margin_summary()` refused
calls for collisions the result never had — both the grouping-column check and
the `.id` check.

The predictor now skips a dot whose own name is non-empty. `names(dots)`
already contributes that name, which is the dot's whole output. The unnamed
path is untouched, and the unnamed forms stay refused: dplyr unpacks those, so
`tibble(g = sum(v))` with no name really does produce a top-level `g`.

The issue left two shapes unmeasured. Both are measured and covered by tests:

- `across(.unpack = "{inner}")` renames *within* the packed column, so a
  template producing a grouping column's name is still no top-level name.
- `pick()` cannot rename at all — `pick(c(g = v))` is a tidyselect error — so
  what it packs is always input column names.

Review record: see the review comment on this pull request.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -148,6 +148,14 @@
   evaluated to know what it calls — `get("pick")(units)` — is an ordinary call
   as it was.
   See *Relationship to dplyr summaries* in `?summarize_with_margins`.
+* A data-frame-valued summary you gave a name to is no longer read for the
+  names it packs. `s = tibble(group = sum(value))`, `s = across(value, mean,
+  .names = "group")`, and `s = pick(value)` each produce one column called `s`,
+  because dplyr packs a named data-frame result rather than unpacking it, so
+  the argument names inside them cannot overwrite a grouping column or collide
+  with `.id`; they were refused for those collisions anyway. `across(.unpack =)`
+  renames within the packed column and is reached the same way. The unnamed
+  forms are unpacked to the top level and stay refused.
 * Dynamically named data-frame summaries now reserve collision-free internal
   grouping names, and opaque collisions fail with a targeted diagnostic.
   Lazy margin-label checks use portable numeric `CASE` aggregates across
--- a/NEWS.md
+++ b/NEWS.md
@@ -155,7 +155,10 @@
   the argument names inside them cannot overwrite a grouping column or collide
   with `.id`; they were refused for those collisions anyway. `across(.unpack =)`
   renames within the packed column and is reached the same way. The unnamed
-  forms are unpacked to the top level and stay refused.
+  forms are unpacked to the top level and stay refused, as is a named one
+  colliding by its own name. A backend that unpacks a named result rather than
+  packing it — Arrow, for `across()` — is refused after its branch runs, with
+  the diagnostic the local backend gives.
 * Dynamically named data-frame summaries now reserve collision-free internal
   grouping names, and opaque collisions fail with a targeted diagnostic.
   Lazy margin-label checks use portable numeric `CASE` aggregates across
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -380,8 +380,10 @@
 #' which it is omitted.
 #'
 #' Summary results may not overwrite a fixed key or grouping dimension,
-#' including through a data-frame-valued summary. The local dplyr backend can
-#' overwrite an existing variable and reuse a newly created summary in a
+#' including through an unnamed data-frame-valued summary, whose columns dplyr
+#' unpacks to the top level. A named one packs its columns into the single
+#' column its name gives, which collides with nothing. The local dplyr backend
+#' can overwrite an existing variable and reuse a newly created summary in a
 #' later expression, but other backends may not. marginplyr rejects grouping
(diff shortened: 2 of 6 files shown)


sayuks:
## Review record

One round of `mattpocock-skills:code-review` over `git diff main...HEAD` at `5de69fd`, two axes in parallel. Answered in `f342c07`.

### Standards

| finding | disposition |
|---|---|
| `R/summarize_with_margins.R:383` — "A named one packs its columns into the single column its name gives, which collides with nothing" is false; the packed column's own name reaches both checks through `names(dots)` | fixed in prose, and pinned by `a named data-frame summary is still checked by its own name` |
| no test pins that the name itself is still refused | fixed in code |
| "dplyr packs a named result" derived in three places (code comment, test comment, NEWS) | fixed by deleting the test comment's copy |
| the code comment enumerates a family list `known_data_frame_output_names()` owns below it | fixed by deleting the enumeration |
| `nzchar(rlang::names2(dots))` vs `names(dots)[nzchar(names(dots))]` are two spellings of one predicate | rejected: they take complementary halves of the same partition in different files, and the comment names the relationship |
| does this need an ADR? | no: ADR 0019 governs resolution by spelling, and this narrows which dots are scanned |

### Spec

| finding | disposition |
|---|---|
| all four reproductions behave as #431 says, matching `dplyr::summarise()` plus the Margin rows; the unnamed forms stay refused; `.unpack` and `pick()` measured rather than assumed | no finding |
| the comment's "produces that one column and nothing else" is false on Arrow, which unpacks a named `across()` to a top-level column and discards the name | fixed in prose; the mechanism was measured — the pre-execution check sees only the name, and the adapters' `check_summary_output_names()` refuses the branch result — and pinned by `Arrow's unpacked named across is caught after the branch runs` |
| a named `tibble()` on DuckDB now reaches SQL and fails with a parser error rather than the grouping-overwrite error | rejected: the unnamed non-colliding shape produced that identical error before this branch, and `dplyr::summarise()` produces it for the same call. It is dbplyr's answer to a data-frame-valued summary, not a diagnostic this branch removed |
| scope creep | none |

### Checks

`testthat::test_local()` under `NOT_CRAN=true` (arrow, duckdb, dtplyr, data.table installed), `lintr::lint_package()`, `spelling::spell_check_package()`, `verify-context-budget.R` — all clean.

PR #438, opened 2026-09-04
Blame the Margin verb on the native adapter's translation error

### LINKED ISSUE #432: The native adapter's translation error blames an internal call rather than the Margin verb
## Summary

A dbplyr translation error raised on the native `GROUPING SETS` path has its
arguments restated (#411, ADR 0022) but its `call` left untouched, so the error
header names a call the caller never wrote, spelled with the internal names
`.data` and `dots`. The union path rewrites the call to the Margin verb.

`CONTEXT.md` (*Condition context*) requires the other:

> the blamed call is the Margin verb the caller wrote, rather than [...] the
> calls marginplyr allocated

and ADR 0021 states that "an error now names the Margin verb".

This is the leftover of #411, which restated the argument spelling and said
explicitly that it was not routing the native adapter through
`with_branch_conditions()`.

## Reproduction

```r

df <- data.frame(a = c("x", "y"), v = c(1, 2))

con  <- DBI::dbConnect(duckdb::duckdb(shared_home = FALSE))
con2 <- DBI::dbConnect(RSQLite::SQLite(), ":memory:")
DBI::dbWriteTable(con,  "t", df)
DBI::dbWriteTable(con2, "t", df)

for (tb in list(tbl(con, "t"), tbl(con2, "t"))) {
  e <- tryCatch(
    summarize_with_margins(tb, n = sum(v, na.rm = TRUE), m = n(), .grouping = rollup(a)),
    error = identity
  )
  cat(class(tb)[1], "->", deparse(conditionCall(e)), "\n")
}
```

```
tbl_duckdb_connection -> dplyr::summarize(dplyr::ungroup(.data), !!!dots)
tbl_SQLiteConnection  -> summarize_with_margins(tb, n = sum(v, na.rm = TRUE), m = n(), .grouping = rollup(a))
```

Printed, the DuckDB error opens
`Error in dplyr::summarize(dplyr::ungroup(.data), !!!dots) :`.

No wrong data results from this; the cost is entirely in the diagnostic.

## Root cause

`summarize_margin_native()` (`R/grouping-adapter-native.R:63-68`) catches the
translation error, applies `restate_condition_arguments()`, and re-raises with
`cnd$call` as dbplyr set it. `with_branch_conditions()`, which the union
adapter uses, sets the call to the operation's own.

## Fix direction

Set `cnd$call` to `operation$call` before re-raising, alongside the existing
`restate_condition_arguments()`. `with_branch_conditions()` is still not the
right vehicle for the reason #411 gave — the native adapter issues one
`summarize()` and repeats nothing, so the deduplication and grouping-value
restatement have nothing to act on.

## What is unmeasured

- Whether the second `summarize()` in `summarize_margin_native()` can raise a
  translation error of its own with the same untouched call. Only the first was
  reached.
- Whether the PostgreSQL native path blames the same call as DuckDB. Only
  DuckDB was run.

sayuks:
Closes #432.

A dbplyr translation error raised on the native `GROUPING SETS` path had its
arguments restated (#411, ADR 0022) but its `call` left as dbplyr set it, so
the error header named `dplyr::summarize(dplyr::ungroup(.data), !!!dots)` —
a call spelled with the internal names `.data` and `dots`. `CONTEXT.md`'s
*Condition context* requires the blamed call to be the Margin verb the caller
wrote, and ADR 0021 records that an error names it. The union adapter already
does this through `with_branch_conditions()`.

`summarize_margin_native()` now takes the operation's `call` and assigns it
alongside the existing `restate_condition_arguments()`, after the restatement
rather than inside it, since that function returns the condition untouched when
the map is empty. The argument is optional and defaults to `NULL`, as
`summarize_margin_union()`'s is, so a caller reaching the adapter directly
leaves the blamed call alone.

`with_branch_conditions()` is still not the vehicle, for the reason #411 gave.

## Verification

The issue's reproduction, before and after:

```
tbl_duckdb_connection -> dplyr::summarize(dplyr::ungroup(.data), !!!dots)
tbl_SQLiteConnection  -> summarize_with_margins(tb, ...)
```
```
tbl_duckdb_connection -> summarize_with_margins(tb, ...)
tbl_SQLiteConnection  -> summarize_with_margins(tb, ...)
```

`test-execution-conditions.R`'s new *a native translation error blames the
Margin verb* fails without the change. It runs on `simulate_postgres()`, so it
needs no optional backend.

`lintr::lint_package()` clean; full suite green under `NOT_CRAN=true`.

## What the issue left unmeasured

Both are recorded in
`investigation/which-native-summarize-raises-a-translation-error.md`, which
ADR 0022's amendment cites.

- **Can the second `summarize()` raise a translation error of its own?** Not
  found. Seventeen expression shapes were probed on duckdb against both the
  ungrouped call `native_summary_output_names()` issues and the grouped call
  below it, and no shape failed the grouped call while the ungrouped one
  succeeded. That is evidence and not a proof, which is what ADR 0022's
  amendment already says: "Which of the two dbplyr reaches first is not a
  contract", and the tests assert the restored context through the verb, so a
  dbplyr that moved the translation fails them rather than dropping this
  silently.
- **Does the PostgreSQL native path blame the same call?** Not run — no
  PostgreSQL driver is named in `optional_suggest_spec()`, so the suite has no
  connection to reach it with. `simulate_postgres()` covers the translation
  through the same adapter code and blames the verb.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -252,9 +252,12 @@
   whose own text spells the line dplyr writes to point at
   `dplyr::last_dplyr_warnings()`, while warnings that differ from each other
   are still reported one by one. Only that context changes: the class, the
-  diagnostic, and the cause you receive are the ones raised. A lazy input is
-  unaffected, because its summary expressions run when you collect the result
-  rather than while the verb runs (#141, #108).
+  diagnostic, and the cause you receive are the ones raised. A lazy input
+  raises most of these when you collect the result rather than while the verb
+  runs, so they are yours to receive and not the verb's to restate; the one
+  exception is an expression the database cannot be given at all, which is
+  refused while the verb runs and is quoted and blamed the same way
+  (#141, #108, #411, #432).
 * `nest_with_margins()` and `nest_by_with_margins()` now use collision-free
   internal columns. `.keep = TRUE` retains original pre-margin key values,
   and nesting rejects duplicate sets with `.duplicates = "keep"` because
--- a/R/grouping-adapter-native.R
+++ b/R/grouping-adapter-native.R
@@ -2,17 +2,19 @@
 # header gives, and read here for its labels as well as its dots.
 #
 # `call` is the Margin verb a Condition context is owed instead of the internal
-# `summarize()` this adapter issues; a caller that reaches this directly has
-# none to name and leaves the blamed call alone, as `summarize_margin_union()`
-# does.
+# `summarize()` this adapter issues. It has no default, unlike
+# `summarize_margin_union()`'s, because the executor is this adapter's only
+# caller: a default would stand for a caller that does not exist, and an
+# optional argument nothing omits is one the reader has to establish is never
+# omitted.
 summarize_margin_native <- function(.data,
                                     summaries,
                                     plan,
                                     margin_labels,
                                     reserved_names,
+                                    call,
                                     set_id_name = NULL,
-                                    set_id_is_internal = FALSE,
-                                    call = NULL) {
+                                    set_id_is_internal = FALSE) {
   con <- dbplyr::remote_con(.data)
   dots <- rewrite_grouping_dots(
     summaries$dots,
@@ -70,14 +72,12 @@ summarize_margin_native <- function(.data,
   # The blamed call is rewritten alongside the argument, both being parts of
   # one Condition context (CONTEXT.md). It is assigned after the restatement
   # rather than inside it, because that function returns the condition
-  # untouched when the map is empty.
+  # untouched when the map is empty, and this half is owed either way.
   output_names <- tryCatch(
     native_summary_output_names(.data, dots),
     error = function(cnd) {
       cnd <- restate_condition_arguments(cnd, restatements)
-      if (!is.null(call)) {
-        cnd$call <- call
-      }
+      cnd$call <- call
       stop(cnd)
     }
   )
--- a/R/grouping-adapter-native.R
+++ b/R/grouping-adapter-native.R
@@ -1,12 +1,18 @@
 # `summaries` is `stage_margin_summaries()`'s, taken whole for the reason its
 # header gives, and read here for its labels as well as its dots.
+#
+# `call` is the Margin verb a Condition context is owed instead of the internal
+# `summarize()` this adapter issues; a caller that reaches this directly has
+# none to name and leaves the blamed call alone, as `summarize_margin_union()`
+# does.
 summarize_margin_native <- function(.data,
                                     summaries,
                                     plan,
                                     margin_labels,
                                     reserved_names,
                                     set_id_name = NULL,
-                                    set_id_is_internal = FALSE) {
+                                    set_id_is_internal = FALSE,
+                                    call = NULL) {
   con <- dbplyr::remote_con(.data)
   dots <- rewrite_grouping_dots(
     summaries$dots,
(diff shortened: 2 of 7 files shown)


sayuks:
## Review round 1 — `mattpocock-skills:code-review` against `main`, at `0555ebb`

Two axes, parallel sub-agents. Nine findings; seven answered in code or prose, two rejected.

### Standards

| # | Finding | Disposition |
|---|---|---|
| 1 | ADR 0022's amendment says "Nothing else about the adapter moves", which `0555ebb` falsified. | Fixed in prose — `aecd58b`. That sentence now says what #411 moved; a second amendment records what #432 moves. |
| 2 | `design/architecture.md`'s portable-adapter section claims it is the one path that owes a Condition context. | Fixed in prose — `aecd58b`. Went false at #411; the sentence now names what is still only that adapter's, and the native adapter's section says what it owes. |
| 3 | Duplicated Code — the handler repeats `restate_branch_error()`'s shape, incl. the `NULL` convention. | Fixed in code — `aecd58b`, via Spec (b): with no default there is no guard, and what the two paths share is one field assignment. |
| 4 | The header's direct-caller clause is a third copy of a convention `R/conditions.R` and `R/grouping-adapter-union.R` state. | Fixed in code — `aecd58b`. The clause described no caller and is gone. |
| 5 | Data Clumps — both adapters take the same ~nine parameters. | **Rejected.** Pre-existing and not this diff's: the signatures were already parallel, and `call` is one the portable adapter already had, so the change converges them. A finding this branch's Acceptance does not reach is a ticket, per *Answering a review*. |

### Spec (issue #432)

| # | Finding | Disposition |
|---|---|---|
| 1 | `NEWS.md`'s "A lazy input is unaffected, because its summary expressions run when you collect the result rather than while the verb runs" is false on this path. | Fixed in prose — `aecd58b`. False since #411; it now names the exception. |
| 2 | ADR 0022's amendment is now false. | Same as Standards 1. |
| 3 | `design/architecture.md:494-496` contradicted. | Same as Standards 2. |
| 4 | The issue's two *unmeasured* items are neither measured nor recorded. | Fixed — `aecd58b`. `investigation/which-native-summarize-raises-a-translation-error.md` measures the first (seventeen expression shapes over both summarizes; none refused by the grouped call alone) and records that the second is not runnable here, no PostgreSQL driver being named in `optional_suggest_spec()`. ADR 0022 cites it. |
| 5 (b) | `call = NULL` and its guard are unreachable — the executor is the only caller. | Fixed in code — `aecd58b`. `call` has no default. Confirmed by `grep -rn summarize_margin_native R/ tests/ vignettes/ design/`: two hits, the definition and the executor. `summarize_margin_union()` keeps its default, tests reaching it directly. |
| 6 (c) | The test asserts a call name and the absence of `"dots"`, and covers only `native_translation_failure()`. | Fixed in code — `aecd58b`. It asserts the whole call, which is what an error header prints, and covers `native_shared_label_failure()` — the empty-map case the comment cites as the reason for the assignment order. |

### Verification after the answers

`lintr::lint_package()` clean. Full suite green under `NOT_CRAN=true` (0 failures). `verify-context-budget.R` and `verify-verifier-invocation.R` pass.

Worst within each axis: Standards — 1 (a false ADR amendment, since an amendment is what a later reader reads instead of the diff). Spec — 1 (a false changelog entry, `NEWS.md` being the widest-reach statement of this contract).

Issue #440, opened 2026-09-05
Every share on PostgreSQL is refused: the dialect probe's control rejects a bigint result


sayuks:
## Summary

Every share on PostgreSQL is refused, whatever the source summary's type. A
numeric source that computes a share on DuckDB and on a local data frame gets
this instead:

```
Error in `summarize_with_margins()`:
! marginplyr cannot establish that the source summaries of Parent shares are
  plain integer or double scalars on this backend, because it could not be
  asked whether its SQL dialect converts a value of another type to a number
  rather than refusing it, and a dialect that converts rejects nothing.
```

`.check_share_source = FALSE` is the only way to get a share on the dialect,
and the staged query it produces is correct — PostgreSQL computes the share
for a numeric source and raises for a character one, which is what the
`"refuses"` classification promises. The dialect earns that verdict and does
not receive it.

## Reproduction

Measured against PostgreSQL 17.11 (Homebrew) through RPostgres, dbplyr 2.6.0,
dplyr 1.2.1, at `1c126b2`.

```r
pkgload::load_all(".", quiet = TRUE)
suppressPackageStartupMessages(library(dplyr))

con <- DBI::dbConnect(RPostgres::Postgres(), host = "/tmp", dbname = "postgres")
tb <- copy_to(con, data.frame(g = c("a", "a", "b"), v = c(1, 2, 3)), "t",
              temporary = TRUE)

summarize_with_margins(
  tb, tot = sum(v, na.rm = TRUE), s = share_of_parent(tot), .grouping = rollup(g)
)
#> Error: ... could not be asked whether its SQL dialect converts ...

probe_share_dialect(con)
#> [1] "unknown"

# The same call with the check off stages a query PostgreSQL answers:
summarize_with_margins(
  tb, tot = sum(v, na.rm = TRUE), s = share_of_parent(tot),
  .grouping = rollup(g), .check_share_source = FALSE
) |> collect()
#>   g       tot     s
#>   a         3   0.5
#>   b         3   0.5
#>   Total     6   1
```

## Root cause

The probe raises, as intended — PostgreSQL rejects `SUM('x')`. The control is
then sent to tell a genuine refusal from a connection that could not answer,
and the control is what fails:

```r
probe_share_dialect_answer(con, quote(sum("x", na.rm = TRUE)), "p")
#> [1] "raised"
probe_share_dialect_answer(con, quote(sum(z, na.rm = TRUE)), "c")
#> [1] "unanswerable"
```

[...]

Issue #429, opened 2026-09-04
DuckDB calculates a share from a character source the eligible-type rule should refuse


sayuks:
## Summary

On DuckDB, `share_of_parent()` and `share_of_total()` compute a number from a
character source summary, with no diagnostic from either marginplyr or the
database. The same call on a local data frame is refused.

`?share_of_parent`'s *Lazy execution boundaries* states the contract this
breaks:

> The eligible-type rule is also enforced on every backend, and no backend
> calculates a share from a source it has shown to be ineligible.

and, for a dialect classified as refusing:

> Where it refuses an ineligible summary, that refusal is the answer and
> reaches you as the database's own diagnostic when `dplyr::collect()` executes
> the staged query.

Neither happens. This is the same class of defect as #106 (closed) on a
different dialect and through a different mechanism, so it is filed separately
rather than reopened.

## Reproduction

```r
pkgload::load_all(".", quiet = TRUE)
suppressPackageStartupMessages(library(dplyr))

df <- data.frame(a = c("x", "x", "y"), s = c("1", "2", "3"))
con <- DBI::dbConnect(duckdb::duckdb())
tb <- copy_to(con, df, "d")

summarize_with_margins(
  tb, total = max(s), ps = share_of_parent(total), .grouping = rollup(a)
) |> collect()

summarize_with_margins(
  df, total = max(s), ps = share_of_parent(total), .grouping = rollup(a)
)
```

```
# DuckDB, with the default `.check_share_source = TRUE`:
      a total        ps
  1   x     2 0.6666667
  2   y     3 1.0000000
  3 Total   3 1.0000000

# local:
Error in `summarize_with_margins()`:
! Parent share `ps` requires source summary `total` to be a plain integer or
  double scalar.
i Detected type character.
```

## Root cause

`probe_share_dialect()` (`R/share.R:2371`) measures the dialect with
`sum("x", na.rm = TRUE)`. DuckDB refuses `SUM('x')`, so the verdict is
`"refuses"` and `check_dialect_share_sources()` (`R/share.R:2233`) returns
without a diagnostic, leaving the rule to the database.

But the staged share query never applies an aggregate to the source. It
compares it to `0.0` and divides two `as.double()` casts (`R/share.R:2646`),
which dbplyr renders for DuckDB as `TRY_CAST(... AS DOUBLE)`. DuckDB casts a
numeric-looking `VARCHAR` implicitly in the `= 0.0` comparison, and `TRY_CAST`
[...]

PR #441, opened 2026-09-05
Record that the staged share can refuse a character source itself

sayuks:
Adds one dated note under `investigation/`. Nothing in `R/`, `tests/`, `man/`,
or `DESCRIPTION` changes.

#429 enumerated three options and left the third blocked on an unknown:
whether an expression exists that a refusing dialect rejects for a character
source value-independently and that fits the staged share's `mutate` context.
The other two both reclassify DuckDB as a converting dialect, which stops every
share on it — including the numeric ones that work today — so the third option
was the only one that fixes the defect without that cost, and it could not be
chosen while its premise was unmeasured.

The note measures it. Twenty-four expressions were run against DuckDB and
PostgreSQL 17.11 over a `DOUBLE` column and two `VARCHAR` columns. DuckDB
accepted a numeric-looking `VARCHAR` in four of them — the equality, `nullif`,
`IS NOT DISTINCT FROM`, and the cast — and the staged share's two rejecting
operations are among those four. The other twenty refused a `VARCHAR` argument
at binding.

*(Corrected: this paragraph first said "exactly two", which the review of this
branch found false. See the review record below and `918f891`.)*

`(x * 1) / (y * 1)` therefore substitutes for the two `as.double()` casts
rather than adding a term beside them, which is what #429 assumed the option
would cost. Applied to `R/share.R` and run end to end, DuckDB and PostgreSQL
both refused a character source whether or not it looked numeric, and both
calculated integer and double sources as before. The whole suite was run with
the edit in place: one assertion failed, and it reads the rendered SQL for a
cast rather than a behaviour.

The edit was reverted before this commit. What is proposed here is the
evidence, not the change — #429 carries the agent brief that spends it, and
#440 was filed from the same measurements.

Per `investigation/README.md`'s *Graduation*, the durable part moves to
`AGENTS.md` or an ADR in the commit that first cites this note as a reason.
Neither #429's fix nor #440's needs one: the note's finding is that the
existing contract in `?share_of_parent` can be made true, not that it should be
restated, so nothing here changes a decision that a document owns.

Refs #429, #440.

### DIFF
--- a/investigation/refusing-a-character-share-source-in-the-staged-query.md
+++ b/investigation/refusing-a-character-share-source-in-the-staged-query.md
@@ -1,0 +1,153 @@
+# Refusing a character share source from the staged query itself
+
+Investigated: 2026-09-05
+
+`investigation/share-source-eligibility-on-coercing-dialects.md` (2026-08-16)
+established that a dialect can be asked whether it converts a non-numeric value
+to a number, and that the answer is a property of the dialect. #429 then found
+that the question that probe asks and the operations the staged share performs
+are two different dialect behaviours, so DuckDB calculated a Parent share from
+a character source while classified as refusing.
+
+This note measures the third option #429 enumerated: changing the staged share
+so that a refusing dialect produces the refusal the contract promises. It
+establishes that such an expression exists, that it is one substitution rather
+than an addition, and what it costs the existing suite.
+
+Measured on 2026-09-05 against R 4.6.1, dplyr 1.2.1, dbplyr 2.6.0, duckdb
+1.5.5, RPostgres 1.4.10 with PostgreSQL 17.11 (Homebrew), RSQLite 3.53.3, and
+bit64 4.8.6, with marginplyr loaded from the working tree at `1c126b2` via
+`pkgload::load_all()`. Scripts were throwaway; commands are quoted inline.
+
+## What the staged share applied to its source
+
+The expression built for a lazy Parent share divided two `as.double()` casts
+and compared the denominator with `0`. dbplyr rendered that on DuckDB as:
+
+```sql
+CASE WHEN (... OR "..marginplyr_denominator_of_total_1" = 0.0) THEN NULL
+     ELSE (TRY_CAST(total AS DOUBLE) / TRY_CAST("...denominator..." AS DOUBLE)) END
+```
+
+Both operations were measured against DuckDB directly:
+
+| SQL | Result |
+|---|---|
+| `TRY_CAST('x' AS DOUBLE)` | `NA`, no error |
+| `TRY_CAST('1' AS DOUBLE)` | `1` |
+| `'x' = 0.0` | Conversion Error |
+| `'1' = 0.0` | `FALSE` |
+| `SUM('x')` | Binder Error |
+| `SUM('1')` | Binder Error |
+
+`TRY_CAST` raised for nothing, so the error a non-numeric source produced came
+entirely from the `= 0.0` comparison. `SUM`'s refusal held whatever the string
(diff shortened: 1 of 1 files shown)


sayuks:
## Review record

Ran `mattpocock-skills:code-review` against the merge-base with `main`
(`1c126b2`), both axes, over `ef4eb32`. Nine findings; all were about the note's
prose, and each falls in one of the five classes *Answering a review* permits.
Eight were accepted and answered in `918f891`; one was rejected.

|---|---|---|---|
| 1 | both | The option is misnumbered: the note calls #429's option 2 "the third option" twice, and quotes option 2's cost string as option 3's | fixed in prose |
| 2 | both | "the only two of the twenty-four" is false — DuckDB accepted four | fixed in prose |
| 3 | both | The table showed 22 of the 24 expressions the note claims | fixed in prose |
| 4 | Standards | "Four dplyr-level forms" — the table lists five | fixed in prose |
| 5 | Spec | "Arrow, dtplyr, and data.table … do not reach the staged expression at all" is false about the code | fixed in prose |
| 6 | Spec | The suite claim's conditions are unrecorded, so "the whole suite" is not checkable | fixed in prose |
| 7 | Spec | The failing assertion is quoted under the wrong test name, and "implementation rather than behaviour" is not carried | fixed in prose |
| 8 | Standards | Three sentences state current package behaviour | fixed in prose |
| 9 | Standards | The root-cause sentence restates #429 and cites nothing | rejected |

Findings 1, 2, 4, and 5 were verified against the sources before being accepted:
#429's numbered list, the battery's own output, and `share_adapter()` with
`apply_joined_shares()`.

**Finding 5 is the one worth naming.** `share_adapter()` routes `local`,
`dtplyr`, and `other` to `execute_row_matched_shares()`, and both adapters call
`apply_joined_shares()`, which is where the ratio is built. Those kinds reach
the same expression as the SQL kinds. The note's *What was not measured* bullet
said they do not reach it at all, so it denied a gap instead of recording one.
The suite run does cover them — every optional package was installed and
`MARGINPLYR_REQUIRED_SUGGESTS` was empty — but that is weaker than what the
bullet claimed, and it is what the note says now.

**Finding 2 also reached this pull request's body and two comments on #429**,
which asserted the same false count. Both are corrected there; a note is
authoritative for evidence, so a count it states and does not show is the
failure `investigation/README.md`'s *Authority* exists to prevent.

**Finding 9 rejected.** The sentence names what the note's own table then
measures, and is the premise the section exists to test rather than a second
copy of #429's argument. `AGENTS.md`'s *Code comments* reaches a paragraph that
re-derives an argument another file owns; this one states the question and
stops.

No finding touched `R/`, `tests/`, or a generated file — the branch changes
neither.

sayuks:

## The red CI here is inherited, not caused

This branch adds one file under `investigation/`, which `.Rbuildignore` keeps
out of the tarball. The failing jobs are `main`'s.

`test-execution-conditions.R:785` fails on `origin/main` at `535e3c0` with this
branch nowhere involved, and it has failed there since `ed3ae00`:

| commit | | test |
|---|---|---|
| `1c126b2` — this branch's merge-base | | passes |
| #438 branch head `aecd58b` | | passes |
| `ed3ae00` — merge of #438 | | **fails** |
| `535e3c0` — merge of #437, current `main` | | **fails** |

A semantic merge conflict: #438 branched from `a1cfe56`, where the blamed call
dropped the caller's argument names, while `main` gained `52584d9`, which moved
where a summary is named to before the Condition context reads those labels.
Neither branch was wrong on its own, and a merge runs no branch's CI.

Fixed in #443. This branch goes green once that lands and this one is rebased
or merged forward.

PR #442, opened 2026-09-05
Expand a data-frame summary marginplyr named

### LINKED ISSUE #435: A rewritten data-frame-valued summary is named, so dplyr packs what it used to expand

sayuks:
Closes #435.

## What changed

An unnamed summary a rewrite reaches takes an **Assigned summary name** — the column name marginplyr writes for it, read from the caller's own expression (#430). dplyr expands a data-frame-valued summary's columns into the result only while the summary is unnamed, so such a summary came back packed under that name where it used to expand.

Which summaries are data-frame-valued is a question about the value's type, and `nrow(pick(v, w))` — which #430 requires be named — is one call from `range_frame(pick(x))`, so no static reading separates them (ADR 0019). The decision is made after the summarize instead:

- `new_summary_arguments()` carries the assigned name for each dot beside the caller labels it already carried, under the same length invariant, subscripted by the same origin positions share planning reports.
- `summarize_margin_union()` expands a local branch's data-frame column sitting under an assigned name, between the branch's key rename and its margin labelling.
- `check_summary_output_names()` is asked a second time, of the names that replaced the assigned one. Where the package allocated a Grouping set identifier over the caller's `.id`, the adapter holds no name to ask the `.id` question with, so `stage_margin_summaries()` asks that one over the staged result.

`is.data.frame()` is the whole of the scope: Arrow packed such a summary before #430 too, dtplyr cannot hold a packed column, and SQL never expanded one, so nothing there is being restored. ADR 0020 is not engaged — the expansion reads a result already in memory, and `get_col_names()` is not an execution entry point.

## Documents

- New ADR 0028, *Expand a data-frame summary marginplyr named*.
- ADR 0016 gains one line naming it as the exception.
- `CONTEXT.md` gains **Assigned summary name**.
- `?summarize_with_margins`, under *Relationship to dplyr summaries*, states the rule a caller needs.
- The `NEWS.md` paragraph covering #430 is amended in place.

## Two deviations from the agent brief

1. The brief's *Key interfaces* says the shared helper is called from "Both grouping adapters (native and union)". It is called from the union adapter only. `stage_margin_summaries()` chooses the native adapter only where `supports_grouping_sets()` holds, which needs `native_grouping_sets`, which `backend_capabilities()` grants to the `duckdb` and `postgres` kinds alone — both lazy. So the helper's `is.data.frame()` guard could never pass there, and a branch nothing executes is one nothing asserts.
2. The brief says the expansion asks `check_summary_group_overwrite()` and `check_margin_id_collision()` but "Not `check_summary_output_names()`: its internal-name question has no subject at that point". That premise is false — `add_grouping_set_id()` writes the internal Grouping set identifier *after* the expansion runs, so an inner column of that name would be silently overwritten. The composite is asked instead, with no internal names of its own; `set_id_is_internal` is what puts the identifier back among them.

Both are argued in ADR 0028's *Where the expansion runs* and *What is checked again*.

## Checks

`lintr::lint_package()` after `pkgload::load_all()`, `roxygen2::roxygenise()`, `spelling::spell_check_package()`, the full test suite under `NOT_CRAN=true` (768 tests), `verify-context-budget.R`, `verify-verifier-invocation.R`, `verify-must-error.R`, `verify-suite-coverage.R`, and `verify-site.R` over a local render are all clean.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -196,6 +196,19 @@ runs when it is. A diagnostic refusing one tells the caller their spelling is
 glossary, and not a second term.
 _Avoid_: Contextual function, masked helper, reserved argument
 
+**Assigned summary name**:
+The column name marginplyr writes for an unnamed summary it rewrites, read from
+the caller's own expression rather than from the rewrite. It is never a name
+the caller wrote, and it exists only while the query is built: where a local
+result holds a data-frame column under one, that column's own columns take its
+place and the assigned name does not appear, so an unnamed summary expands as
+`dplyr::summarize()` expands it. A name the caller wrote is not one of these
+and packs, as it does under dplyr. Which unnamed summaries take one is decided
+by spelling, and whether one expands is decided from the value after the
+summary runs; ADR 0028 is authoritative for why the two are settled in
+different places.
+_Avoid_: Generated name, internal summary name, rewritten name
+
 **Option argument**:
 An argument whose value is one of a fixed set of strings, spelled in full. An
 abbreviation is not shorthand for the value it begins, and a `NULL` is not
--- a/NEWS.md
+++ b/NEWS.md
@@ -161,13 +161,17 @@
   after the caller's own expression, identically on every backend. A summary no
   rewrite reaches is named by dplyr exactly as before, and so is a summary
   written as `across()` or `pick()` itself, so both go on expanding a
-  data-frame value's columns into the result. A data-frame-valued summary a
-  rewrite *does* reach is now named, and dplyr packs a named one:
-  `range_frame(pick(x))` returned `lo` and `hi` and now returns one
-  data-frame column named after the call. Telling that summary from
-  `nrow(pick(v, w))` is a question about the value's type rather than about how
-  it is spelled, so give either one a name of your own where you want the
-  columns expanded or packed regardless (#430, #435).
+  data-frame value's columns into the result. So does a summary the naming
+  *does* reach whose value is a data frame, which dplyr would pack under any
+  name: on a local input marginplyr expands that column again after the
+  summary runs, so `range_frame(pick(x))` returns `lo` and `hi` as it did
+  before, and the name marginplyr assigned does not appear. Write a name of
+  your own to pack those columns into one column under it. Which summaries
+  those are is a question about the value's type — `nrow(pick(v, w))` is one
+  call away and has to be named — so it is answered from the value rather than
+  from the spelling, and only for an input already in memory: Arrow packed such
+  a summary before this change too, and no SQL backend expanded one
+  (#430, #435).
 * Dynamically named data-frame summaries now reserve collision-free internal
   grouping names, and opaque collisions fail with a targeted diagnostic.
   Lazy margin-label checks use portable numeric `CASE` aggregates across
--- a/R/grouping-adapter-union.R
+++ b/R/grouping-adapter-union.R
@@ -477,6 +477,18 @@ summarize_margin_union <- function(.data,
         result <- dplyr::rename(result, !!!rename_pairs)
       }
 
+      # After the rename and before the labelling: the grouping columns are
+      # under their own names by here, and the label and identifier columns
+      # the checks below would have to be told to ignore are not there yet.
+      # ADR 0028 is what expands.
+      result <- expand_assigned_data_frames(
+        result,
+        assigned_names = summaries$assigned_names,
+        group_vars = group_vars,
+        set_id_name = set_id_name,
+        set_id_is_internal = set_id_is_internal
+      )
+
       result <- label_margin_branch(
         result,
         plan = plan,
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -381,6 +381,16 @@
 #' every branch: a dimension remains excluded even in a grouping set from
 #' which it is omitted.
 #'
+#' An unnamed summary takes its column name from the expression you wrote
+#' rather than from what marginplyr rewrote it into, so
+#' `sum(v) + grouping_bit(a)` names the same column on every backend. Where
+#' such a summary's value is a data frame, a local input expands its columns
+#' into the result and the name marginplyr assigned does not appear, exactly as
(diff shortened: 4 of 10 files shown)


sayuks:
## Review record

Three rounds of `mattpocock-skills:code-review` against the merge-base with `main`, standards sources `AGENTS.md`, `design/architecture.md` (*Answering a review*), `CONTEXT.md`, and the ADRs the diff reaches; spec source #435 and its Agent Brief. Rounds 2 and 3 were scoped to the executable behaviour the previous round's answer produced, per `design/agents/code-review.md`.

### Round 1 — Standards (6 raised)

| finding | disposition |
|---|---|
| the header above `expand_assigned_data_frames()` re-derives three ADR 0028 sections beside its citation | fixed in prose, 38cd9e6 |
| same shape above `expand_packed_summary_column()` | fixed in prose, 38cd9e6 |
| the new test comment re-derives ADR 0028's *Why the decision cannot be made from the spelling* | fixed in prose, 38cd9e6 |
| Repeated Switches: the `set_id_is_internal` branch now exists twice in one file in two shapes | fixed in code, 38cd9e6 — the composite is called instead of its parts |
| Duplicated Code: `data.frame(sum = ..., margin = ...)` defined three times across two test files | rejected; the files already define their fixtures test-locally (`range_frame` in two tests on `main`), and the two files' fixtures answer different questions |
| `setdiff(name, names(columns))` is a scalar test spelled as a set operation | fixed in code, 38cd9e6 |

### Round 1 — Spec (2 raised)

| finding | disposition |
|---|---|
| the stated reason for the union-only deviation is weaker than the one that carries it — `dbplyr::remote_con()` answers `NULL` for a data frame rather than erroring | fixed in prose, 38cd9e6; the reason is dispatch, and the comment and ADR both say so |
| scope creep in the reference paragraph: it also documented #430's naming rule and the per-backend shapes, both outside what the brief asked for | fixed in prose, 38cd9e6 |
| *(note, brief-prescribed)* skipping the composite leaves the internal Grouping set identifier unasked about | the premise turned out false, so this became the round-1 code fix above |

All acceptance criteria were verified empirically by the Spec axis.

### Round 2 — scoped to the composite call (1 raised)

| finding | disposition |
|---|---|
| where a share replaces the caller's `.id` with an allocated identifier, an expanded inner name equal to that `.id` is asked by nobody, and the rename at the end destroys the summary rather than refusing it | fixed in code, f33285c; `stage_margin_summaries()` asks it over the staged result, and the test *an expanded inner name is checked against a replaced `.id`* fails without the fix |

Four further questions — the other three `set_id_name`/`set_id_is_internal` combinations, new false positives, whether the new test reaches its intended path, and whether the diagnostic names something actionable — were answered with no finding.

### Round 3 — scoped to that fix (1 raised)

| finding | disposition |
|---|---|
| the premise and consequence are stated three times: the comment, the ADR paragraph, and the test comment | fixed in prose, a442d09; the ADR keeps them and the two comments say which identifier the site holds and stop |

No false positive was found across local, dtplyr, Arrow, DuckDB native, and SQLite union; `get_col_names()` was traced to 0 queries on a DuckDB lazy tbl and is not in `lazy_execution_entry_points()`, so the query-policy snapshots are correctly unchanged.

PR #444, opened 2026-09-05
Refuse a character share source from the staged query itself

### LINKED ISSUE #429: DuckDB calculates a share from a character source the eligible-type rule should refuse

sayuks:
Closes #429.

## What changed

The staged share's ratio multiplies each operand by `1` inside the existing
`as.double()` casts. dbplyr renders that as `* 1.0`, which binds by type
rather than by value, so a dialect classified as refusing rejects a character
source whatever the column holds.

This is option 2 of the three the issue enumerates, per the Agent Brief.
`probe_share_dialect()`, `share_dialect_verdict()`, ADR 0020's exemption 2,
`?share_of_parent`'s *Lazy execution boundaries*, and CONTEXT.md's
*Converting dialect* are all unchanged — the change makes the existing
contract true rather than restating it. No dialect is reclassified, so the
numeric shares that work on DuckDB today still work.

The casts stay, and the commit says why: without them PostgreSQL and RSQLite
divide an integer pair as integers, and the multiplication alone leaves a
DuckDB `DECIMAL(18,3)` source at `DECIMAL(18,4)` and a `BIGINT` at
`DECIMAL(21,1)`. That second measurement is what chose
`as.double(x * 1) / as.double(y * 1)` over the bare `(x * 1) / (y * 1)` the
investigation note recommended equally, and it is recorded there as a
revisions section.

## Verified

- On DuckDB, a Parent share and a Total share over a character source raise at
  `collect()` for `"1", "2", "3"` and for `"n", "m", "o"` alike, with a Binder
  Error naming `*(VARCHAR, DECIMAL(2,1))` — a binding refusal, not a
  value-dependent conversion one.
- Integer, double, and local/dtplyr results are unchanged.
- All fifteen fallback simulators render both guards, so the assertion that
  failed under the bare `mul1` form holds; it gains a second assertion for the
  multiplication, each naming the executed measurement it stands for.
- Full suite under `NOT_CRAN=true MARGINPLYR_REQUIRED_SUGGESTS=""`: no
  failures. `lintr::lint_package()` clean after `pkgload::load_all(".")`.
  `roxygen2::roxygenise()` produced no diff.

### DIFF
--- a/R/share.R
+++ b/R/share.R
@@ -2643,15 +2643,13 @@ apply_joined_shares <- function(result,
               is.na(!!margin_column_pronoun(denominator)) |
               (!!margin_column_pronoun(denominator)) == 0,
             NA_real_,
-            # `* 1`, which dbplyr renders `* 1.0`, is what makes a dialect
-            # classified as refusing reject a character source: the casts alone
-            # accept a numeric-looking character column on DuckDB (#429). The
-            # casts stay because they are what makes the ratio a double —
-            # without them PostgreSQL and RSQLite divided an integer pair as
-            # integers, and the multiplication alone leaves a DuckDB `DECIMAL`
-            # source decimal. What the probe measures is an aggregate while
-            # this depends on arithmetic, so a dialect where those two diverge
-            # reproduces #429 here.
+            # `* 1`, which dbplyr renders `* 1.0`, is what refuses a character
+            # source on a dialect classified as refusing: the cast alone
+            # accepts a numeric-looking character column on DuckDB (#429). The
+            # cast is not made redundant by it — the cast is what makes the
+            # ratio a double whatever the source's type. The verdict this rests
+            # on is measured with an aggregate while this is arithmetic, so a
+            # dialect answering the two differently reproduces #429 here.
             as.double((!!margin_column_pronoun(source)) * 1) /
               as.double((!!margin_column_pronoun(denominator)) * 1)
           )
--- a/R/share.R
+++ b/R/share.R
@@ -2643,8 +2643,17 @@ apply_joined_shares <- function(result,
               is.na(!!margin_column_pronoun(denominator)) |
               (!!margin_column_pronoun(denominator)) == 0,
             NA_real_,
-            as.double(!!margin_column_pronoun(source)) /
-              as.double(!!margin_column_pronoun(denominator))
+            # `* 1`, which dbplyr renders `* 1.0`, is what makes a dialect
+            # classified as refusing reject a character source: the casts alone
+            # accept a numeric-looking character column on DuckDB (#429). The
+            # casts stay because they are what makes the ratio a double —
+            # without them PostgreSQL and RSQLite divided an integer pair as
+            # integers, and the multiplication alone leaves a DuckDB `DECIMAL`
+            # source decimal. What the probe measures is an aggregate while
+            # this depends on arithmetic, so a dialect where those two diverge
+            # reproduces #429 here.
+            as.double((!!margin_column_pronoun(source)) * 1) /
+              as.double((!!margin_column_pronoun(denominator)) * 1)
           )
         )
       )
--- a/R/share.R
+++ b/R/share.R
@@ -2643,17 +2643,16 @@ apply_joined_shares <- function(result,
               is.na(!!margin_column_pronoun(denominator)) |
               (!!margin_column_pronoun(denominator)) == 0,
             NA_real_,
-            # `* 1L` renders as `* 1` and binds the source by type, which is
-            # what refuses a character one on a dialect classified as refusing;
-            # the cast alone accepts a numeric-looking character column on
-            # DuckDB (#429). `1L` and not `1`: a double literal widens a DuckDB
-            # `DECIMAL(18,2)` to `DECIMAL(18,3)`, which overflows at the
-            # declared type's own maximum. The denominator carries the source's
-            # type, so a second copy there refuses nothing the source has not
-            # already refused. The cast stays regardless — it is what makes the
-            # ratio a double. The verdict this rests on is measured with an
-            # aggregate while this is arithmetic, so a dialect answering the
-            # two differently reproduces #429 here.
+            # `* 1L` binds the source by type, which is what refuses a
+            # character one on a dialect classified as refusing; the cast alone
+            # accepts a numeric-looking character column there (#429). The
+            # literal is `1L` because a double one changes the source's
+            # declared type, which `DuckDB shares a source at its declared
+            # type's maximum` fails on. The denominator holds the source's
+            # type, so a second copy there refuses nothing new. The cast stays:
+            # it is what makes the ratio a double. The verdict this rests on is
+            # measured with an aggregate while this is arithmetic, so a dialect
+            # answering the two differently reproduces #429 here.
             as.double((!!margin_column_pronoun(source)) * 1L) /
               as.double(!!margin_column_pronoun(denominator))
           )
--- a/R/share.R
+++ b/R/share.R
@@ -2643,15 +2643,19 @@ apply_joined_shares <- function(result,
               is.na(!!margin_column_pronoun(denominator)) |
               (!!margin_column_pronoun(denominator)) == 0,
             NA_real_,
-            # `* 1`, which dbplyr renders `* 1.0`, is what refuses a character
-            # source on a dialect classified as refusing: the cast alone
-            # accepts a numeric-looking character column on DuckDB (#429). The
-            # cast is not made redundant by it — the cast is what makes the
-            # ratio a double whatever the source's type. The verdict this rests
-            # on is measured with an aggregate while this is arithmetic, so a
-            # dialect answering the two differently reproduces #429 here.
-            as.double((!!margin_column_pronoun(source)) * 1) /
-              as.double((!!margin_column_pronoun(denominator)) * 1)
+            # `* 1L` renders as `* 1` and binds the source by type, which is
+            # what refuses a character one on a dialect classified as refusing;
+            # the cast alone accepts a numeric-looking character column on
+            # DuckDB (#429). `1L` and not `1`: a double literal widens a DuckDB
+            # `DECIMAL(18,2)` to `DECIMAL(18,3)`, which overflows at the
+            # declared type's own maximum. The denominator carries the source's
+            # type, so a second copy there refuses nothing the source has not
+            # already refused. The cast stays regardless — it is what makes the
+            # ratio a double. The verdict this rests on is measured with an
+            # aggregate while this is arithmetic, so a dialect answering the
+            # two differently reproduces #429 here.
+            as.double((!!margin_column_pronoun(source)) * 1L) /
+              as.double(!!margin_column_pronoun(denominator))
           )
         )
       )
--- a/investigation/refusing-a-character-share-source-in-the-staged-query.md
+++ b/investigation/refusing-a-character-share-source-in-the-staged-query.md
@@ -1,7 +1,7 @@
 # Refusing a character share source from the staged query itself
 
 Investigated: 2026-09-05
-Revised: 2026-09-05 — #429
+Revised: 2026-09-05 — `R/share.R`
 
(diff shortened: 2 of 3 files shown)


sayuks:
> *This was generated by AI.*

## Review record — `mattpocock-skills:code-review`, one round

Run against `git diff main...HEAD` at `fe0ee52`, two axes in parallel. Standards
sources: `AGENTS.md` (*Code comments*, *Linting*, *Dependency metadata*,
*Release matrix*, *Queries against a lazy input*, *Investigation notes*),
`design/architecture.md`'s *Answering a review*, `investigation/README.md`,
`tests/testthat/helper-optional-backends.R`. Spec source: #429, its Agent Brief
authoritative. Answers landed in `20d0270`.

### Standards — 10 raised

| # | finding | disposition |
|---|---|---|
| 1 | The comment beside the ratio re-derives measurements the note owns, in the past tense | fixed in code |
| 2 | "the probe measures an aggregate while this is arithmetic" is a claim nothing established | rejected — the claim is not that the two diverged but that nothing ties one to the other, which is what this site rests on; #429's brief asks for it in these words. Reworded to state it as a present property of the site |
| 3 | The simulator-loop comment is a third copy of finding 1's argument | fixed in code |
| 4 | "The test above" cites a position, not a target | fixed in code |
| 5 | `expect_error()` with no pattern passes on a dropped table | fixed in code — the eligible-source case runs first, so a collect against the same table has to succeed. Not a message assertion: `DuckDB reports an ineligible share source against its summary` already argues why DuckDB's wording is not asserted |
| 6 | Duplicated Code — the new test repeats its sibling's shape (judgement) | rejected — the shape is `duckdb_test_connection()` plus `copy_to()` plus a local `summarize()` closure, which every DuckDB test in the file shares; extracting it would hide which source each test refuses |
| 7 | `Revised: … — R/share.R` points at an artifact, against *Supersession* and *Pointers* | fixed in prose |
| 8 | The revisions section makes present-tense claims about the package | fixed in prose |
| 9 | "all fifteen simulators `available_simulators()` names" is false — it filters a list the test supplies | fixed in prose |
| 10 | Positional pointers where headings exist (judgement) | fixed in prose |

### Spec — 1 raised, 8 acceptance criteria checked

Criteria 1–5, 7, 8 satisfied, verified by the reviewer independently:
both character sources raise a DuckDB *BINDER* error for Parent and Total;
integer and double sources unchanged; own-denominator and
missing/zero-denominator rows unchanged; `probe_share_dialect()` untouched and
`test-sent-queries.R` green; `skip_if_suggest_absent("duckdb", "DBI")` with DBI
a Suggest and not a backend, confirmed by `verify-suite-coverage.R` passing.
Out-of-scope list clean — no ADR 0020, `?share_of_parent`, CONTEXT.md, or probe
change. No requirement wrongly implemented.

**Criterion 6 — the fallback-simulator assertion "rewritten to assert the
behaviour it stood for" — partially satisfied.** The criterion's premise is
void: it rested on that assertion being the only test that failed under the
substitution, and the committed form keeps the cast, so it does not fail. The
assertion is answered rather than rewritten. What it stood for is now asserted
where it can execute — the cast's integer-division property by `RSQLite
executes portable Parent shares end to end`, which compares an integer-source
share against a local result, and the multiplication by this branch's DuckDB
test — and the loop's comment names both. A simulator has no database behind
it, so a render-level assertion is what that site can hold.

sayuks:
> *This was generated by AI.*

## Second review round — `/code-review`, 2 findings, both fixed in `c200fb1`

Both were confirmed by measurement before being answered, and both are
regressions this branch introduced: the queries they break returned results on
`main`.

### 1. `* 1.0` overflows a source at its declared type's maximum — fixed

R's `1` is a double, so dbplyr rendered `* 1.0` and DuckDB grew the source's
scale by one without growing its width. Measured on duckdb 1.5.5 over a column
holding its declared type's maximum:

| source type | `x` | `x * 1.0` | `x * 1` |
|---|---|---|---|
| `DECIMAL(18,2)` | answered | `DECIMAL(18,3)`, `Out of Range Error: Overflow in multiplication of DECIMAL(18)` | `DECIMAL(18,2)`, answered |
| `DECIMAL(38,10)` | answered | `DECIMAL(38,11)`, **came back missing with no diagnostic** | `DECIMAL(38,10)`, answered |

The `DECIMAL(38,10)` case is worse than the review reported: it raises nothing,
so the share would have been `NA`.

The fix is `1L`. It renders `* 1`, changes no type — `HUGEINT` stays `HUGEINT`
where `* 1.0` made it `DECIMAL(21,1)` — and refuses a `VARCHAR` at binding
identically, reporting `'*(VARCHAR, INTEGER_LITERAL)'` in place of
`'*(VARCHAR, DECIMAL(2,1))'`.

`DuckDB shares a source at its declared type's maximum` is the gate. Its table
is created in SQL because `copy_to()` cannot express a declared `DECIMAL`. Both
it and the simulator assertion — now `\\* 1[^.0-9]`, which a decimal literal
fails — were checked against the regressed form.

### 2. The denominator's multiplication is redundant — fixed

Accepted as stated. The denominator is the source summary carried through a
join, so it holds the source's type; a dialect refusing one refuses the other.
Dropping it changed no result and no refusal, and halves how many operands the
widening above can reach.

### Not split into an issue

Judged in-scope for this branch: these are defects in the expression this
branch changed, not pre-existing ones, and the fix was a measured one-token
change rather than a decision across unmeasured dialects. The investigation
note's revisions section carries the measurements; its earlier table recorded
the widening but never asked what it costs, which is the gap that let this
through.

Full suite green under `NOT_CRAN=true MARGINPLYR_REQUIRED_SUGGESTS=""`;
`lintr::lint_package()` clean.

sayuks:
> *This was generated by AI.*

## Third review round — two axes, answers in `47383af`

Same fixed point (merge-base with `main`). Standards raised 3, Spec raised 1
plus one residual outside this branch.

### Standards

| # | finding | disposition |
|---|---|---|
| 1 | The overflow measurement is stated in the ratio's comment, the new test's header, and the simulator loop's — three copies of what the note owns, and the shape `20d0270` deleted from that same comment | **fixed in code.** The ratio's comment and the simulator loop's now name `DuckDB shares a source at its declared type's maximum`, the test that fails when the fact drifts. That test's header keeps the arithmetic, because it is what the test is about |
| 2 | `Revised: 2026-09-05 — #429` is unreachable by the grep `investigation/README.md` prescribes, and round 2's reason for it is contradicted by `convenience-api-proposal.md`'s `Revised: 2026-08-07 — R/share.R, ADR 0018` | **fixed in prose.** Round 2's finding 7 was wrong on the precedent; `Revised:` names `R/share.R` again |
| 3 | The maximum-value guard exists only for DuckDB `DECIMAL(18,2)`; T-SQL's `DECIMAL(p,s) * INT` rule is untested (judgement, no standard breached) | **rejected** — the note's *What was not measured* records exactly this, and dialects beyond DuckDB, PostgreSQL, and SQLite are out of scope per #429's brief. The simulators render and never execute, which the loop's comment says |

The reviewer verified every claim the comments make by execution, including
that the denominator carries the source's type (`build_total_denominator()`'s
`transmute()` and `build_parent_denominator()`'s `select()` only rename it).

### Spec

Criteria 1–5, 7, 8 satisfied, all exercised by running: the reviewer exported
`main` to a second tree and ran an identical battery on both — local,
data.table, dtplyr, DuckDB; integer and double measures; `rollup(g1, g2)` —
byte-identical at 15 digits, and identical across `BIGINT, HUGEINT, INTEGER,
DECIMAL(18,2), DECIMAL(38,10), FLOAT, DOUBLE, UBIGINT`. A nine-type source
battery (integer, double, integer64, difftime, Date, POSIXct, factor,
character, logical) confirmed every non-plain type is refused before the ratio,
so `x * 1L` never sees a classed source.

**Criterion 6, the simulator assertion — raised for the third time, rejected
with evidence.** A simulator has no database behind it, so no assertion at that
site can execute; there is no behavioural formulation available there. What the
criterion is for is answered instead: both assertions were checked against the
regressed form and both fail on it (15 failures on `\\* 1[^.0-9]`), so they gate
the defect they exist for rather than merely describing the code. The
properties themselves are asserted where they execute, and the loop's comment
names those two tests.

### Filed rather than fixed here — #446

The Spec reviewer found a residual of #429's shape that this branch does not
reach and that is **pre-existing on `main`**: when every occurrence is its own
denominator, `apply_joined_shares()` returns `rlang::expr(1.0)` before the
ratio is built, so `grouping_sets(grouping_set())` with a character source
returns `1` on DuckDB while the local path refuses it. Reproduced at `47383af`
and on `main` alike. It is `ready-for-human` because whether the constant
branch should consult eligibility at all is an ADR 0020 question, not a
mechanical fix.

Full suite green under `NOT_CRAN=true MARGINPLYR_REQUIRED_SUGGESTS=""`;
`lintr::lint_package()` clean.

PR #445, opened 2026-09-05
Give the dialect probe's control its own reading of a result

### LINKED ISSUE #440: Every share on PostgreSQL is refused: the dialect probe's control rejects a bigint result

sayuks:
Closes #440.

## What was wrong

`probe_share_dialect_answer()` accepted a collected result only when
`is_share_source_type()` held of it. That predicate is the *eligible-type rule*
for a share source, and it requires `!is.object(value)`. PostgreSQL's
`SUM(integer)` is `bigint`, which RPostgres maps to `bit64::integer64` — a
classed double. The control therefore read as "the question could not be put
here", `probe_share_dialect()` returned `"unknown"`, and every share on
PostgreSQL was refused whatever the source's type. Nothing was cached, so both
queries were re-sent on every share request on the dialect.

Two questions had been conflated. The eligible-type rule is about a caller's
data. The control asks about the connection: did this query come back with the
number the scaffolding selected?

## What changed

`probe_share_dialect_holds()` is where each of the two questions is now
written. The probe asks that a number came back, which is the conversion. The
control asks for the number the scaffolding selected, in whatever type the
driver mapped the database's type to, and tests the type as well as the value —
`==` alone is the weaker question of what R coerces equal to that number. The
comparison is wrapped, because a class that passes the type test can raise
inside `==` and nothing between there and the caller catches it.

`share_probe_scaffold()` holds the scaffolding and the number it selects
together, and is read outside every wrap whose value decides the verdict: an
error from a marginplyr frame is a defect, and reading it as a dialect that
could not answer would hide it behind a refused share.

`is_share_source_type()` is unchanged, and its header's "two raising sites" is
true again. So are `probe_share_dialect()`'s structure, the verdict vocabulary,
the cache policy, ADR 0020's exemption 2, and ADR 0027's two purposes.

## Measured

Against PostgreSQL 17.11 (Homebrew) through RPostgres, dbplyr 2.6.0, dplyr
1.2.1, across all four `bigint` mappings the driver offers — `bit64::integer64`,
`integer`, `numeric`, and `character`:

- `probe_share_dialect(con)` returns `"refuses"` and is recorded, so a second
  share request sends only its result query.
- Parent and Total shares over a double source calculate, as does a Parent
  share over `n()`, whose `bigint` result is the value that used to refuse.
- A character source raises with PostgreSQL's own
  `function sum(text) does not exist`, for a numeric-looking source
  (`c("1","2","3")`) and a non-numeric one alike — earlier than the `= 0.0`
  comparison #429 is about.

DuckDB's `"refuses"` and SQLite's `"converts"` are unchanged. Full suite green
under `NOT_CRAN=true`; `lintr::lint_package()` clean.

## Tests

All in `test-share-backends.R`, none requiring a live server or any member of
`optional_backends()`:

- the control reads a classed number, an `integer`, a `double` and a character
  `"1"` as the scaffolding's number, and reads `7`, `"x"`, `NA`, and five values
  R coerces equal to `1` — a logical, a factor, a Date, a length-1 list, a raw —
  as no answer, along with a class whose `==` raises;
- the probe reads a driver-classed number as the conversion;
- a defect raised by `share_probe_scaffold()` reaches the caller instead of
  being reported as a dialect that cannot answer.

The classed value carries a name no package defines methods for, deliberately: a
double built by hand does not hold a `bit64::integer64` bit pattern, so with
bit64 loaded `structure(1, class = "integer64") == 1` is `FALSE` where
`bit64::as.integer64(1) == 1` is `TRUE`, and borrowing the name would assert the
driver's arithmetic rather than this reading.

Eight review rounds ran; the record is in a comment below.

### DIFF
--- a/R/share.R
+++ b/R/share.R
@@ -2409,10 +2409,9 @@ probe_share_dialect <- function(con) {
 # `purpose` is the name this query is recorded under (ADR 0027). `control` says
 # which of the two questions this is, which is what decides the reading below.
 probe_share_dialect_answer <- function(con, expr, purpose, control = FALSE) {
-  # Bound outside the wrap below for the reason `probe_share_dialect_holds()`
-  # binds its number outside its own: what these wraps read as a connection
-  # that could not be asked is an error, and marginplyr's own frames raise
-  # errors that are defects.
+  # Bound outside the wraps below for the reason `probe_share_dialect_holds()`
+  # binds its number outside its own: what each of them reads as something the
+  # connection did is an error.
   scaffold <- share_probe_scaffold()$sql
   query <- tryCatch(
     dplyr::summarize(dplyr::tbl(con, scaffold, vars = "z"), p = !!expr),
--- a/R/share.R
+++ b/R/share.R
@@ -2409,11 +2409,13 @@ probe_share_dialect <- function(con) {
 # `purpose` is the name this query is recorded under (ADR 0027). `control` says
 # which of the two questions this is, which is what decides the reading below.
 probe_share_dialect_answer <- function(con, expr, purpose, control = FALSE) {
+  # Bound outside the wrap below for the reason `probe_share_dialect_holds()`
+  # binds its number outside its own: what these wraps read as a connection
+  # that could not be asked is an error, and marginplyr's own frames raise
+  # errors that are defects.
+  scaffold <- share_probe_scaffold()$sql
   query <- tryCatch(
-    dplyr::summarize(
-      dplyr::tbl(con, share_probe_scaffold()$sql, vars = "z"),
-      p = !!expr
-    ),
+    dplyr::summarize(dplyr::tbl(con, scaffold, vars = "z"), p = !!expr),
     error = function(cnd) NULL
   )
   if (is.null(query)) {
--- a/R/share.R
+++ b/R/share.R
@@ -2428,23 +2428,33 @@ probe_share_dialect_answer <- function(con, expr, purpose, expected = NULL) {
     return("raised")
   }
   value <- answer$value
-  if (
-    !is.data.frame(value) ||
-      nrow(value) != 1L ||
-      ncol(value) != 1L ||
-      !is.numeric(value[[1L]])
-  ) {
+  if (!is.data.frame(value) || nrow(value) != 1L || ncol(value) != 1L) {
     return("unanswerable")
   }
-  # Not `is_share_source_type()`, which is the eligible-type rule about a
-  # caller's summary and rejects a classed value. What class the driver mapped
-  # the database's type to is a fact about the driver: PostgreSQL's
-  # `SUM(integer)` is `bigint`, which RPostgres maps to `bit64::integer64`, a
-  # classed double that `is.numeric()` holds of (#440).
-  if (!is.null(expected) && !isTRUE(value[[1L]] == expected)) {
+  if (!probe_share_dialect_holds(value[[1L]], expected)) {
     return("unanswerable")
   }
   "answered"
+}
+
+# Whether the one value a query came back with is the answer its caller asked
+# for. Not `is_share_source_type()`: that is the eligible-type rule about a
+# caller's summary, and it rejects a classed value, so it also rejects what a
+# driver mapped a database type to. RPostgres alone offers four mappings for
+# the `bigint` PostgreSQL sums an integer to -- `bit64::integer64`, `integer`,
+# `numeric`, and `character`, the first a classed double and the last not a
+# number at all -- and the caller chooses between them when they connect
+# (#440).
+#
+# So the probe asks only that a number came back, which is the conversion. The
+# control names the number the scaffolding selected and asks for that, in
+# whatever type the mapping produced: `==` compares across those four, where a
+# type test admits three of them.
+probe_share_dialect_holds <- function(value, expected) {
+  if (is.null(expected)) {
+    return(is.numeric(value))
+  }
+  isTRUE(value == expected)
 }
 
 # The table-free scaffolding both questions are asked against, and the number
--- a/R/share.R
+++ b/R/share.R
@@ -2378,7 +2378,8 @@ probe_share_dialect <- function(con) {
     control <- probe_share_dialect_answer(
       con,
       quote(sum(z, na.rm = TRUE)),
-      purpose = "share_dialect_control"
+      purpose = "share_dialect_control",
+      expected = share_probe_scaffold()$number
     )
     if (identical(control, "answered")) {
       return("refuses")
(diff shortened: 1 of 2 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, Standards and Spec axes, against the merge-base
with `main` (`git diff main...HEAD`). Eight rounds; each round read the commit
above it and its answer is the commit below.

| round | snapshot read | raised | answered by |
|---|---|---|---|
| 1 | `15e13c0` | 3 documented-standard breaches, 2 baseline smells; Spec: 2 unaddressed *Verify while implementing* items, 1 untested behaviour change | `34cd34f` |
| 2 | `34cd34f` | 2 hard (correctness), 2 judgement | `7e6517f` |
| 3 | `88142dc`'s parent | 0 hard, 2 judgement | `88142dc` |
| 4 | `88142dc` | 0 hard, 2 judgement | `96cf749` |
| 5 | `96cf749` | 1 hard, 1 judgement (rejected) | `3e1a283` |
| 6 | `3e1a283` | 1 hard, 1 judgement | `f7168e7` |
| 7 | `f7168e7` | 1 judgement | `5a7108a` |
| 8 | `5a7108a` | no findings; branch clean | — |

Two findings changed what the code does, and each landed with the test that
fails without it:

- **Round 2.** Keeping `is.numeric()` on the control left #440's defect standing
  for one of the four mappings RPostgres offers for `bigint`: a connection
  opened with `bigint = "character"` came back with `"1"`, and every share on
  the dialect was still refused. Measured on PostgreSQL 17.11; all four
  mappings now answer `"refuses"`.
- **Round 5.** `share_probe_scaffold()` was read inside the wrap that builds the
  query, so a defect in marginplyr's own frame read as a dialect that could not
  be asked. This branch introduced that site — `main` had the scaffolding SQL
  inline, which is not a marginplyr frame.

One finding was rejected, with its evidence: round 5's observation that
`is.numeric()` is an internal generic, so a class carrying a method that raises
escapes the type test the same way. `is.numeric.Date`, `is.numeric.difftime`
and `is.numeric.POSIXt` are base R's only methods; a scan of 294 installed
namespaces adds `s2::is.numeric.s2_cell` and zoo's two; none raises, and
`is.character` is not an internal generic at all. Nothing reaches it.

Every other finding was about prose, and rounds 5 through 7 each found exactly
one claim that measured false or over-reached in a comment written to answer
the round before — the execute wrap does not read an error as a connection that
could not be asked, `haven::labelled` does not raise comparing a numeric
instance, and `record_sent_query()`'s wrap decides no verdict. Each was deleted
rather than reworded, per *Answering a review*.

Round 8 re-measured every comment claim the branch adds and found each true,
and re-derived the fix independently: DuckDB `"refuses"` and SQLite `"converts"`
live, `simulate_*` `"unknown"`, all four PostgreSQL mappings `"refuses"`, and
every non-scaffolding control value `"unknown"`. Full suite green under
`NOT_CRAN=true` with `MARGINPLYR_REQUIRED_SUGGESTS` naming every installed
backend (0 skips); `lintr::lint_package()` clean; `_snaps/` unchanged.

PR #447, opened 2026-09-05
Name a rewritten summary the way dplyr names it

### LINKED ISSUE #439: An assigned summary name diverges from dplyr's for an expression long enough to abbreviate

## Summary

#430 names an unnamed summary marginplyr rewrites, from the expression the
caller wrote, through `rlang::as_label()`. dplyr names an unnamed summary
through `expr_as_label()`, which is `rlang::as_label()` with rlang's infix
labelling suppressed through an undocumented option. The two abbreviate a long
expression differently, so a rewritten summary long enough to abbreviate takes
a column name dplyr would not have given the same expression.

`?summarize_with_margins` documents `...` as "Name-value pairs as used in
`dplyr::summarize()`", which fixes dplyr's naming as the contract, and #430 was
fixed to that contract.

## Reproduction

```r

data <- data.frame(g = c("a", "a", "b"), x = c(1, 3, 5), y = c(2, 4, 6))

# One expression, spelled once. A rewrite reaches it through `pick()`.
names(summarize_with_margins(
  data,
  sum(pick(x)[[1]]) + 1 + sum(y) + 2 + mean(x) + 3 + mean(y) + 4 + median(x) + 5,
  .grouping = rollup(g)
))
#> [1] "g"       "... + 5"

names(summarize(
  data,
  sum(pick(x)[[1]]) + 1 + sum(y) + 2 + mean(x) + 3 + mean(y) + 4 + median(x) + 5,
  .by = g
))
#> [1] "g"     "+..."
```

The same holds for a rewrite reached through a grouping helper, and on a
backend holding `native_grouping_sets` as well as locally — duckdb returns
`... + grouping_bit(g)` for

```r
sum(x) + 1 + sum(y) + 2 + mean(x) + 3 + mean(y) + 4 + max(x) + grouping_bit(g)
```

where dplyr's own labeller writes `+...`.

The divergence is the one `name_rewritten_summary_dots()`'s header already
names — "dplyr asks rlang for a variant spelling of that abbreviation through
an internal option" — recorded there as costing nothing. It costs a column
name.

## What was measured and is not a bug

The first statement of this issue reported two properties of the labeller.
Both were measured against `9bc0ae2` and neither is a defect:

1. **The abbreviation itself.** `rlang::as_label()` abbreviates a long call to
   `<fn>(...)`, and the assigned name is that abbreviation — but
   `dplyr::summarize()` abbreviates the same call to the same `long(...)`.
   marginplyr matches the contract.
2. **Following the caller's literal spelling.** `pick(x)` and `dplyr::pick(x)`
   produce two different column names — and so they do under plain
   `dplyr::summarize()`: `nrow(pick(x, y))` and `nrow(dplyr::pick(x, y))` are
   two names there too. marginplyr matches the contract. The reproduction that
   was filed for this — `range_frame(pick(x))` — no longer reproduces at all,
   ADR 0028 having landed: it now expands to `lo` and `hi`, and no assigned
   name reaches the result.

Two further observations from that statement follow from the assigned name
rather than standing on their own. A diagnostic quoting an assigned name
quotes whatever the name is, so it is repaired by repairing the name; and
`check_internal_summary_names()`'s remedy is not reachable by an assigned name
at all, its internal names being the `..marginplyr_key_` columns.

## Relationship to ADR 0022

ADR 0022 considered and rejected *Reproduce `expr_as_label()` faithfully,
undocumented option included*, on the ground that what it covers is
unobservable: a label dplyr truncated renders the same whichever expression it
came from. That argument is about a **condition label**, and it holds there.
An assigned name is a column name, so the same divergence is observable, and
that ADR's rejection does not reach it.

## What is unmeasured

- `expr_as_label()`'s other branch — a data pronoun deparsed rather than
  labelled — appears unreachable here, since a summary that is a bare `.data$x`
  is not rewritten and so is never assigned a name, while a pronoun inside a
  larger expression is not the expression the labeller reads. This is argued,
  not measured.
- Whether any assigned name in the test suite or the documentation changes
  under the fix. Every one measured so far is short enough not to abbreviate.

## Measured against

marginplyr at `9bc0ae2`, R 4.6.1, dplyr 1.2.1, rlang as installed, duckdb
1.5.5.

sayuks:
Closes #439.

An unnamed summary marginplyr rewrites was named through `rlang::as_label()`,
while `dplyr::summarize()` names one through `expr_as_label()` — `as_label()`
with rlang's infix labelling suppressed through an undocumented option. The two
agree until an expression is long enough for rlang to abbreviate it, where
dplyr writes `+...` and `as_label()` writes `... + 5`. `?summarize_with_margins`
documents `...` as dplyr's name-value pairs, so the divergence reached the
caller as a column name.

`dplyr_auto_name()` sets the option dplyr sets, and
`name_rewritten_summary_dots()` labels through it. Nothing else about who gets
named changes.

The test computes both names rather than writing either out — the assigned name
and `dplyr::summarize()`'s own name for the same expression — and asserts
separately that the expression is still one the two labellers disagree about,
so an rlang that dropped the option fails the first and one that stopped
abbreviating at that width fails the second. Both rewrite routes are covered,
and the grouping-helper one is asserted on the native adapter over
`dbplyr::simulate_postgres()`.

ADR 0022 rejected reproducing `expr_as_label()` for a **condition label**, on
the ground that the divergence there is unobservable; that ground does not
reach a column name, and its amendment records the split. ADR 0028's closing
paragraph named two things it left to #439 — one turned out to be dplyr's own
behaviour and is dropped, the other is this.

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -346,8 +346,9 @@ report_branch_warnings <- function(conditions) {
 # The label dplyr quotes an argument by, reproduced so that the expression
 # marginplyr handed it can be recognised in a context it rendered. Plain
 # `rlang::as_label()` under dplyr's `name = ` convention is what this uses,
-# rather than dplyr's `expr_as_label()`; ADR 0022 is authoritative for where
-# the two diverge and for why neither divergence is observable.
+# rather than dplyr's `expr_as_label()`. What this site writes is a condition
+# label, and that is what ADR 0022 decides; a name a caller reads back falls
+# the other way, under the same ADR's amendment for #439.
 summary_argument_labels <- function(dots) {
   arg_names <- names(dots)
   if (is.null(arg_names)) {
--- a/R/summary-selections.R
+++ b/R/summary-selections.R
@@ -361,6 +361,18 @@ new_summary_arguments <- function(dots,
   list(dots = dots, labels = labels, assigned_names = assigned_names)
 }
 
+# The name `dplyr::summarize()` gives an unnamed argument, for an argument
+# marginplyr names in dplyr's place. dplyr's `expr_as_label()` is
+# `rlang::as_label()` with rlang's infix labelling suppressed through an
+# undocumented option, and the two spellings differ once an expression is long
+# enough to abbreviate. Setting that option is the whole of the reproduction:
+# the other branch of `expr_as_label()` deparses a bare data pronoun, which no
+# expression reaching the one caller here can be (ADR 0022, amended).
+dplyr_auto_name <- function(expr) {
+  rlang::local_options(`rlang:::use_as_label_infix` = FALSE)
+  rlang::as_label(expr)
+}
+
 # The name dplyr would have given each unnamed summary marginplyr rewrites,
 # read from what the caller wrote rather than from the rewrite. The named dots
 # arrive beside the Assigned summary name each one took, `NA` where none was
@@ -392,11 +404,10 @@ new_summary_arguments <- function(dots,
 # and `range_frame(pick(v))` has to not be, which is a question about the
 # value's type. ADR 0028 is where that one is answered.
 #
-# `rlang::as_label()` is `rlang::quos_auto_name()`'s own labeller, so the name
-# is dplyr's up to the width at which rlang abbreviates a long expression:
-# dplyr asks rlang for a variant spelling of that abbreviation through an
-# internal option, and neither spelling stands for an expression the caller
-# could read back.
+# The label is `dplyr_auto_name()` and not `rlang::as_label()`, because what is
+# written here is a column name rather than a condition label: ADR 0022's
+# rejection of reproducing dplyr's spelling reaches the label alone, and its
+# amendment for #439 says so.
 name_rewritten_summary_dots <- function(original, resolved) {
   stopifnot(length(original) == length(resolved))
   arg_names <- rlang::names2(resolved)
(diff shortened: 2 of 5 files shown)


sayuks:
## Review record

Two rounds through `mattpocock-skills:code-review`, both axes, against the merge-base with `main`.

### Round 1 — `29d0939`

| Finding | Disposition |
|---|---|
| Standards: `summary_argument_labels()`'s comment cited ADR 0022 for "why neither divergence is observable", which this branch's amendment falsifies | fixed in prose (`c069392`) — the comment is rescoped to the condition label and says which way that site falls |
| Standards: the rlang/dplyr divergence fact stood in three copies (the R header, the ADR amendment, the test comment) | fixed in prose (`c069392`) — the test comment's restatement of the mechanism is deleted |
| Standards: the ADR amendment's closing paragraph restated the test's shape | fixed in prose (`c069392`) — it states the cost of depending on the option instead |
| Standards: Duplicated Code — the same assertion triple written out for each of the two expressions | fixed in code (`c069392`) — one local `expect_dplyr_name()` checker |
| Standards: `dplyr_auto_name()` belongs in `R/utils.R` under *Static expression reading* | rejected — that section scopes utils to writing a **call part** back into a diagnostic, and this predicts the output name dplyr assigns a whole summary expression, which is *Summary selection*'s. Round 2 re-checked it against the doc as written and agreed |
| Spec: the native assertion covered only the grouping-helper expression, where the acceptance criterion reads "the same is asserted" | fixed in code (`c069392`) — the native assertion moved into the checker, so each expression is asserted on both adapters |

### Round 2 — `c069392`

No findings. Reverting `dplyr_auto_name()` to `rlang::as_label()` was measured to fail four assertions (two expressions × local and native), so the native half is load-bearing rather than incidental.

PR #448, opened 2026-09-05
Bind the share source where the plan builds no ratio

### LINKED ISSUE #446: A grouping plan with no joined set computes a share from an ineligible source on a lazy backend
> *This was generated by AI during a review of #444.*

## Summary

When every occurrence in a Grouping plan is its own denominator,
`apply_joined_shares()` returns before building the ratio and the share is the
constant `1.0`. Nothing in the staged query then touches the source summary, so
a lazy backend calculates a share from a source the eligible-type rule refuses,
and raises nothing. The local path refuses the same call.

This is the residual of #429 in a corner that issue's fix does not reach: #429
made the ratio refuse an ineligible source, and this is the path with no ratio.
It is **pre-existing on `main`** — the early return is unchanged by #444 — and
was found by that PR's third review round.

## Reproduction

```r

df <- data.frame(a = c("x", "x", "y"), s = c("1", "2", "3"))
tb <- copy_to(con, df, "g")

f <- function(src) {
  summarize_with_margins(
    src,
    total = max(s),
    ps = share_of_total(total),
    .grouping = grouping_sets(grouping_set())
  )
}

f(tb) |> collect()
f(df)
```

```
# DuckDB:
  total ps
1     3  1

# local:
! Total share `ps` requires source summary `total` to be a plain integer or
  double scalar.
i Detected type character.
```

Measured 2026-09-05 at `47383af` against duckdb 1.5.5, dbplyr 2.6.0, dplyr
1.2.1. `main` behaves identically.

## Root cause

`apply_joined_shares()` in `R/share.R`:

```r
      if (length(joined_ids) == 0L) {
        return(rlang::expr(1.0))
      }
```

`joined_ids` is empty when no occurrence has a target, which is what
`grouping_sets(grouping_set())` produces for a Total share — the only set is
the Grand total set, and it is its own denominator. The share is correct
arithmetically: a set's share of itself is one whatever the source holds. What
is wrong is that the contract in `?share_of_parent`'s *Lazy execution
boundaries* says otherwise:

> The eligible-type rule is also enforced on every backend, and no backend
> calculates a share from a source it has shown to be ineligible.

The local path enforces it from the result before any share, so the two
backends disagree.

## Why this is `ready-for-human`, not `ready-for-agent`

Which of two things this is has not been decided, and the answer changes the
fix:

1. **A rule that must hold whatever the plan.** Then the constant branch has to
   consult the source's eligibility even though it does not read its value, and
   the question is what expression does that on a dialect without adding a term
   to every share. #429's option 2 argument applies again, and its measurements
   in `investigation/refusing-a-character-share-source-in-the-staged-query.md`
   are the starting point.
2. **A rule about the ratio.** Then `1.0` is a correct answer for an
   own-denominator set, the local path is the one being stricter than the
   contract requires, and what changes is `?share_of_parent`'s wording rather
   than any query.

Deciding needs a reading of what the eligible-type rule is for, which is an ADR
0020 question.

## What is unmeasured

- Which plans other than `grouping_sets(grouping_set())` reach the empty
  `joined_ids` branch, and whether a Parent share can.
- Whether dtplyr behaves as the local path or as DuckDB here.
- Whether the same asymmetry reaches the cardinality rule, not just the type
  rule.

sayuks:
Closes #446.

## What was wrong

When every occurrence in a Grouping plan is its own denominator,
`apply_joined_shares()` returned the constant `1.0` before mapping any
denominator or building any ratio. Nothing in the staged query then referenced
the source summary, so a refusing dialect was never asked to bind it: DuckDB
answered `1` for a character source that the local and dtplyr paths refuse.

This is the default shape of a Total share — no `.grouping` and no `.by` reach
it — and it is pre-existing on `main`, not a residual of #429. #429 made the
*ratio* refuse an ineligible source, and this is the path with no ratio.

## The fix

Triage settled that the eligible-type rule is a rule about the source, not
about the ratio, so the code is what is wrong. The constant branch now emits

```r
dplyr::if_else(is.na(<source> * 1L), 1.0, 1.0)
```

which applies to the source the same integer multiplication the ratio applies
— what a refusing dialect rejects for a character column whatever it holds —
while both arms give the constant the branch has always produced. The
condition is an `IS NULL`, which SQL never answers with `NULL`, so no third
value can leak from the `CASE` dbplyr renders without an `ELSE`.

Emitted on every backend, as the ratio's `* 1L` already is, and not
conditional on `.check_share_source`, which governs the dialect-verdict
refusal alone. The denominator mapping and the join stay skipped, so no term
is added to a share that has a denominator and the joined path's SQL is
unchanged.

## Verification

- New test `DuckDB refuses a character source for a share that needs no join`,
  beside the joined refusal's, guarded on `duckdb` and `DBI` alone. It covers
  a numeric-looking and a non-numeric character column, asserts a `tbl_lazy`
  from the verb and a non-`marginplyr_error` at `collect()`, asserts eligible
  double, zero, and all-`NA` sources still collect `1` as a double, and
  asserts `.check_share_source = FALSE` relaxes nothing.
- `Total shares need no join when every occurrence is the whole` keeps every
  assertion and gains no optional-backend requirement.
- `investigation/refusing-a-character-share-source-in-the-staged-query.md`
  gains a dated revisions section with the reachability, backend, SQL-form,
  and flag measurements.
- Full suite: `FAIL 0 | WARN 0 | SKIP 0 | PASS 6835` with `NOT_CRAN=true`.
  `lintr::lint_package()` clean after `load_all()`.

No change to `?share_of_parent`, ADR 0010, any vignette, or `NEWS.md` — the
rule they state is the one being restored.

### DIFF
--- a/R/share.R
+++ b/R/share.R
@@ -2672,7 +2672,18 @@ apply_joined_shares <- function(result,
       source <- pair$source
       denominator <- denominator_names[[source]]
       if (length(joined_ids) == 0L) {
-        return(rlang::expr(1.0))
+        # This branch builds no ratio, so this expression is the only thing
+        # binding the source's type, which is what a refusing dialect rejects a
+        # character source by (#446). Both arms are `1.0` and `is.na()` answers
+        # every value, so the constant this branch has always produced is
+        # unchanged for an eligible source.
+        return(rlang::expr(
+          dplyr::if_else(
+            is.na((!!margin_column_pronoun(source)) * 1L),
+            1.0,
+            1.0
+          )
+        ))
       }
       rlang::expr(
         dplyr::if_else(
--- a/R/share.R
+++ b/R/share.R
@@ -2674,10 +2674,9 @@ apply_joined_shares <- function(result,
       if (length(joined_ids) == 0L) {
         # No denominator was joined, so nothing else in the staged query names
         # the source: this is what binds it there, for the reason the ratio's
-        # own `* 1L` below gives (#446). Both arms are the constant this branch
-        # has always produced, and the condition is an `IS NULL`, which SQL
-        # never answers with `NULL`, so no third value leaks from the `CASE`
-        # dbplyr renders without an `ELSE`.
+        # own `* 1L` below gives (#446). Both arms are `1.0`, and the condition
+        # is an `IS NULL`, which SQL never answers with `NULL`, so no third
+        # value leaks from the `CASE` dbplyr renders without an `ELSE`.
         return(rlang::expr(
           dplyr::if_else(
             is.na((!!margin_column_pronoun(source)) * 1L),
--- a/R/share.R
+++ b/R/share.R
@@ -2672,11 +2672,12 @@ apply_joined_shares <- function(result,
       source <- pair$source
       denominator <- denominator_names[[source]]
       if (length(joined_ids) == 0L) {
-        # This branch builds no ratio, so this expression is the only thing
-        # binding the source's type, which is what a refusing dialect rejects a
-        # character source by (#446). Both arms are `1.0` and `is.na()` answers
-        # every value, so the constant this branch has always produced is
-        # unchanged for an eligible source.
+        # No denominator was joined, so nothing else in the staged query names
+        # the source: this is what binds it there, for the reason the ratio's
+        # own `* 1L` below gives (#446). Both arms are the constant this branch
+        # has always produced, and the condition is an `IS NULL`, which SQL
+        # never answers with `NULL`, so no third value leaks from the `CASE`
+        # dbplyr renders without an `ELSE`.
         return(rlang::expr(
           dplyr::if_else(
             is.na((!!margin_column_pronoun(source)) * 1L),
--- a/investigation/README.md
+++ b/investigation/README.md
@@ -101,10 +101,13 @@ overturned.
 1. Add one line to the header block:
 
    ```text
-   Revised: <YYYY-MM-DD> — investigation/<successor>.md
+   Revised: <YYYY-MM-DD> — <what established the correction>
    ```
 
-   one per revision, in date order after `Investigated:`.
+   one per revision, in date order after `Investigated:`. What follows the dash
+   is what step 2 names — usually `investigation/<successor>.md`, and the
+   artifact where the correction was established in code or configuration
+   rather than in another note.
 
 2. Add a section headed `## Revisions (<YYYY-MM-DD>)`, placed after the
    findings it amends, stating what changed and why. Name the note or artifact
--- a/investigation/README.md
+++ b/investigation/README.md
@@ -111,11 +111,13 @@ overturned.
    that established the correction.
 
    A second revision on a date the note already carries writes the issue
-   number into both — `Revised: <YYYY-MM-DD> (#<issue>) — <target>` and
-   `## Revisions (<YYYY-MM-DD>, #<issue>)` — so that a header line names one
-   section. Without it a reader has two entries and two sections and no way to
-   pair them, which is what `refusing-a-character-share-source-in-the-staged-query.md`
-   reached the day it was written: two investigations, one date, one target.
+   number into both — `(#<issue>)` after the date in step 1's line, and
+   `## Revisions (<YYYY-MM-DD>, #<issue>)` here — so that a header line names
+   one section. Without it a reader has two entries and two sections and no way
+   to pair them, which is what
+   `refusing-a-character-share-source-in-the-staged-query.md` reached the day
+   it was written: two investigations, one date, and the same thing named
+   after the dash.
 
 3. Leave the original findings in place, unedited. The point of the note is
    what was believed on its date; deleting the superseded text destroys the
--- a/investigation/README.md
+++ b/investigation/README.md
@@ -110,6 +110,13 @@ overturned.
    findings it amends, stating what changed and why. Name the note or artifact
    that established the correction.
 
+   A second revision on a date the note already carries writes the issue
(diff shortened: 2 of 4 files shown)


sayuks:
## Review record

(`9bc0ae2`), read at `c8c1df0`. Standards and Spec ran as parallel sub-agents.

**Spec** raised nothing on the acceptance criteria: every checkbox in #446's
Agent Brief was found satisfied, no *Out of scope* item was reached, and no
requirement was found implemented wrongly. Its two remarks are dispositioned
below with Standards'.

| finding | disposition |
|---|---|
| The inline comment re-derives the ratio's `* 1L` argument, and its "the only thing binding the source's type" is false on the local and dtplyr paths | fixed in code, `7d224a3` |
| The revisions section states present state — a bare `now`, "it is the form taken", and a present-tense heading | fixed in prose, `7d224a3` |
| `(#446)` in the `Revised:` line and the section heading is a form `investigation/README.md` does not fix | fixed in prose — the clause is added to `investigation/README.md` rather than excused in the note, `7d224a3` |
| Nothing reaches the no-join branch on dtplyr, so data.table's translation of the new expression is unasserted | fixed in code — `dtplyr Total shares needing no join match local results`, `7d224a3` |
| Without duckdb the local character-source refusal is unasserted | rejected — `Parent-share sources are numeric scalar summaries` asserts it unguarded, over `paste(value, collapse = "")` (`tests/testthat/test-share.R`) |
| The new DuckDB test duplicates its neighbour's fixture and loop | rejected — they cover different branches of `apply_joined_shares()`, and a shared fixture would make either one's columns a constraint on the other |
| `expect_type()` lacks the `info = column` its neighbours carry | rejected — `expect_type()` takes no `info`; the `expect_identical()` beside it carries the label |

After the answers: full suite `FAIL 0 | WARN 0 | SKIP 0 | PASS 6842` with
`NOT_CRAN=true`, `lintr::lint_package()` clean, and `verify-context-budget.R`,
`verify-doc-references.R`, `verify-verifier-invocation.R`, and
`verify-suite-coverage.R` all green locally.

sayuks:
## Review record — round 2

Same fixed point (`9bc0ae2`), read at `7d224a3`, so this round read round 1's
answers.

**Spec** raised nothing: all nine acceptance criteria found satisfied — the
reviewer re-rendered a rollup Total share on DuckDB and confirmed the join arm
carries no extra term — and neither of round 1's two additions is on the *Out
of scope* list.

**Standards** raised four, all about prose.

| finding | disposition |
|---|---|
| `Both arms are the constant this branch has always produced` is a claim about the branch's history, which *Code comments* keeps out of a comment | fixed in code, `54b8243` |
| `the errors below are read for their class alone` is false of two of the four the loop raises — the local refusals are matched on their message | fixed in code, `54b8243` |
| The clause added to `investigation/README.md` introduces a `<target>` placeholder the file does not define; step 1 names `investigation/<successor>.md` | fixed in prose, `54b8243` |
| `expect_type()` beside an `expect_identical()` against a double vector asserts nothing further | rejected — it does not, and the pair is the shape `DuckDB shares a source at its declared type's maximum` already uses for the same reading, in this file (`tests/testthat/test-share-backends.R:1622`) |

After the answers: full suite `FAIL 0 | WARN 0 | SKIP 0 | PASS 6842` with
`NOT_CRAN=true`, `lintr::lint_package()` clean, `verify-doc-references.R` and
`verify-context-budget.R` green.

sayuks:
## Review record — round 3

Same fixed point (`9bc0ae2`), read at `54b8243`.

**Spec** raised nothing, and this round ran rather than read. The joined path's
rendered DuckDB SQL is byte-identical to `main`'s across five plans (Parent
`drop`/`keep`, Total-joined `drop`/`keep`, `grouping_sets(group) + ()`; 877
lines, same MD5). Every eligible source measured gives `1` as a double on
local, dtplyr, and DuckDB: integer, double, all-`NA`, zero, zero-row (with and
without `.by`), integer64, and each of `DECIMAL(38,0)`, `DECIMAL(38,10)`,
`DECIMAL(18,2)`, `BIGINT`, `UBIGINT`, `HUGEINT`, `UHUGEINT`, `FLOAT`,
`TINYINT`, `UTINYINT` at its declared maximum — so `* 1L` does not overflow
where this note's earlier revision found `* 1.0` did. Six dbplyr simulators
render the expression without a verb error; Arrow refuses a Total share before
reaching the branch, unchanged.

**Standards** raised three.

| finding | disposition |
|---|---|
| The dtplyr test's `on every backend` is false — Arrow never reaches the branch | fixed in code, `ba733c3` |
| `#446 put an expression in the branch` narrates a past change, which *Code comments* keeps out of a comment | fixed in code, `ba733c3` |
| *Supersession* step 1 templates its target as `investigation/<successor>.md` while step 2 says "note or artifact" | fixed in prose — step 1 now says what step 2 says, `ba733c3` |

**Observation, ticketed.** Spec reported a DuckDB `BOOLEAN` source going from
`1` on `main` to a Binder Error on `HEAD`. A test for it was written and
reverted: R's `max()` coerces a logical to integer, so the local summary is
eligible and there is no pair of backends to assert against each other. The
**joined** path splits the same way on `main`, so this branch makes DuckDB's
two paths agree rather than opening the divergence. #446's Acceptance does not
reach it and the fix is an ADR 0010 reading, so it is #449.

After the answers: full suite `FAIL 0 | WARN 0 | SKIP 0 | PASS 6842` with
`NOT_CRAN=true`, `lintr::lint_package()` clean, `verify-doc-references.R`
green, and `verify-suite-coverage.R` verifying all 779 tests reach one of the
five backends.

PR #450, opened 2026-09-05
State that the eligible-type rule reads each backend's own summary

### LINKED ISSUE #449: A logical share source is eligible locally and refused on DuckDB
> *This was generated by AI during the third review round of #448.*

## Summary

`max(<logical>)` is an eligible share source on the local path and refused on
DuckDB. R's `max()` coerces a logical to integer, so the local summary is an
integer and the eligible-type rule accepts it; DuckDB's `MAX` returns a
`BOOLEAN`, which the staged share's `* 1L` refuses at binding.

`?share_of_parent` lists Logical as ineligible, so the *rule* agrees with
DuckDB. What differs is which type each backend's `max()` produces, which the
rule is then applied to.

## Reproduction

```r

df <- data.frame(region = c("E", "E", "W"), flag = c(TRUE, TRUE, FALSE))
tb <- copy_to(con, df, "b")

f <- function(src) {
  summarize_with_margins(src, total = max(flag), whole = share_of_total(total))
}
f(df)
#>   total whole
#> 1     1     1
f(tb) |> collect()
#> Error: Binder Error: No function matches the given name and argument types
#> '*(BOOLEAN, INTEGER_LITERAL)'
```

`typeof(summarize_with_margins(df, total = max(flag))$total)` is `"integer"`,
which is why the local eligible-type check passes.

## Not introduced by #446

Measured 2026-09-05 at `54b8243` against duckdb 1.5.5, dbplyr 2.6.0, dplyr
1.2.1. The **joined** path shows the same split on `main`: a `rollup()` Parent
share over `max(flag)` computes locally and raises the same Binder Error on
DuckDB. #446 extended the binding to the branch that builds no ratio, so the
two paths on DuckDB now agree with each other; this issue is the divergence
that was already there on the joined one.

It was found while reviewing #448 and left out of that branch: #446's
Acceptance does not reach it, and the fix is not obviously in `R/share.R` —
the question is whether the eligible-type rule is about the summary's local
type or about what each backend's aggregate returns, which is an ADR 0010
reading.

## What is unmeasured

- Whether any aggregate other than `max()`/`min()` reaches the same split.
- PostgreSQL, and whether a converting dialect answers it at all.
- dtplyr, whose `max()` is R's.

sayuks:
Closes #449.

`max()` over a logical column is an eligible share source locally and refused
on DuckDB: R's `max()` coerces the logical to an integer the eligible-type rule
accepts, while DuckDB's `MAX` returns a `BOOLEAN` that the staged share's
`* 1L` refuses at binding. Both backends applied the rule ADR 0010 states; what
differed is the summary each applied it to. No behavioral change is made — the
reference is what was missing.

## What changed

- `?share_of_parent`'s *Lazy execution boundaries* section now states that the
  eligible-type rule reads the type the backend itself computed, not the type
  the same expression would produce in R, with `max()` over a logical column as
  the aggregate whose result type differs. The backend table's four rows are
  unchanged: the addition is about the input to the rule.
- A DuckDB test pins the split so a change on either side is visible. The local
  call succeeds with an integer summary and a share of `1`; the same call on
  DuckDB builds a `tbl_lazy` that fails at `collect()` with an error that is not
  a `marginplyr_error`. It collects an eligible source first, as the
  neighbouring character-source tests do, so a dropped table cannot satisfy an
  assertion that reads the error for its class alone.

The database-backends vignette was left alone. Its staged-query section already
carries the dialect-verdict argument, and stating the split there too would be a
second copy of what the reference topic now owns.

## Checks

- `testthat::test_local()` with `NOT_CRAN` and every optional Suggest required:
  0 failures, 0 skips.
- `roxygen2::roxygenise()`: `man/share_of_parent.Rd` is regenerated and
  committed.
- `spelling::spell_check_package()`: clean.
- The site was not rendered locally. No page, section heading, or
  `verify-site.R` marker is touched by this diff; `altdoc.yaml` covers it.

### DIFF
--- a/R/share.R
+++ b/R/share.R
@@ -386,12 +386,12 @@
 #' from a source it has shown to be ineligible. None of it reads a row of your
 #' data.
 #'
-#' That rule reads the type the backend itself computed, not the type the same
-#' expression would produce in R, so an aggregate whose result type differs
-#' across backends is eligible on some and refused on others. `max()` over a
-#' logical column is one: R coerces it to an integer, which the rule accepts,
-#' while a database whose `MAX` returns a boolean has produced an ineligible
-#' summary and refuses the share as its own error at [dplyr::collect()].
+#' The summary that rule judges is the one the backend computed, not the one
+#' the same expression would produce in R, so an aggregate whose result type
+#' differs across backends is eligible on some and refused on others. `max()`
+#' over a logical column is one: R coerces it to an integer, which the rule
+#' accepts, while a database whose `MAX` returns a boolean has produced an
+#' ineligible summary and refuses the share.
 #'
 #' What differs is what establishes the rule, and whether the
 #' exactly-one-value cardinality rule is established with it:
--- a/R/share.R
+++ b/R/share.R
@@ -384,7 +384,16 @@
 #' reported locally, before execution, on every backend. The eligible-type
 #' rule is also enforced on every backend, and no backend calculates a share
 #' from a source it has shown to be ineligible. None of it reads a row of your
-#' data. What differs is what establishes the rule, and whether the
+#' data.
+#'
+#' That rule reads the type the backend itself computed, not the type the same
+#' expression would produce in R, so an aggregate whose result type differs
+#' across backends is eligible on some and refused on others. `max()` over a
+#' logical column is one: R coerces it to an integer, which the rule accepts,
+#' while a database whose `MAX` returns a boolean has produced an ineligible
+#' summary and refuses the share as its own error at [dplyr::collect()].
+#'
+#' What differs is what establishes the rule, and whether the
 #' exactly-one-value cardinality rule is established with it:
 #'
 #' | Backend | What establishes the source rules |
--- a/man/share_of_parent.Rd
+++ b/man/share_of_parent.Rd
@@ -404,12 +404,12 @@ rule is also enforced on every backend, and no backend calculates a share
 from a source it has shown to be ineligible. None of it reads a row of your
 data.
 
-That rule reads the type the backend itself computed, not the type the same
-expression would produce in R, so an aggregate whose result type differs
-across backends is eligible on some and refused on others. \code{max()} over a
-logical column is one: R coerces it to an integer, which the rule accepts,
-while a database whose \code{MAX} returns a boolean has produced an ineligible
-summary and refuses the share as its own error at \code{\link[dplyr:collect]{dplyr::collect()}}.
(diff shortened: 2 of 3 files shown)


sayuks:
## Review record

Two rounds through `mattpocock-skills:code-review` against the merge-base with
`main` (`git diff main...HEAD`, base `15a7725`).

### Round 1 — Standards (4) and Spec (2), answered in 7f24d4a

| Finding | Disposition |
|---|---|
| The new paragraph said the rule "reads the type the backend itself computed", which the same section contradicts thirty lines below: on a general dbplyr backend the rule is the dialect's to apply rather than marginplyr's to read | Fixed in prose — it now says the *summary* that rule judges is the one the backend computed |
| Its closing clause restated the `collect()` mechanics the paragraph below the backend table already owns | Fixed in prose — the clause is deleted |
| The test header restated the roxygen's local half, and the probe comment restated the reason it cited | Fixed — both cut to what only the site knows |
| Third copy of the DuckDB scaffold across three adjacent tests (judgement, Duplicated Code) | Rejected. The three tests pin different subjects — a character source through the joined ratio (#429), one through the branch that builds no ratio (#446), and this backend split (#449) — and one merged test would not say which failed |
| Spec: the DuckDB half was pinned only as an error that is not a `marginplyr_error`, so it would stay green if `max(flag)` itself stopped computing on DuckDB — the excluded `mean(<logical>)` shape | Fixed in code — the test now collects `summarise(remote, m = max(flag))` first, pinning that the summary succeeds and only the share on it is refused |
| Spec: AC5 (`verify-site.R`) not executed locally | Recorded. No page, heading, or hand-written marker quotes the sentence this diff split; `altdoc.yaml` is the gate |

Nothing out of scope: `is_share_source_type()`, the staged `* 1L`, the backend
table's four rows, and ADR 0010 are untouched.

### Round 2 — clean

Ran over the executable behaviour round 1 did not see. It measured the
exclusion the new assertion's comment claims: `MAX(flag)` computes on DuckDB
and returns `TRUE`, while `mean(flag)` fails at bind from plain dbplyr with no
marginplyr involved. No findings survived.

### Gates

`testthat::test_local()` with `NOT_CRAN` and every optional Suggest required —
6855 pass, 0 fail, 0 skip. `lintr::lint_package()` after `pkgload::load_all(".")`
— no lints. `roxygen2::roxygenise()` — no further diff.
`spelling::spell_check_package()` — clean. `verify-suite-coverage.R` and
`verify-must-error.R` — exit 0.

PR #469, opened 2026-09-06
Refuse a dtplyr step a backend would write to

### LINKED ISSUE #451: expand_with_margins() and nest_with_margins() corrupt a dtplyr input built with immutable = FALSE and return a wrong result
## Observed

With a `dtplyr` step built from `lazy_dt(dt, immutable = FALSE)`, the union adapter's branches mutate the same underlying `data.table` by reference. Two consequences, both silent apart from a `data.table` shallow-copy warning that leaks to the caller:

1. The result is wrong: every row of `expand_with_margins(ldt, .grouping = rollup(region))` carries `region = "Total"`.
2. The caller's source table is corrupted: its column *names* are permuted while the column *data* is not, so after the call `dtx$region` holds the integers that were `dtx$year`.

`nest_with_margins()` shows the same corruption. `summarize_with_margins()` leaves the source intact.

```r
dtx <- data.table::as.data.table(df)          # df: year, region, store, ...
ldt <- dtplyr::lazy_dt(dtx, immutable = FALSE)
r <- dplyr::collect(expand_with_margins(ldt, .grouping = rollup(region)))
table(r$region, useNA = "always")
#> Total  <NA>
#>    48     0
head(dtx, 2)
#>    region   year  store ...
#>     <int> <char> <char>
#> 1:   2023   East      a
```

The generated query is `funion(_DT[, := (region = as.character(region))], _DT[, := (region = rep("Total", .N))], all = TRUE)`, so the second branch overwrites the first before the union runs.

## Attribution

dtplyr's own `union_all(mutate(ldt, region = "T"), ldt)` also returns a wrong result under `immutable = FALSE`, so multi-branch queries over a mutable step are dtplyr's hazard. What marginplyr adds is that it builds several branches from one step on the caller's behalf, corrupts the caller's object, and says nothing.

## Proposed remedy

A step built with `immutable = FALSE` is recognisable: `step$implicit_copy` is `TRUE` there and `FALSE` under the default. Refuse such an input with a Package condition naming `dtplyr::lazy_dt(immutable = TRUE)` as the rewrite, in the verbs that build more than one branch, or document the hazard where the dtplyr backend is described. Nothing in `R/`, `man/`, or `vignettes/` mentions `immutable` today.

sayuks:
Fixes #451.

A dtplyr step whose root was built with `dtplyr::lazy_dt(immutable = FALSE)` — a *Mutable step* — is refused by every verb taking `.grouping`, before any branch is built.

A margin verb builds one branch per grouping set from the one step it was handed, and data.table writes each branch to the caller's own table by reference. Over such an input `expand_with_margins()` returned a result in which every row carried the margin label, and left the caller holding a table whose column *names* had been permuted while its column *data* had not; a `select()` in the input's own pipeline lost the caller a column outright.

```
Error in `expand_with_margins()`:
! `.data` comes from `dtplyr::lazy_dt(immutable = FALSE)`.
i A margin verb builds one branch per grouping set from the same step, and data.table writes each branch to your table by reference.
i Rebuild the input with `dtplyr::lazy_dt(immutable = TRUE)`.
```

## What is here

- The refusal, on `prepare_grouping_plan()` immediately after `grouping_backend()` resolves the kind — the one preparation path every verb taking `.grouping` reaches, and a point that has read no rows (ADR 0005). The predicate walks `$parent` to the root and reads `implicit_copy` there.
- **Mutable step** in `CONTEXT.md`, the analogue of *Absorbing backend*.
- ADR 0029, recording why the input is refused rather than silently copied or merely documented, with ADR 0025 as the precedent and what differs: 0025 refuses because a backend would read more than the caller needs, this refuses because a backend would destroy what the caller holds.
- A `must_error: marginplyr_error` chunk in the database-backends vignette, with the matching marker in `verify-site.R`.
- Tests in `test-grouping-backends.R`, guarded by `skip_if_suggest_absent("dtplyr")` and requiring dtplyr alone.
- A `NEWS.md` entry.

## The line, and what it costs

The predicate reads the root and nothing below it. `lazy_dt(d, immutable = FALSE) |> filter(...)` returns a correct result today and is refused anyway: `filter()` and `select()` produce the same step class carrying the same `implicit_copy` value and fall on opposite sides, so nothing below the root separates the destructive derivations from the safe ones. ADR 0029 records that over-refusal and the alternatives discarded for it.

A field the walk cannot find lets the input through rather than refusing on a name that no longer means what it did. Both fields are pinned in both directions and on a derived step, so a dtplyr that renames either fails CI instead of silently switching the refusal off.

A bare `data.table` is unaffected (it resolves to the `local` kind), as is every other backend.

## Verification

- `lintr::lint_package()` — no lints
- full test suite under `NOT_CRAN=true` — 0 failures, 0 skips
- `verify-must-error.R`, `verify-context-budget.R`, `verify-doc-references.R`, `verify-verifier-invocation.R` — pass
- the vignette rendered standalone against the installed working tree: the `must_error` chunk raises, and the new site marker matches the rendered page

### DIFF
--- a/.github/scripts/verify-site.R
+++ b/.github/scripts/verify-site.R
@@ -372,6 +372,12 @@ markers <- list(
     "DuckDB covers native SQL end to end",
     "dozen-plus fallback dialects",
     "Arrow and dtplyr are tested for the lazy",
+    # The unconditional half of the mutable-step refusal (ADR 0029). Its
+    # `must_error` chunk is behind `has_dtplyr`, so the prose introducing it is
+    # what a marker can reach, and this run of it holds no apostrophe and no
+    # double hyphen — pandoc's `smart` extension rewrites both in prose, where
+    # only a code span keeps them literal.
+    "builds one branch per grouping set from the step it is given",
     # The unconditional half of the absorbed-summary section. Its `must_error`
     # chunk is behind `has_arrow` like the two share refusals below it, so the
     # prose introducing it is what a marker can reach. Chosen from a run of
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -161,6 +161,22 @@ absorbing reads every column of the input while a caller who is told can read
 fewer.
 _Avoid_: Pulling backend, fallback backend, collecting backend
 
+**Mutable step**:
+A dtplyr step whose root was built with `dtplyr::lazy_dt(immutable = FALSE)`,
+which is the caller giving data.table permission to write to their own table.
+It is a property of how the input was constructed rather than of its class,
+and it is therefore finer-grained than a backend kind, one kind holding both,
+which is why it is established by walking the step's parents to its root
+rather than looked up. Nothing below the root separates the derivations that
+destroy the caller's table from the ones that survive: a `filter()` and a
+`select()` over the same mutable root produce the same step class carrying the
+same field value and fall on opposite sides. A Margin verb refuses one before
+any branch is built, because it builds one branch per grouping set from the
+same step and data.table writes each of them to the caller's table by
+reference — leaving a result in which every row carries the margin label, and
+a table whose columns were dropped or whose names were permuted.
+_Avoid_: Mutable input, in-place backend
+
 **Sent query**:
 One SQL statement marginplyr gives to a backend while compiling a Grouping
 plan or while executing the Margin operation built on one, recorded with the
--- a/NEWS.md
+++ b/NEWS.md
@@ -206,6 +206,18 @@
   returned the right answer, having read the whole input to get it — so on
   those versions this is a breaking change, and the rewrites the refusal names
   reproduce the old result while letting you choose what is read.
+* A dtplyr input built with `dtplyr::lazy_dt(immutable = FALSE)` is now
+  refused, before any branch is built. A margin verb builds one branch per
+  grouping set from the step you gave it, and a mutable step lets data.table
+  write every one of those branches back to your own table by reference: the
+  call returned a result whose every row carried the margin label, and left
+  your table with a column dropped or its names permuted while its data stayed
+  put. The refusal names the one rewrite, `dtplyr::lazy_dt(immutable = TRUE)`,
+  which is what `lazy_dt()` does by default. It is deliberately wider than the
+  damage — a mutable step you only filtered comes back correct today and is
+  refused with the rest — because which derivations survive is decided by the
+  query dtplyr generates rather than by anything marginplyr can promise. A
+  `data.table` passed directly is unaffected, and so is every other backend.
 * A `.grouping` or `.by` selection whose failure the column names already
   settle is now refused without querying your input. This reaches the
   set-difference operator `/` that tidyselect reads, the arithmetic and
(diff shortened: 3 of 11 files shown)


sayuks:
## Review round 1 — `mattpocock-skills:code-review`

Fixed point: merge-base with `main` (`4c404a5`), `git diff main...HEAD`. Reviewed at `8d39865`; answered in `ea03214`.

### Standards — 7 findings

| finding | disposition |
|---|---|
| ADR 0029's *Documentation consequences* says the site marker is quoted from the refusal's own prose; the marker added is the vignette's, and the ADR had already drifted from `verify-site.R`'s comment on the same entry | fixed in prose — the chunk is behind `has_dtplyr`, so the marker cannot come from the diagnostic; the ADR now says that |
| `abort_mutable_dtplyr_step()`'s header re-derives the over-refusal argument beside its ADR 0029 citation — a second copy `AGENTS.md` *Code comments* refuses | fixed in code |
| `prepare_grouping_plan()`'s inline comment reproduces ADR 0029's *Where it is raised* and cites 0005 but not 0029 | fixed in code — cut to the ordering constraint the site knows, which that rule admits, and the ADR cited for the placement |
| "`filter()` and `select()` both produce a `dtplyr_step_subset`" is false — a `select()` naming every column returns a `dtplyr_step_first` (verified on dtplyr 1.3.3) | fixed in code and prose — the class claim is deleted rather than restated; both sites now state the field value, which is what the predicate reads |
| `data.table::` called under a dtplyr-only guard, with the Import-closure reason nowhere recorded | fixed in code and prose — and a guard on data.table would be wrong, not merely redundant: `verify-suite-coverage.R` hides every backend but the selected one, so it would skip this section in the only configuration that executes it |
| the six verb wrappers duplicate `forwarded_verbs` in `test-grouping-plan.R` | fixed in code — that list moves to `helper-margin-verbs.R`, by the argument that file's header already makes for `verbs_taking()` |
| reference pages still name dtplyr support unqualified (`?nest_with_margins`, `R/share.R:377`, `R/nest_by_with_margins.R:19`, `R/grouping-context.R:78`); ADR 0025, the cited precedent, updated its claiming pages | ticket — #470. A finding this branch's Acceptance does not reach, per *Answering a review* |

Worst: the false ADR sentence, which had already diverged from the file it described.

### Spec — 4 findings, none a defect

| finding | disposition |
|---|---|
| the criterion asks every verb's refusal be "pinned by snapshot"; one was, and the header — the only part that differs per verb — was unpinned | fixed in code — all six pinned |
| "returns the same result it does today" asserted against a written-out expectation rather than a baseline | fixed in code — now the same call over a local frame |
| a grouped mutable step now reaches this refusal rather than the fixed-key rejection its groups used to earn, falsifying the triage note's "a grouped input errors earlier"; nothing pinned it | fixed in code and prose — pinned, and recorded in the ADR as a consequence of the mandated placement |
| `$implicit_copy` partial-matches on a list, so a dtplyr rename keeping the old name as a prefix would be read rather than let through — weaker than the ADR's stated behaviour | fixed in code — both fields read with `[[`, and the pinning test spells such a rename out |

Nothing out of scope was found: no `local`/arrow/duckdb/postgres/sql path is touched, no copy-and-continue, no per-step predicate.

### Verification after the answers

`lintr::lint_package()` clean; full suite under `NOT_CRAN=true` 0 failures and 0 skips; `verify-doc-references.R` passes.

PR #471, opened 2026-09-06
Refuse a selection predicate with the context a verb owes

### LINKED ISSUE #453: where() on a backend with no readable schema is refused with a bare tidyselect diagnostic
## Observed

On a backend whose kind lacks `can_read_schema` (SQLite, Postgres — kinds `sql` and `postgres`), a `where()` predicate in a summary selection or in `.grouping` is refused with tidyselect's message and nothing else:

```r
summarize_with_margins(pg, across(where(is.numeric), sum), .grouping = rollup(region))
#> Error: This tidyselect interface doesn't support predicates.
summarize_with_margins(pg, n = n(), .grouping = rollup(where(is.character)))
#> Error: This tidyselect interface doesn't support predicates.
```

The same calls succeed on local data, dtplyr, DuckDB, and Arrow, because there the selection proxy carries column types.

## Why it is a finding

The refusal itself matches dbplyr: `summarize(pg, across(where(is.numeric), sum), .by = region)` is refused too. What differs is the Condition context. dplyr's own diagnostic reads

```
ℹ In argument: `across(where(is.numeric), sum)`
Caused by error in `across()`:
! This tidyselect interface doesn't support predicates.
```

while marginplyr's names neither the argument nor the verb, which is the context CONTEXT.md's *Condition context* entry says a Margin verb owes. The reason for the refusal — that this backend's proxy is the lazy table itself, per `grouping_selection_proxy()` in `R/backend-metadata.R`, because reading a schema would send a query (ADR 0020) — is also stated nowhere a caller reads.

## Proposed remedy

Catch the tidyselect refusal at `resolve_summary_selection()` and at the `.grouping` selection site, and re-raise it with the argument as the caller spelled it, the Margin verb as the blamed call, and a line saying that a predicate needs column types this backend does not expose without a query. Document the limit in *Relationship to dplyr summaries*.

sayuks:
Closes #453.

A backend whose selection proxy is the lazy table itself reports no column types, so tidyselect refuses a `where()` predicate with its own diagnostic and nothing else. The refusal stands — reading the types would be a query nobody asked for (ADR 0020) — but it named neither the argument the caller wrote nor the verb they wrote it in, which is the Condition context `CONTEXT.md` says a Margin verb owes.

`abort_selection_predicate()` re-raises it as a Package condition, carrying tidyselect's own condition as its cause, at the three selections resolved against the input's proxy: a summary selection, a grouping dimension, and a fixed `.by` key. `.by` is the one the ticket did not name; it reaches `resolve_column_selection()` by the same route and was refused the same way.

The class `tidyselect_error_predicates_unsupported` is the whole of the test, and nothing reads the backend: a proxy carrying types answers a predicate rather than refusing one.

*This PR body was generated by AI.*

### DIFF
--- a/R/conditions.R
+++ b/R/conditions.R
@@ -73,6 +73,50 @@ condition_chain <- function(cnd) {
   c(list(cnd), condition_chain(cnd$parent))
 }
 
+# Whether a failure is tidyselect refusing a predicate for want of column
+# types. The whole chain is read for the reason `condition_chain()` above
+# gives: a predicate written under a selection helper is refused one level in.
+#
+# The class is tidyselect's own and is the whole of the test. A backend whose
+# selection proxy carries types answers a predicate rather than refusing one,
+# so no reading of the backend is needed to tell the two cases apart, and a
+# tidyselect that stopped raising this class leaves the caller the untyped
+# diagnostic they had before this refusal existed.
+is_unsupported_predicate <- function(cnd) {
+  any(vapply(
+    condition_chain(cnd),
+    inherits,
+    logical(1),
+    what = "tidyselect_error_predicates_unsupported"
+  ))
+}
+
+# The refusal a selection predicate gets where the proxy has no types. `label`
+# is the argument as the caller spelled it, and `parent` is tidyselect's own
+# condition, kept so the caller reads why tidyselect refused.
+#
+# The blamed call is left to `with_margin_error_call()`, which every selection
+# site reaching this runs under: what it puts there is the Margin verb the
+# caller wrote, which is what CONTEXT.md's *Condition context* asks for and
+# what no frame here could name.
+#
+# Reading the types would mean a query the caller did not ask for, which
+# ADR 0020 refuses; that is why the proxy is typeless here rather than an
+# omission this could fix.
+abort_selection_predicate <- function(label, parent) {
+  abort_marginplyr(
+    c(
+      "Can't select with a predicate in {.code {label}}.",
+      i = paste0(
+        "This input's backend doesn't report column types without a query, ",
+        "and marginplyr sends none you didn't ask for."
+      ),
+      i = "Select the columns by name, or collect the input first."
+    ),
+    parent = parent
+  )
+}
+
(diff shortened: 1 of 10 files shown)


sayuks:
## Review round

`mattpocock-skills:code-review` against `ce1915c...HEAD` (the merge-base carries two commits from another ticket, reviewed in their own round). Standards and Spec axes.

Reviewed at 3bed65a; answered in eba1db8.

**Standards — 4 findings.**

- The new documentation paragraph sat in *Relationship to dplyr summaries*, which no other verb inherits, while the `.by` and `.grouping` refusals are raised in shared code every Margin verb reaches. **Fixed in prose**: the paragraph moved to *Fixed columns and grouping dimensions*, inherited by all four verbs; the summary section keeps a pointer. `man/expand_with_margins.Rd`, `man/nest_with_margins.Rd`, and `man/nest_by_with_margins.Rd` now carry it.
- `abort_selection_predicate()`'s header re-derived ADR 0020's argument beside the citation, which *Code comments* refuses. **Fixed in prose.**
- The two test files stated the same rationale. **Fixed in prose**: the grouping test cites the summary test.
- `is_unsupported_predicate()` reads as a predicate that is unsupported; `is_unsupported_selection_predicate()` would reveal it. **Rejected**: `object_length_linter()` fails that name at 34 characters, and `lint.yaml` is what records it.

One smell, judged and left: the catch shape appears three times. **Left alone** — the third site is folded into `resolve_grouping_selection()`'s existing handler, where it has to run after that function's own predicate reading, so a wrapper would leave two sites and a special case rather than one shape.

**Spec — 1 finding, no defect.**

- Scope: `.by` is a third site the ticket named two of. **Left as built** — it reaches `resolve_column_selection()` by the same route and was refused the same way, and the documentation would have been false for it otherwise. The commit message says so.
- The brief's `skip_if_suggest_absent()` criterion is satisfied vacuously: both tests use `dbplyr::tbl_lazy()`, and dbplyr is an Import, so there is no guard to register and no member of `optional_backends()` involved.

> *This was generated by AI during code review.*

PR #472, opened 2026-09-06
Force the input before emptying the sent-queries record

### LINKED ISSUE #455: last_sent_queries() spans two calls when Margin verbs are piped
## What happens

With `options(marginplyr.audit_sql = TRUE)`, piping one Margin verb into another leaves the record holding both calls' rows:

```r
q <- db |>
  summarize_with_margins(revenue = sum(revenue, na.rm = TRUE), .grouping = rollup(region, store)) |>
  summarize_with_margins(revenue = sum(revenue, na.rm = TRUE), .grouping = rollup(region))
last_sent_queries()$purpose
#> [1] "selection_proxy" "result" "selection_proxy" "result"
```

Measured on DuckDB. The same happens piping into `inspect_grouping()` (`selection_proxy, result, selection_proxy`) and with a dplyr verb between the two (`... |> dplyr::filter(...) |> expand_with_margins(...)`).

## Why

`|>` expands to `g(f(x))`. `g` runs `reset_sent_queries()` as its first statement, then `assert_margin_input(.data)` forces the `.data` promise, which runs `f`: `f` resets again, records its own rows, and `g` appends to them. `"result"`, the one promised purpose, then appears twice, and nothing says the record spans two calls. That is the `dbplyr::last_sql()` defect ADR 0027 states the feature exists to remove, reproduced by the idiomatic way of writing the call.

## What is involved

- `R/sent-queries.R`: `reset_sent_queries()` and its header, which fix the reset at the entry point's first statement.
- The five entry points (`summarize_with_margins()`, `expand_with_margins()`, `nest_with_margins()`, `nest_by_with_margins()`, `inspect_grouping()`), each opening with `reset_sent_queries()`.
- `tests/testthat/test-sent-queries.R`: the gate asserting that every exported function taking `.grouping` calls `reset_sent_queries()` *first*. It currently fixes the defective ordering.
- ADR 0027, *Amendment: the reset is the first statement of each entry point*, which records the placement and the "call began" reading.

A fix along the lines of forcing `.data` before the reset (`force(.data); reset_sent_queries()`) keeps the "record belongs to the call whose body ran" reading: an inner verb that refuses leaves its own reset record, and the outer never resets. The gate and the ADR amendment move in the same commit.

sayuks:
Closes #455.

Every entry point emptied the sent-queries record as its first statement, but
`.data` is a promise that the argument validation each of them opens with is
the first to force. `|>` expands to `g(f(x))`, so `g` reset, then forced
`.data`, which ran `f`; `f` reset again, recorded its own rows, and `g`
appended to them. The record held both calls, with `"result"` — the one
promised `purpose` — in it twice and nothing saying it spanned two. That is the
`dbplyr::last_sql()` defect ADR 0027 exists to remove, reached by the idiomatic
spelling.

Each entry point now opens with `force(.data)` and empties the record second,
which puts the input's whole execution before this call's record exists.

Forcing ahead of the validation block is the same move as forcing ahead of the
`tryCatch()` that block runs in, so an input verb's `marginplyr_error` keeps
its own call instead of being reported against the verb reading it. The two are
not separable.

The gate in `test-sent-queries.R` fixed the defective ordering; it now holds
each entry point to the two statements in order, and its reader takes a
statement by position rather than only the first. ADR 0027 gains an amendment
for the placement, and the Rd, the vignette, and NEWS say which call a piped
record belongs to.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -276,6 +276,10 @@
   forwarding where that argument is not bound. The check carries no call
   there now, and a share's conditions still name the call you wrote,
   forwarding and all (#357).
+* A Margin verb piped into another one now names itself when it refuses.
+  `f(x) |> g()` runs `f` while `g` reads its input, and `g` reported every
+  marginplyr error raised there against its own call, so a refusal you would
+  fix by rewriting `f` was blamed on `g` (#455).
 * A condition raised while your summary expression runs now reports its
   context in names you can act on. A margin operation summarizes that
   expression once per grouping set, so the grouping values it reported were
@@ -344,5 +348,8 @@
   collision or by an ineligible share source leaves readable every query it
   had already sent. A call one of these verbs refused before sending anything
   reads back as the empty record it is, rather than as the previous call's,
-  and is a call the session has recorded. The option is off by default
-  (#318, #400, #401, #409).
+  and is a call the session has recorded. Piping one Margin verb into another
+  leaves the outer call's record and not both calls': the input runs before
+  the outer call empties one, so `"result"` is in the record once and is the
+  query you were handed. The option is off by default
+  (#318, #400, #401, #409, #455).
--- a/R/expand_with_margins.R
+++ b/R/expand_with_margins.R
@@ -92,6 +92,7 @@ expand_with_margins <- function(.data,
                                 .duplicates = c("error", "drop", "keep"),
                                 .id = NULL,
                                 .sort = c("none", "last", "first")) {
+  force(.data)
   reset_sent_queries()
   call <- rlang::current_call()
   with_margin_error_call(
--- a/R/inspect-grouping.R
+++ b/R/inspect-grouping.R
@@ -136,6 +136,7 @@ inspect_grouping <- function(.data,
                              .grouping = NULL,
                              .duplicates = c("error", "drop", "keep"),
                              .format = c("text", "list")) {
+  force(.data)
   reset_sent_queries()
   call <- rlang::current_call()
   grouping_quo <- rlang::enquo(.grouping)
--- a/R/nest_by_with_margins.R
+++ b/R/nest_by_with_margins.R
@@ -119,6 +119,7 @@ nest_by_with_margins <- function(.data,
                                  .sort = c("none", "last", "first"),
                                  .key = "data",
                                  .keep = FALSE) {
+  force(.data)
   reset_sent_queries()
   call <- rlang::current_call()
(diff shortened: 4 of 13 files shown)


sayuks:
## Review round: `mattpocock-skills:code-review`

Fixed point `main`, three-dot, read at `2390e9d`. Two axes, parallel sub-agents. Answers landed in `a2527de`.

### Standards — 3 findings

| finding | disposition |
|---|---|
| `reset_sent_queries()`'s header re-derived ADR 0027's amendment beside the citation — the second copy *Code comments* refuses | fixed in prose; the sentence is deleted, leaving the placement and the citation |
| The gate's second position was never read on a body whose first position passes, so a statement pushed between the forcing and the reset — a validation moved above the reset, under the new order — was untested; the `\|\|` short-circuited on `swapped` | fixed in code: `validates_between` in `test-sent-queries.R` |
| `nth_statement()` compared `n` with `identical(n, 1L)` on the unbraced branch and `n + 1L` on the braced one, so a double `1` answered the two differently | fixed in code |

### Spec — 3 findings, 1 rejected

Spec source: #455 and its agent brief.

| finding | disposition |
|---|---|
| The piped-result test carried no counterpart to its sibling's `expect_no_match(record$sql, "first")` | fixed in code, by a different assertion: the outer query nests its input's, so no substring is absent from one of them. The row is now compared against `dbplyr::sql_render()` of the query the call returned |
| The brief names three piped cases and the diff collapsed `expand_with_margins()` into the dplyr-between test | fixed in code: the bare `summarize \|> expand_with_margins()` case is its own test |
| The record half and the condition-call half are asserted in different files, never both of one scenario | rejected. The brief asks that both be asserted, not that one test hold both, and each is in the file whose subject it is: `test-sent-queries.R` owns the record, `test-grouping-interface.R` owns which call a condition names, where its siblings already are |

Both axes reported no scope creep against the brief's *Out of scope* list.

One observation outside the brief, raised as new: `force(.data)` puts an omitted input's error outside the block that rewrites a call. Measured — it was never caught there, which rewrites only `marginplyr_error`, and the message and `missingArgError` class are unchanged. What moved is R's own subclass, `getvarError` to `evalError`. `test-verb-argument-admission.R` now holds what a caller acts on, over every entry point.

Checks after the round: suite 0 failures / 0 skips / 0 warnings under `NOT_CRAN=true`, `lintr::lint_package()` clean, `verify-context-budget.R`, `verify-must-error.R`, and `verify-verifier-invocation.R` all pass.

PR #473, opened 2026-09-06
Give the live-SQLite verification scope one owner

### LINKED ISSUE #457: SQLite verification scope is stated three ways
## Problem

Three sources describe what is live-tested on SQLite, and they disagree:

- `README.Rmd:339`: "Live portable Parent-share execution tested, with `.check_share_source = FALSE`; the default refuses the share on these dialects".
- `vignettes/database_backends.qmd:672-674`: live claim is "portable `UNION ALL` execution for Parent shares", with no `.check_share_source` qualifier, although the vignette's own rule at 706-710 says the share is refused by default on SQLite.
- `R/summarize_with_margins.R:547-548` (*Database backend coverage*) states a third scope.

The tests also run live RSQLite for margin order (`test-margin-order.R`), the observed-collision scan (`test-margin-label.R`), flat `UNION ALL` rendering (`test-branch-union.R`), and the audit record (`test-sent-queries.R`), so every stated scope understates what is covered.

## Proposal

Make the reference section the single owner of the verification table. Reduce `database_backends.qmd:664-683` to the two-sentence caveat plus the link it already carries at 682, and keep the README table in step with the reference (or link instead of restating).

sayuks:
Closes #457.

Three documents stated what is verified against a live SQLite database and no two agreed:

- the README's *Backend verification* table — Parent-share execution, qualified with `.check_share_source = FALSE`;
- the database-backends vignette's *Backend verification levels* — Parent-share `UNION ALL` execution, unqualified, beside its own rule saying the share is refused by default there;
- *Database backend coverage* on `summarize_with_margins()` — contextual shares, end to end.

All three understated the suite. Live RSQLite connections are opened in five test files, covering Parent shares under `.check_share_source = FALSE`, the refusal the default raises on that dialect, portable Margin orders through `collect()` and `compute()` including missing-value placement SQLite's own ordering would not produce, the observed half of the Margin label collision check, and the queries `last_sent_queries()` records.

## What this does

The reference section is now the single owner of the scope. It also separates what is executed against a live SQLite connection from what is only rendered there — several tests `copy_to()` that database and then assert on `sql_render()` alone, so a statement conflating the two would have been the fourth wrong scope.

The vignette keeps the argument a vignette owes — simulator coverage is evidence about rendering and not about a running server — and links to the reference instead of restating a scope. The README carries a pointer in place of its table, and the earlier README cross-reference now names the owner rather than the guide that forwards to it.

`verify-site.R` pinned four sentences of the rewritten vignette section; its markers follow the new prose and name no backend, since a marker naming one would reintroduce what the section was reduced to remove.

## Checks

- `spelling::spell_check_package()` — clean
- full `testthat::test_local()` with every optional Suggest present — passing
- `roxygen2::roxygenise()` and `rmarkdown::render("README.Rmd")` re-run and committed (local pandoc 3.10.1 matches the `document.yaml` pin)
- `altdoc::render_docs(parallel = FALSE, freeze = FALSE)` then `verify-site.R` — passing
- `verify-must-error.R`, `verify-context-budget.R`, `verify-verifier-invocation.R` — passing

### DIFF
--- a/.github/scripts/verify-site.R
+++ b/.github/scripts/verify-site.R
@@ -368,10 +368,12 @@ markers <- list(
     # renders: every chunk in it is behind `has_duckdb`, so on a machine
     # without DuckDB the section produces no output for a marker to quote.
     "Record the SQL marginplyr sends",
-    "Only DuckDB and SQLite are exercised as live SQL databases",
-    "DuckDB covers native SQL end to end",
-    "dozen-plus fallback dialects",
-    "Arrow and dtplyr are tested for the lazy",
+    # The verification section, which states no scope of its own: the
+    # `summarize_with_margins()` reference owns that, so a marker naming a
+    # backend here would be the restatement the section was reduced to remove.
+    "SQL simulation is useful evidence about rendering",
+    "treat a simulator-only backend as SQL-generation support",
+    "Which claim rests on which is stated in one place",
     # The unconditional half of the mutable-step refusal (ADR 0029). Its
     # `must_error` chunk is behind `has_dtplyr`, so the prose introducing it is
     # what a marker can reach, and this run of it holds no apostrophe and no
--- a/.github/scripts/verify-site.R
+++ b/.github/scripts/verify-site.R
@@ -370,7 +370,7 @@ markers <- list(
     "Record the SQL marginplyr sends",
     # The verification section, which states no scope of its own: the
     # `summarize_with_margins()` reference owns that, so a marker naming a
-    # backend here would be the restatement the section was reduced to remove.
+    # backend here would pin a restatement this section does not carry.
     "SQL simulation is useful evidence about rendering",
     "treat a simulator-only backend as SQL-generation support",
     "Which claim rests on which is stated in one place",
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -556,14 +556,15 @@
 #' tests execute DuckDB queries against a live in-memory database and verify
 #' PostgreSQL SQL with dbplyr's simulator.
 #'
-#' SQLite is the other live database, and it runs the portable `UNION ALL`
-#' path end to end: Parent shares under `.check_share_source = FALSE`, the
-#' refusal the default raises there -- SQLite converts a value of another type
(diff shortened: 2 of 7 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, both axes, against `git diff main...HEAD` at `0b91828`. Answers are in `6b8fdbb`.

**Standards — four findings, all fixed in prose.**

1. *(hard, `design/architecture.md` *Answering a review* clause 2)* The README stated the simulator-versus-live argument a second time and carried the pointer to *Database backend coverage* twice, ~50 lines apart. Before the diff the second block held the table, so the duplication was newly introduced.
2. The coverage section's clause on why SQLite triggers the share refusal was a fourth copy of an argument `@param .check_share_source`, *Contextual shares*, and the diagnostic `_snaps/share-backends.md` pins already carry.
3. `expand_with_margins()` inherits the section and documents no share argument, so the bare `.check_share_source` resolved to nothing on that page.
4. The new `verify-site.R` comment was in the past tense, against *Code comments* in `AGENTS.md`.

**Spec — three findings, all fixed in prose; nothing missing.**

1. The section listed the refusal the default raises among the things the portable path runs end to end, where nothing of that path runs. This is the conflation the brief named, inverted.
2. Same site as Standards 2.
3. Same site as Standards 1.

Both axes confirmed every remaining claim in the section against the live-RSQLite tests, and the repo-wide scan found no surviving live-SQLite scope outside the reference.

**One deviation from the brief.** Its criterion *"The README's earlier cross-reference names the document that now owns the scope, not one that forwards to it"* is answered by deleting the forwarding claim rather than repointing it: repointing is what put the pointer in the README twice, and the section that now owns the pointer is the one a reader scans for.

**Verified after the answers:** `lintr::lint_package()`, `spelling::spell_check_package()`, `testthat::test_local()` with every optional Suggest required (6929 passed, 0 skipped), `roxygen2::roxygenise()`, `rmarkdown::render("README.Rmd")` under pandoc 3.10.1, and `altdoc::render_docs()` followed by `verify-site.R`. The site was rendered from the tree at `6b8fdbb` less the final README sentence rewording, which no `docs/index.html` marker quotes.

PR #474, opened 2026-09-06
Render the package page, and hold it in the derived set

### LINKED ISSUE #456: Package page is unreachable from the site, taking the error-class contract and audit option with it
## Problem

`R/marginplyr-package.R:163` carries `@keywords internal`, and `.github/scripts/verify-site.R` states that altdoc renders no page for an internal topic. That page is the only place that

- lists the five guides together (*Guides* section),
- documents the `"marginplyr_error"` condition class contract (*Errors and warnings*),
- explains `options(marginplyr.audit_sql = TRUE)` and how `last_sent_queries()` answers (*Recording the SQL marginplyr sends*).

`README.Rmd` mentions none of `marginplyr_error`, `audit_sql`, `inspect_grouping()`, or `grouping_spec()`. A website reader therefore has no way to find the error-class contract.

## Options

- Drop `@keywords internal` so the page renders, or
- move *Errors and warnings* to a topic that has a page (e.g. `summarize_with_margins`) and add the guide list to README.

`verify-site.R` derives the required pages from non-internal `man/*.Rd`, so whichever option is taken, that gate follows automatically.

sayuks:
Closes #456.

`@keywords internal` on the `"_PACKAGE"` topic kept altdoc from rendering any page for it — `.render_one_man()` skips an `.Rd` matching `\keyword{internal}` — and `verify-site.R` derives its required pages from the non-internal `man/*.Rd`, so nothing asked for one. That topic is the only place documenting the `"marginplyr_error"` class contract, the `marginplyr.audit_sql` option and `last_sent_queries()`, and the five guides as one list; the README names none of them. The site had no path to any of it.

Dropping the keyword is the whole of the fix in the positive direction. altdoc names the page after the `.Rd`, which is what `page_for()` already derives, so `docs/man/marginplyr-package.html` joins the derived set with no edit to the script.

It is not the whole of the fix in the other direction. Restoring the keyword would take the page back out of `reference_sources`, requiring no page and failing nothing — which is how the three sections left the site for as long as they did. Naming the page in `markers` closes that: a `markers` key naming a page the site does not produce stops the script. Verified by re-adding the keyword to the generated `.Rd`, which now fails with `` `markers` names pages the site does not produce ``.

## Verified

- `Rscript .github/scripts/verify-site.R` — passes; 21 pages from 5 vignettes and **11** help topics, up from 10, with `docs/man/marginplyr-package.html` carrying markers.
- The rendered page carries all three sections, and `docs/index.html` links it from the Reference sidebar.
- `lintr::lint_package()` — no lints. Full suite under `NOT_CRAN=true` — all pass.
- `verify-context-budget.R`, `verify-verifier-invocation.R` — pass.

### DIFF
--- a/.github/scripts/verify-site.R
+++ b/.github/scripts/verify-site.R
@@ -27,9 +27,10 @@ read_page <- function(path) {
 }
 
 # `\keyword{internal}` is the rule altdoc applies when deciding that a topic
-# documents internals and gets no page of its own. Stating the rule rather than
-# naming the topic it currently excludes keeps a second internal topic from
-# demanding a page that will never be rendered.
+# documents internals and gets no page of its own. No topic has carried it
+# since #456, so this filter drops nothing today; it stays written as the rule
+# so that a topic given the keyword later leaves the set rather than demanding
+# a page that will never be rendered.
 documents_internals <- function(path) {
   any(grepl(
     "\\keyword{internal}",
@@ -419,12 +420,10 @@ markers <- list(
     "cannot select any column named in the complete grouping plan",
     "cur_group_id"
   ),
-  # Naming this page here is what holds it in the derived set. `@keywords
-  # internal` on the topic takes it out of `reference_sources` above, which
-  # requires no page and so fails nothing -- how the three sections below left
-  # the site for as long as they did (#456). A `markers` key naming a page the
-  # site does not produce is an error, so restoring that keyword now stops this
-  # script rather than silently dropping the page again.
+  # `@keywords internal` on the `"_PACKAGE"` topic takes it out of
+  # `reference_sources` above, which requires no page and so fails nothing --
+  # how the three sections below left the site for as long as they did (#456).
+  # Naming the page here is what fails the run if the keyword returns.
   "docs/man/marginplyr-package.html" = c(
     # One marker per section that has no other page. The condition class and
     # the audit option are each quoted by the sentence stating what it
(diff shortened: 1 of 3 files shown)


sayuks:
## Review round — `mattpocock-skills:code-review`

Two axes against `origin/main...HEAD` at `ec19623`, standards sources `AGENTS.md` (*Site verification*, *Documentation*, *Code comments*) and `verify-site.R`'s own header. Four findings, all answered in `ed70423`.

|---|---|---|---|
| 1 | Standards + Spec | The Guides marker `"Complete absent keys before margins"` passes on a page whose Guides section was deleted. | Fixed in code |
| 2 | Standards + Spec | The comment beside it asserts the inverse of what the rendered page shows. | Fixed in prose |
| 3 | Spec | The entry's opening comment says naming a page in `markers` "holds it in the derived set", which the file header refuses. | Fixed in prose |
| 4 | Standards | `documents_internals()`'s header declines to name "the topic it currently excludes"; after this branch there is no such topic. | Fixed in prose |

Both axes raised finding 1 independently. Measured: all five Guides entries are vignette titles, and the sidebar renders every one on all 21 pages, so `"Complete absent keys before margins"` occurred at two offsets on the package page — 8992 in the sidebar and 24308 in the Guides list. The markers are now the heading and one entry's absolute href, which sit at 23726 and 24240, both inside `<section id="guides">`; the sidebar links the same vignette relatively, so neither run survives the section's deletion.

Re-verified after the fix: `verify-site.R` passes, and re-adding `\keyword{internal}` to the generated `.Rd` still fails it with `` `markers` names pages the site does not produce ``. `lintr::lint_package()` clean.

Not findings, recorded because both agents reached them: the README gaps #456 notes are *Out of scope* in the brief, and `AGENTS.md` needed no edit — its *Site verification* sentence about writing a marker from a rendered diagnostic conditions on `must_error`, which a man page has none of.

Issue #465, opened 2026-09-06
Documented features with no example in any guide


sayuks:
- **Bare `grouping_id()`** (`R/grouping-context.R:12-16`: reads every `.grouping` column in plan order and equals `inspect_grouping()$grouping_id`) appears in no vignette. *Keep only one grouping set* (`recipes.qmd:540`) warns about guessing the bit order and is exactly where this form helps.
- **`.sort` with `.id`**: `grouping_identity.qmd:117` and the reference say `.id` is the Margin order's final tiebreak, which keeps duplicate occurrences adjacent under `.duplicates = "keep"`, but no vignette shows them together. Get started's `.duplicates = "keep"` example (289-303) is the natural home.
- **`.sort` on `expand_with_margins()`** (`R/expand_with_margins.R:94`): every vignette mention of `.sort` is on `summarize_with_margins()`. `recipes.qmd:819-822` says the expansion result "is in no Margin order because the pipeline ends in an ordinary `summarize()`", leaving the reader unaware the verb itself can be ordered.
- **`.sort` on a lazy backend**: `database_backends.qmd:697-699` recommends it; nothing shows it.
- **`compute()`** is named at `database_backends.qmd:205-206` and 342 and never shown.
- **`nest_by_with_margins()`** is absent from the database vignette although it is the one Margin verb that collects a lazy input (`R/nest_by_with_margins.R:19,142`) and the one exemption *When marginplyr queries your data* names. *Other lazy backends* (472-484) covers only `nest_with_margins()`.
- **`last_sent_queries()`** has no `@family` (`R/sent-queries.R:132-135`) and is not mentioned from *When marginplyr queries your data* (`R/summarize_with_margins.R:557-611`), which is where a reader wonders how to see the queries.
- `database_backends.qmd:35`: the install list omits `data.table`; harmless since dtplyr imports it, but the guard at 63 checks only dtplyr, so say why that suffices.

Found in a documentation review on 2026-09-06.

sayuks:
> *This was generated by AI during triage.*

## Verification (against 671d8b1)

All eight items confirmed. Two notes.

- **Bare `grouping_id()`**: it is named in prose in four vignettes, but no
  chunk calls it with no arguments. Read the first bullet as "no worked
  example", which is what the title says.
- **`last_sent_queries()`**: no `@family`, confirmed. Its `@seealso` already
  points *to* *When marginplyr queries your data*; what is missing is the
[...]

Issue #464, opened 2026-09-06
Anchorless cross-links and duplicated blocks between guides


sayuks:
## Links that name a heading but land at the page top

`README.Rmd:60-62`, `get_started.qmd:541-542`, `:826-828`, `:912-913` read "[Put each subtotal with the rows it summarizes]", "[Compare each row with the right total]", "[Compute what the summary pass cannot express]" but link to `recipes.html` with no fragment, so the reader lands at the top of a seven-section page. `grouping_identity.qmd:311` shows the fragment form works. Add `#put-each-subtotal-with-the-rows-it-summarizes`, `#compare-each-row-with-the-right-total`, `#compute-what-the-summary-pass-cannot-express`.

`recipes.qmd:290-298` sends the reader to Get started for the `share_of_parent()` vs `share_of_total()` comparison, and `get_started.qmd:753-829` sends them back to this recipe. Link to `get_started.html#compare-a-summary-with-the-whole`.

`grouping_identity.qmd:271` is the only man-page anchor used from a vignette (`summarize_with_margins.html#display-labels-and-grouping-identity`) and nothing in `verify-site.R`'s markers covers it. Worth one check after the next render, or a marker.

## Duplicated blocks

- The `status` factor example is worked in both `get_started.qmd:500-533` and `grouping_identity.qmd:228-273` with different prose and no link between them. Give it one owner (tracked with the Get started issue).
- `database_backends.qmd:363-389`: one 27-line paragraph restates *Lazy execution boundaries* from `R/share.R:420-445` (probe/control, dialect caching, dropped connections, `.check_share_source` scoping). Keep three sentences plus the link at 439.
- `database_backends.qmd:701-710` (*Practical rules*) repeats 214-275 and 363-389; bullets ending "See ... above" are pointers and can shrink to one line each.
- `database_backends.qmd:664-683` restates README's verification table in prose (tracked with the SQLite scope issue).
- `grouping_identity.qmd:186-200`: the `show_query()` demonstration under *Inspect a Grouping plan as data* duplicates `get_started.qmd:852-864` and the database guide; a sentence with a link would do.

## Reference-to-guide links

[...]

PR #475, opened 2026-09-06
Correct what four guides claim about themselves

### LINKED ISSUE #458: recipes.qmd: four factual errors
Each of these was verified against the code.

1. **Line 48**: "Three recipes execute against DuckDB" — four do. DuckDB chunks (`eval: !expr has_duckdb`) appear in *Write the sort key yourself*, *Compare a row with an ancestor level*, *Name the grouping set on every row*, and *Compute what the summary pass cannot express*. Say "Four" or drop the count.
2. **Lines 614-617**: "dbplyr refuses the join outright" — the refusal is dplyr's `auto_copy()` (`` `x` and `y` must share the same source ``, with the `copy = "temp-table"` / `copy = "inline"` hints); dbplyr never sees the call. The next paragraph relies on the reader having read those hints, so say the join is refused by dplyr and that the message itself names the `copy` options.
3. **Line 122**: "`grouping_bit()` puts the Grouping bits the key reads into the result" — `margin_order_terms()` in `R/margin-operation.R` reads the bit from a hidden set-identifier column, which the reference (*Margin order*: "a column the result does not expose") also states. `grouping_bit()` produces an equal value, not the same column. Rephrase to "`grouping_bit()` exposes the same Grouping bits the key uses".
4. **Lines 564-566**: claims the Grouping identity guide "compares all four identity values and states their duplicate-occurrence rules". `grouping_identity.qmd:35-46` gives one sentence and defers that comparison to the `grouping_bit()` reference. Point to the reference (as `get_started.qmd:419-421` does), or describe what the guide actually does.

### LINKED ISSUE #459: completing_keys.qmd: no-op example, inexact VALUES claim, no entry or exit pointers
1. **Lines 135-144**: the "display replacement" example is a no-op. `facts$revenue` has no `NA`, so `mean_revenue` is never `NA` and `coalesce(mean_revenue, 0)` changes nothing, while the prose says it "replace[s] the finished display value". Use a case where a value is actually replaced.
2. **Lines 240-241**: "embeds the rows in the lazy SQL with `VALUES`" is only approximately true on SQLite, where dbplyr renders `copy_inline()` as a zero-row `SELECT NULL ... WHERE (0 = 1) UNION ALL VALUES (...)` subquery with casts. The two `show_query()` outputs (264 and 279) are printed without commentary; say "a `VALUES` subquery" and tell the reader what to look for.
3. **Lines 16-20**: the opening paragraph is all house terms ("Grouping sets requested by a Grouping specification", "synthetic facts", "Margin operation") and, unlike `database_backends.qmd:20-22`, never points a newcomer to Get started.
4. **No next-step pointer**: the guide links nowhere at the end. `database_backends.qmd:443-451` and `get_started.qmd:831-841` link here; the return edges are missing, and the database guide is the natural continuation of the lazy section (181-289).
5. **Line 29** (also `database_backends.qmd:45`, `get_started.qmd:43`): "Only dplyr and dbplyr are required" is true of what the user loads, but DESCRIPTION imports cli, rlang, tidyselect, and glue too. "Nothing beyond marginplyr's own dependencies" avoids the question.

### LINKED ISSUE #460: Get started carries advanced contracts that stop a first read
## Problem

Two blocks in `vignettes/get_started.qmd` belong to a reader who already uses the package, not to one deciding whether to:

- **Lines 493-533**: the `NA_character_` vs `NULL` distinction, the factor NA level vs missing code, and the eight-case factor contract, with a worked `structure(c(1L, 2L, NA), levels = c("kept", NA))` example. The same example is worked again in `grouping_identity.qmd:228-273` with different prose, and neither links to the other.
- **Lines 650-662**: refusal by spelling, call-head resolution against the calling environment, `(pick)(units)` and `(pick(units))`, why `cur_group_id()` is stricter than `dplyr::summarize()`.

Smaller points in the same guide:

- **Line 16**: "grain" is used from the first sentence and never defined. README.Rmd:102 defines it in passing.
- **Line 1126**: "The ordered factor makes it the first tab" — the code at 1106 calls `factor()`, not `ordered()`. Say "the factor's level order".
- **Line 1075-1079**: the paragraph on `dplyr::nest_by()` possibly being deprecated upstream is a maintainer's note, not guidance a first reader needs.

## Proposal

Keep one sentence plus a link for each advanced block, and let `grouping_identity.qmd` own the factor NA example and the `summarize_with_margins` reference own the resolution rules. Define "grain" at first use.

sayuks:
Four documentation issues from the 2026-09-06 review, taken in one pass
because three of them are the same kind of defect: a guide stating something
the code does not do.

**#458** -- the DuckDB chunk count, the claim that `grouping_bit()` puts the
bits the Margin order key reads into the result, and a pointer crediting the
Grouping identity guide with a comparison it defers to the `grouping_bit()`
reference. The issue's fourth item was verified wrong and is not implemented
as written: the lazy-join refusal is dbplyr's, raised by
`dbplyr:::dbplyr_auto_copy()`, so the sentence keeps its subject and gains
only what the reader needs from the diagnostic.

**#459** -- the post-summary example replaced nothing, because `facts$revenue`
has no missing value. It now summarizes the observed facts out of
`completed_facts`, where a completed key with no observed fact averages an
empty selection. `copy_inline()` renders a `VALUES` subquery, not bare
`VALUES`. The guide gains an entry pointer to Get started and an exit pointer
to Database backends, closing a link the database guide already made the other
way.

**#460** -- the `NA_character_`/`NULL` factor block and the call-head
resolution block both belong to a reader who already uses the package, and
both have an owner: *Three kinds of apparent missingness* in the Grouping
identity guide, and *Relationship to dplyr summaries* in the reference. Get
started keeps a sentence and a link for each. "Grain" is defined at first use;
an ordered factor that is a plain `factor()` is named correctly; a note on
`dplyr::nest_by()`'s upstream stability is dropped.

**#464 / #465, in part** -- four links that named a heading and landed at a
page top now carry its fragment, and `last_sent_queries()` is connected to its
only guide coverage in both directions. The rest of both issues is left open.

"Only dplyr and dbplyr are required" answered a question about `library()`
calls with a claim about the dependency closure; all four guides now say
nothing beyond marginplyr's own dependencies is required.

Closes #458
Closes #459
Closes #460
Refs #464, #465

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -5,6 +5,12 @@ consistently to local and remote data.
 
 ## Language
 
+**Grain**:
+What one row of a summary stands for: the dimensions it is grouped by. An
+ordinary grouped summary has one grain; a Grouping plan produces several in
+one result.
+_Avoid_: Granularity, aggregation level
+
 **Grouping specification**:
 A user-declared description of the grouping sets, rollups, cubes, or products
 to be applied.
--- a/R/sent-queries.R
+++ b/R/sent-queries.R
@@ -132,8 +132,12 @@ record_sent_query <- function(purpose, query) {
 #' ```
 #'
 #' @seealso *[When marginplyr queries your data][summarize_with_margins]* for
-#'   the queries a Margin verb sends without being asked, and
-#'   [dplyr::show_query()] for the result query alone.
+#'   the queries a Margin verb sends without being asked,
+#'   [dplyr::show_query()] for the result query alone, and *Record the SQL
+#'   marginplyr sends* in the [database-backends guide][guide] for the record
+#'   read against a live connection.
+#'
+#' [guide]: https://sayuks.github.io/marginplyr/vignettes/database_backends.html
 #' @export
 #' @examples
 #' # The record is kept only while the option is set, so the example restores
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -633,6 +633,9 @@
 #' [summarize_with_margins()]'s `.check_share_source` reads nothing on any
 #' backend, so it defaults to `TRUE` there.
 #'
+#' [last_sent_queries()] reports what a call actually sent, which is how this
+#' rule is checked against a connection rather than read here.
+#'
 #' @family summarize and expand data with margins
 #' @export
 #' @examples
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -633,8 +633,9 @@
 #' [summarize_with_margins()]'s `.check_share_source` reads nothing on any
 #' backend, so it defaults to `TRUE` there.
 #'
-#' [last_sent_queries()] reports what a call actually sent, which is how this
-#' rule is checked against a connection rather than read here.
+#' [last_sent_queries()] reports what a call actually sent, which is how the
+#' two unasked queries above are counted against a connection rather than
+#' taken from this page.
 #'
 #' @family summarize and expand data with margins
 #' @export
--- a/README.Rmd
+++ b/README.Rmd
@@ -58,7 +58,7 @@ Row order is unspecified: the same call on the same data in a different order
 can arrange the result differently. `.sort` asks for a Margin order, which puts
 each subtotal with the rows it summarizes, and `arrange()` gives any other
 presentation order. [Put each subtotal with the rows it
-summarizes](https://sayuks.github.io/marginplyr/vignettes/recipes.html) shows
+summarizes](https://sayuks.github.io/marginplyr/vignettes/recipes.html#put-each-subtotal-with-the-rows-it-summarizes) shows
 `.sort` at work and what it does not promise.
 
 - Works with local data frames and lazy `dbplyr` tables.
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ order can arrange the result differently. `.sort` asks for a Margin
 order, which puts each subtotal with the rows it summarizes, and
 `arrange()` gives any other presentation order. [Put each subtotal with
 the rows it
-summarizes](https://sayuks.github.io/marginplyr/vignettes/recipes.html)
+summarizes](https://sayuks.github.io/marginplyr/vignettes/recipes.html#put-each-subtotal-with-the-rows-it-summarizes)
 shows `.sort` at work and what it does not promise.
 
 - Works with local data frames and lazy `dbplyr` tables.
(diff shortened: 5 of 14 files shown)


sayuks:
## Review record

`mattpocock-skills:code-review`, two axes in parallel, against
`git diff origin/main...HEAD` with the merge-base at 671d8b1. Standards
carried `AGENTS.md` (*Linting*, *Documentation*, *Chunks that must fail*,
*Site verification*, *Dependency metadata*, *Code comments*), the Fowler smell
baseline, and *Answering a review* in `design/architecture.md` as the bound on
what a finding about prose may be. Spec carried #458, #459, #460, #464 and
#465, each with its triage verification comment, since three items in those
issues were verified wrong and are deliberately not implemented as written.

**Standards: no hard violations.** Both axes independently re-ran the
corrected claims rather than reading them: four `##` sections hold
`has_duckdb` chunks, the tabset chunk builds an unordered `factor()`, SQLite
renders `copy_inline()` as a zero-row `SELECT` unioned with `VALUES` inside
casts, and the `join-without-copy` refusal is dbplyr's and carries no
marginplyr class, so `must_error: true` is the correct form there and the
class form would fail. No deleted prose was serving as a `verify-site.R`
marker; all five new anchors resolve to real headings.

Five findings, dispositioned:

|---|---|---|
| "this rule" in the `last_sent_queries()` pointer takes `.check_share_source`'s default as its antecedent, which that function cannot check | Spec | fixed in prose, 77db7ba |
| the completion guide calls the produced value missing where it renders `NaN` | Spec | fixed in prose, 77db7ba |
| `CONTEXT.md` uses "multi-grain" in its first sentence and defines no **Grain**, so Get started's new gloss has no owner and can drift | Standards | fixed by adding the glossary entry, 77db7ba |
| `get_started.qmd`'s `[Recipes]` link is a fifth anchorless link the triage comment on #464 called "same fix" | Spec | rejected. Its link text names the guide, and the sentence after it names two recipes; one link cannot carry two fragments. The triage comment overstated it and has been corrected on #464. |
| the "Nothing beyond marginplyr's own dependencies" paragraph is near-identical across four vignettes | Standards | rejected as pre-existing shape. The branch changed one sentence inside a paragraph whose duplication predates it; each copy names a different package list, which is the part a reader needs in place. |

The `mean(revenue[is_fact])` recomputation in `completing_keys.qmd` was raised
as possible Duplicated Code and withdrawn by the reviewer in the same
sentence: the new chunk needs a named object to `mutate()`.

Gates after the answers: `lintr::lint_package()` clean,
`spelling::spell_check_package()` clean, `roxygen2::roxygenise()` committed,
`verify-must-error.R` and `verify-context-budget.R` pass, full `testthat`
suite green.

PR #482, opened 2026-09-06
Record why the dtplyr reference pages stay unqualified (#470)

### LINKED ISSUE #470: Reference pages claim dtplyr support without naming the mutable-step refusal
Split out of the review round on #469 (fixes #451), as a finding that branch's Acceptance does not reach.

#469 refuses a dtplyr step whose root was built with `dtplyr::lazy_dt(immutable = FALSE)` (ADR 0029). Its documentation deliverables were bounded to three places: **Mutable step** in `CONTEXT.md`, the ADR, and a `must_error` chunk in `vignettes/database_backends.qmd`. The reference pages were not among them, and several of them state dtplyr support without qualification:

- `R/nest_with_margins.R:16` — `@param .data A local data frame or a `dtplyr` step.`
- `R/nest_with_margins.R:4` — "It works with local data frames and `dtplyr` steps."
- `R/share.R:377` — "Share execution supports local data frames and lazy dbplyr and dtplyr inputs".
- `R/share.R:402` — the backend table's `` `dtplyr` step `` row.
- `R/nest_by_with_margins.R:19` and `R/grouping-context.R:78` — the same unqualified naming.

ADR 0025 is the precedent #469 cites, and its *Documentation consequences* updated "the pages that claim Arrow summaries are supported and lazy" in the same change. The parallel step was not taken here.

## What is worth deciding

Whether these sentences are false or merely unqualified. A mutable step is a dtplyr step and is now refused, so the `.data` sentence over-claims; but the class is still supported, and every construction `lazy_dt()` produces by default is accepted, which is a weaker gap than the one 0025 closed — there a whole class of summaries stopped working.

So this is a scoping question rather than a defect: one clause on the pages that name the input class, or nothing on the grounds that the refusal names its own rewrite and `?marginplyr` already promises the class a caller catches.

## Acceptance

Either a clause added to the pages named above and `man/` regenerated with `roxygen2::roxygenise()`, or the decision recorded in ADR 0029's *Documentation consequences* that they stay as they are, with the reason.

sayuks:
Closes #470.

## What was decided

The reference pages that name `dtplyr` as a supported input keep their wording, and ADR 0029's *Documentation consequences* now records that as a decision with its reason. #470 offered either that or a clause added to those pages; this takes the first.

The reason has two halves, both checkable:

- ADR 0025's page edit — the precedent #470 cites — carried what a caller could not otherwise know. The Mutable step refusal is reached only by a caller who wrote `dtplyr::lazy_dt(immutable = FALSE)` themselves, against that argument's default of `TRUE`, and the diagnostic names the one rewrite.
- `@param .data` is not where this package documents a refusal. `summarize_with_margins()`'s and `inspect_grouping()`'s *A data frame or lazy table* name none of the refusals a Margin verb already raises — the Arrow ones are carried by sections. So `nest_with_margins()`'s is the only definition a dtplyr clause would fit, and there it would read as nesting's refusal rather than one raised for every verb `verbs_taking(".grouping")` returns.

## What #470 got wrong

Three of the six sites it lists claim nothing about what may be passed, and the ADR now says so rather than leaving them to be re-filed:

- `R/nest_by_with_margins.R` — when a `dtplyr` *result* is collected to be made row-wise.
- `R/grouping-context.R` — the integers a step *returns* for a grouping identity.
- The `share.R` backend table — what establishes the source rules on an input that was accepted. A Mutable step never reaches share execution.

The other three do claim the class, and the ADR names each and leaves it broader.

## Checks

No R source changed, so `man/`, `NAMESPACE`, and `README.md` are untouched and `document.yaml` has nothing to regenerate. Nothing under `tests/` or `.github/scripts/` reads `design/adr/`. `devtools::test()`: 6981 pass, 0 fail, 0 skip. `Rscript .github/scripts/verify-context-budget.R` passes (22005 bytes, at the baseline — ADRs are outside the always-loaded closure).

The review record is in a comment below.

### DIFF
--- a/design/adr/0029-refuse-a-step-a-backend-would-write-to.md
+++ b/design/adr/0029-refuse-a-step-a-backend-would-write-to.md
@@ -129,22 +129,30 @@ and the one the Arrow refusal beside it already takes.
 that page already promises.
 
 The reference pages are unchanged too, and that is decided rather than passed
-over (#470). A Mutable step is a dtplyr step, so `nest_with_margins()`'s
-`@param .data` — *A local data frame or a `dtplyr` step* — is broader than
-what is accepted, and it is left broader. This is where the parallel with ADR
-0025 stops. There the page had to carry what a caller could not otherwise know,
-which summaries Arrow absorbs being Arrow's to decide and moving with its
-version; here the refusal is reached only by a caller who wrote `immutable =
-FALSE` themselves, against `lazy_dt()`'s default, and the diagnostic names the
-one rewrite. A clause would have to go on every `@param .data` definition in
-the package, the refusal being raised for every verb `verbs_taking(".grouping")`
-returns, and nothing would hold those copies in step — an amendment here would
-leave the ones it did not reach standing and wrong.
+over (#470). A Mutable step is a dtplyr step, so the three sentences #470 names
+that claim the class are broader than what is accepted, and they are left
+broader: `nest_with_margins()`'s `@param .data` — *A local data frame or a
+`dtplyr` step* — its *It works with local data frames and `dtplyr` steps*, and
+*Share execution supports local data frames and lazy dbplyr and dtplyr inputs*.
 
-The other pages #470 names describe an input that was accepted: when a
-`dtplyr` result is collected to be made row-wise, the integers a step returns
+This is where the parallel with ADR 0025 stops. There the page had to carry
+what a caller could not otherwise know; here the refusal is reached only by a
+caller who wrote `immutable = FALSE` themselves, against `lazy_dt()`'s default,
+and the diagnostic names the one rewrite.
+
+`@param .data` is also not where this package documents a refusal. The two
+definitions naming no backend — `summarize_with_margins()`'s and
+`inspect_grouping()`'s *A data frame or lazy table* — name none of the
+refusals a Margin verb already raises, the Arrow ones being carried by sections
+instead.
+So `nest_with_margins()`'s is the only definition a dtplyr clause would fit,
+and putting it there alone would say the refusal is nesting's when it is raised
+for every verb `verbs_taking(".grouping")` returns.
+
+The other three sentences #470 names describe an input that was accepted: when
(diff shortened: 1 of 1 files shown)


sayuks:
## Review record

Round run with `mattpocock-skills:code-review` against `origin/main...HEAD` at `c827108`, two axes in parallel. Standards sources passed: `AGENTS.md` *Code comments*, *Documentation*, *Investigation notes*; `design/architecture.md` *Answering a review*; `design/agents/code-review.md`. Spec: #470.

Three findings, all about the paragraph `c827108` added. Answered in `20d9fab`.

|---|---|---|
| The closing sentence quantified over "the other pages #470 names" and asserted none claims what may be passed. #470 names six sites; three were accounted for, and two of the omitted ones are class-support claims — so the universal was false on the set it quantified over. | Standards (hard), Spec (a) | Fixed in prose. All six are now disposed of by name. |
| The reason recorded — a clause would go on every `@param .data` definition with nothing holding the copies in step — describes a duplication the package does not have: of the three definitions only `nest_with_margins()`'s names dtplyr, and the other two are reached by `@inheritParams`, which is what holds copies in step. | Spec (c) | Fixed in prose. Replaced with a checkable reason. |
| A clause re-derived ADR 0025's own "decided by the backend's version" argument beside the citation to it. | Standards (judgement call) | Fixed in prose. Clause cut; the citation carries it. |

Axes that raised nothing: Spec (b) scope creep — none; `?marginplyr` duplication — none, the existing sentence stands alone; missing artifact — none, `man/` regeneration is required only on the other branch of #470's Acceptance and no roxygen changed.

No second round: `design/agents/code-review.md` runs one again only over executable behaviour the first did not see, and this diff has none. Each answer landed as the diff that changed the prose, which is what *Answering a review* records a prose finding by.

Local: `devtools::test()` — 6981 pass, 0 fail, 0 skip (at `c827108`; `20d9fab` touches no file under `tests/` or `.github/scripts/`). `Rscript .github/scripts/verify-context-budget.R` passes.

PR #483, opened 2026-09-06
Announce three user-visible features in NEWS.md (#481)

sayuks:
Closes #481.

## What changed

Three user-visible features had **no** `NEWS.md` entry at all. One bullet each now, in the shape #466 established: the user-visible change in one or two sentences, an issue number where one exists, and a pointer to the reference section that owns the argument.

| Feature | Issue | Points at |
| --- | --- | --- |
| `.margin_label_position` | #16 | *Display labels and grouping identity* in `?summarize_with_margins` |
| `retail_sales` | — | `?retail_sales` |
| `.key` | — | *Relationship to tidyr and dplyr* in `?nest_with_margins` |

Each bullet sits with the ones covering its own feature area — `.margin_label_position` after `.check_margin_label`, `retail_sales` before the guides entry, `.key` at the head of the nesting group — rather than at the end of the file, which is how #466 ordered the file it left.

## Two choices worth naming

**`.margin_label_position` does not point at *Option arguments*, which #481 named.** That section owns the string-vocabulary rule — an abbreviation and `NULL` both refused — and the existing option-arguments bullet already cites it, listing this argument as one of its examples. What the argument *does* is written in *Display labels and grouping identity*. Pointing at the section #481 named would send a reader to prose that says nothing about the argument's effect.

**The `.key` bullet carries the `NULL` asymmetry between the two nesting verbs.** `nest_with_margins()` reads `NULL` as `"data"`, as `tidyr::nest()` does; `nest_by_with_margins()` refuses it, as `dplyr::nest_by()` does. Nothing else in the file states this, and folding the two verbs into one bullet — which the neighbouring collision-free-columns bullet already does — otherwise reads as though they treat the argument alike. Both halves were confirmed against the loaded package rather than read off the roxygen:

```r
names(nest_with_margins(d, .grouping = rollup(g), .key = NULL))
#> [1] "g"    "data"
nest_by_with_margins(d, .grouping = rollup(g), .key = NULL)
#> Error: `.key` must be a character vector of length 1.
```

## Issue numbers

`.margin_label_position` came from #16. Neither `retail_sales` (added in `e5844fc`) nor `.key` (added in `4f4197a`, before issue #1) has one — `4f4197a` predates the issue tracker. #481's "an issue number where one exists" is what admits the two uncited bullets.

## `CONTEXT.md` is unchanged

[...]

Issue #481, opened 2026-09-06
NEWS.md announces no entry for three user-visible features


sayuks:
## Problem

Found while re-taking #466's "not covered" list. Three user-visible features have **no** `NEWS.md` entry at all — not a shortened one, none:

- **`.margin_label_position`** — in the `summarize_with_margins()` signature and documented under *Option arguments*. Since #466 it appears in `NEWS.md` only as an *example* of an option argument, never as a feature that was added.
- **`retail_sales`** — `LazyData: true`, `man/retail_sales.Rd`, used by every guide and example. Zero mentions.
- **`.key`** on `nest_with_margins()` and `nest_by_with_margins()` — zero mentions.

## Why this is a separate ticket

#466 named its own "not covered" list, and its triage comment scoped the re-take to what the file gained after filing (`2390e9d`). These three predate that and were missing from the 36-bullet file the issue measured, so they are outside what that branch's Acceptance reaches — `design/architecture.md`'s *Answering a review*: "A finding this branch's Acceptance does not reach becomes a ticket rather than prose added here."

## Proposal

Add one bullet each, in the shape #466 established: one or two sentences naming the user-visible change, an issue number where one exists, and a pointer to the reference section that owns the argument.

Found in the code review of #480 (PR for #466).

Issue #466, opened 2026-09-06
NEWS.md entries are design notes, not a changelog


sayuks:
## Problem

`NEWS.md` has 36 bullets over 346 non-blank lines. Single bullets run to 30+ lines and carry the argument for a decision (how nested specifications are recognized, why `(s)` is transparent, what `rollup(area = region)` reports) rather than the change. A user scanning for what changed cannot find it, and the argument already lives in the ADRs and the reference.

Also not covered: the *Option arguments* contract (abbreviations and `NULL` refused, full-vector forwarding), `inspect_grouping(.format = )`. The `nest_*` `.keep` semantics appear only as a one-line bug-fix note (296-297).

## Proposal

One or two sentences per bullet naming the user-visible change and the issue number; the rationale stays in the ADR or reference it already has. Add the missing entries.

Found in a documentation review on 2026-09-06.

sayuks:
> *This was generated by AI during triage.*

## Verification (against 671d8b1)

Confirmed, with the counts updated: `NEWS.md` is now **38 bullets over 365
non-blank lines** (was 36 over 346 when filed) — 2390e9d added entries after
this issue went up. The shape the issue describes is unchanged: single bullets
running 30+ lines carrying the argument for a decision rather than the change.

Re-check the "not covered" list against the current file before writing, since
it grew in between.

PR #487, opened 2026-09-06
Delete a clause `?retail_sales` already owns

sayuks:
Follows #483, which merged while its third review round was still running.

## Why this is a separate PR

#483's merge took `1dac813`, not the `f6fcec4` that carried this fix, so the clause below reached `main`. Nothing on `main` is broken by that: every commit the merge took passed every gate, both round-1 defects are fixed there, and the three bullets #481 asked for are present and correct. This is the one clause that did not make it.

## What changed

```diff
-* Added the `retail_sales` data set, 24 rows of synthetic monthly sales that
-  the guides and examples are written against. See `?retail_sales`.
+* Added the `retail_sales` data set, 24 rows of synthetic monthly sales.
+  See `?retail_sales`.
```

## The clause failed two different tests in two rounds

**Round 2 — false as a universal.** It said the data set is what "the guides and examples are written against". `vignettes/completing_keys.qmd` builds a `facts` tibble of its own and names `retail_sales` nowhere.

**Round 3 — duplication, after narrowing.** Narrowing it to "the documentation is written against" made it true; the round-3 Standards axis confirmed as much by counting every documentation surface (11/11 `man/*.Rd`, `README.Rmd`, 4/5 vignettes, no extra page in the site nav — 16 of 17 files). But it left the defect that matters: `?retail_sales` opens

> A small synthetic data set used throughout the package documentation.

and the bullet asserting that same fact points at that very topic. That is the duplication #466 removed 178 lines to fix.

So the clause goes rather than being narrowed a second time. #483's *first* review round had already applied that disposition to the identical defect one sentence over — `1dac813` deleted a sentence for "restating the Description of the very topic the bullet points at". Round 2 met the same kind of finding on the adjacent clause and rewrote instead, which is what produced round 3's finding. What survives names the change — a data set was added, and its size — which is what #466's shape asks a bullet for.

## Where the overclaim came from

#481's own Problem paragraph: "`retail_sales` — `LazyData: true`, `man/retail_sales.Rd`, **used by every guide and example**." Adopted without checking. That is the second sentence of that ticket #483 could not take verbatim; the first was its *Option arguments* pointer, which names where `.margin_label_position` is *mentioned* rather than what owns it. Both were caught by review, neither by a gate.

## Gates

- `testthat::test_local()` — running at the time this description was written; result posted as a comment below.
- `spelling::spell_check_package()` — no spelling errors.
- `verify-doc-references.R` — 176 paths across 153 documents, all resolve.
- `verify-context-budget.R` — 22005 bytes, within 22005.
- `lintr::lint_package()` — no lints.

## Review

Three rounds ran on #483, the third landing after its merge; this PR is that round's answer. Round 3 also verified the rest of the merged diff from scratch and found it clean — `.margin_label_position` on all four verbs with #16 correctly cited and its section owning both sentences, `.key`'s asymmetry owned by both `\item{.key}` entries and confirmed by execution, House terms correct, and the three differing pointer forms each citing what owns its own claim rather than being an inconsistency.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -40,8 +40,8 @@
   structure of its Grouping plan rather than by displayed values, so a subtotal
   sits with the rows it summarizes rather than wherever its Margin label falls,
   and lazy inputs stay lazy. See *Margin order* in `?summarize_with_margins`.
-* Added the `retail_sales` data set, 24 rows of synthetic monthly sales that
-  the guides and examples are written against. See `?retail_sales`.
+* Added the `retail_sales` data set, 24 rows of synthetic monthly sales.
+  See `?retail_sales`.
 * Added guides for Grouping identity and explicit key completion, and made the
   function references the canonical source of the Margin, Parent-share, and
   Margin-label contracts.


sayuks:
### Test suite result

`testthat::test_local()` — pass, exit code 0, no failures, warnings, or skips.

This is the gate the description could not state when it was written. It asserted the result before the run finished; that line was corrected to say the run was still going, and this comment is it. The change is `NEWS.md` prose only, and no test reads `NEWS.md` — the file is named in `altdoc/quarto_website.yml`'s `$ALTDOC_NEWS` slot and in `verify-site.R`'s `imported_sources` table, neither of which reads its content — so the suite was never going to be the gate that catches a defect here. Three review rounds were. Running it anyway is what makes that a measurement rather than an assumption.

All five gates now stand:

- `testthat::test_local()` — pass, exit 0, no failures, warnings, or skips.
- `spelling::spell_check_package()` — no spelling errors.
- `verify-doc-references.R` — 176 paths across 153 documents, all resolve.
- `verify-context-budget.R` — 22005 bytes, within 22005.
- `lintr::lint_package()` — no lints.

PR #490, opened 2026-09-06
Measure cran-comments.md against the tree it describes (#65)

### LINKED ISSUE #65: Refresh cran-comments.md against the final tree
## Parent

Part of #23. Found by the final readiness review (#40).

## Problem

`cran-comments.md` reports test counts the current tree does not produce. It
records "1487 tests pass" for the fully provisioned run and "1083 tests pass"
for the dependency-only run. A tarball built from `bd0b3b3` and checked on
R 4.6.1, aarch64-apple-darwin23, Quarto 1.9.38 produces 1506 and 1102. The
counts were accurate when #37 wrote them and drifted as #55, #56, and #57
added tests.

This is the file CRAN reads, so a stale number in it is a claim the submission
makes and cannot support.

## What is still accurate

The rest of the file holds and should not be rewritten for its own sake:

- Both runs are `Status: OK`, so "0 errors | 0 warnings" is right, and the
  single NOTE is the new-submission NOTE.
- The skip accounting is right: the dependency-only run skips 64 tests for a
  missing optional package, 2 for no supported lazy backend, and 2 for CRAN
  snapshot semantics, which is the file's "66 ... and the other 2".
- The provisioned run's 3 skips are the `expect_snapshot()` diagnostics.

## Acceptance criteria

- [ ] The counts in `cran-comments.md` match a check of the tarball built from
      the tree being submitted, and the file names the R version, platform,
      and Quarto version they came from.
- [ ] The two commands behind them are stated precisely enough to re-run:
      `R CMD check --as-cran` and `_R_CHECK_DEPENDS_ONLY_=true R CMD check
      --as-cran`, both against one `R CMD build` tarball.
- [ ] Whatever keeps these numbers honest is written down, since they drift
      whenever a test is added. Either the file stops quoting exact counts and
      states what it is actually claiming, or refreshing it becomes a named
      step of the release procedure. State which, and why, in the file or in
      `AGENTS.md`.
- [ ] `design/review-dispositions.md` section 8 is updated from "not fixed" to
      the disposition this ticket reaches.

## Blocks

- #40

## Blocked by

GitHub's native issue dependencies are the authority; this list mirrors them so
the blockers are readable without leaving the ticket.

- #162 — adds tests for the `<-` / `for` / `local()` scoping shapes
- #163 — adds tests for the one-sided-formula sites

Both add tests, so refreshing the counts here before they land would restage
the same drift this ticket exists to remove.

sayuks:
Closes #65.

## What changed

`cran-comments.md` reported test counts no current run produces — 1487 and
1083, accurate when #37 wrote them and moved by every test added since. It
also called five vignettes four, named DBI among the packages the
dependency-only run proves absent, and omitted data.table.

Both runs were re-measured against **one** tarball built by `R CMD build`
from this tree.

| | |
| --- | --- |
| R | 4.6.1 |
| Platform | aarch64-apple-darwin23 |
| OS | macOS Tahoe 26.6.2 |
| Quarto | 1.9.38 |

| Run | Command | Result |
| --- | --- | --- |
| Fully provisioned | `R CMD check --as-cran` | `Status: 1 NOTE` (new submission). `FAIL 0 | WARN 0 | SKIP 13 | PASS 6953`; all 13 skips are `On CRAN`. |
| Dependency-only | `_R_CHECK_DEPENDS_ONLY_=true _R_CHECK_FORCE_SUGGESTS_=false R CMD check --as-cran` | `Status: 1 NOTE` (new submission). `FAIL 0 | WARN 0 | SKIP 215 | PASS 5257`; dtplyr 84, arrow 39, duckdb 39, RSQLite 38, data.table 8, `On CRAN` 7. |

The dependency-only entry previously claimed "status OK, no notes". That was
true of a run without `--as-cran`; with the command the acceptance criteria
name, the incoming-feasibility NOTE appears in both runs, so the file's
`0 errors | 0 warnings | 1 note` heading now covers both.

The tarball was built from a `git archive` export rather than in place: in a
git worktree `.git` is a regular file, which `R CMD build`'s exclusion does not
catch, and the resulting tarball draws a hidden-files NOTE that a submission
built from an ordinary clone would not.

## Acceptance criteria

- [x] Counts match a check of the tarball built from the tree being submitted,
      and the file names R version, platform, and Quarto version.
- [x] Both commands are stated precisely enough to re-run, against one
      `R CMD build` tarball. `_R_CHECK_FORCE_SUGGESTS_=false` is named too:
      without it `--as-cran` treats the deliberately withheld Suggests as a
      failure, so the command as previously written could not be re-run.
- [x] **The file stops quoting exact counts**, and says what it is claiming
      instead. Rationale is in `cran-comments.md` itself.
- [x] Answered in https://github.com/sayuks/marginplyr/issues/65#issuecomment-5562726900:
      `design/review-dispositions.md` was deleted by #288, so there is no
      section 8 to update, and this pull request is the record #288 put in its
      place.

### Why the counts go rather than get a refresh step

A count is accurate on the day it is measured and silently wrong afterwards;
nothing in this repository can check one; and a stale count reads exactly like
a fresh one. The ticket's alternative — making a refresh a named step of the
release procedure — has no home that is not worse:

1. a new `design/release.md` is the unchecked maintained file #288 retired;
2. `AGENTS.md` spends always-loaded context budget on a step only a release
   reaches, against that file's own rule for where an argument goes;
3. the *Publication day* comment in `tests/testthat/test-documentation.R`
   explicitly draws the line at "not when `cran-comments.md` is written".

What the runs are actually offered as evidence for — the status line and the
account of the skips — survives a test being added, and is what the file now
states.

## An additional correction the measurement forced

The *Optional backends* section claimed "none of those backends is installed
in [the dependency-only run]" while listing DBI. dbplyr imports DBI, so it is
installed in that run and nothing there skips for its absence — the fact
`optional_suggest_spec()` records as `DBI = FALSE` and `AGENTS.md` states under
*Dependency metadata*. data.table, which the run does withhold, was missing
from the list. The section now names the five packages actually withheld and
states DBI's position separately.

## Review

Run after this branch is pushed; the record follows in a comment.

### DIFF
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -23,12 +23,14 @@ Quarto 1.9.38.
   pass. Every skipped test is an `expect_snapshot()` diagnostic, which
   testthat skips on CRAN by default.
 * `_R_CHECK_DEPENDS_ONLY_=true _R_CHECK_FORCE_SUGGESTS_=false R CMD check
-  --as-cran`, Suggested packages withheld. Examples, tests, and vignettes all
+  --as-cran`, against a library holding only marginplyr's hard dependencies,
+  testthat, the declared `VignetteBuilder`, and what those require; every
+  other Suggested package is absent. Examples, tests, and vignettes all
   complete. Every skipped test is either one of those same snapshot
   diagnostics or a test whose optional backend is absent, and the testthat
   output names the missing package for each of the latter.
   `_R_CHECK_FORCE_SUGGESTS_=false` is what stops `--as-cran` from treating the
-  deliberately withheld packages as a failure.
+  absent packages as a failure.
 
 Neither entry quotes a test count, deliberately: a count is accurate on the day
 it is measured and silently wrong afterwards. What these runs are offered as
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -9,36 +9,50 @@ Maintainer: 'Yusuke Sasaki <sayuks.dev@gmail.com>'
 New submission
 ```
 
-This is the expected new-submission NOTE; the run reports nothing else.
+This is the expected new-submission NOTE; both runs below report it and
+nothing else.
 
 ## Test environments
 
-Both runs below check the built source tarball, not the development tree, on
-R 4.6.1, aarch64-apple-darwin23, macOS Tahoe 26.5.2, with Quarto 1.9.38.
+Both runs check one source tarball produced by `R CMD build`, not the
+development tree, on R 4.6.1, aarch64-apple-darwin23, macOS Tahoe 26.6.2, with
+Quarto 1.9.38.
 
-* Fully provisioned `R CMD check --as-cran`, every Suggested package
-  installed: 0 errors, 0 warnings, 1 note. Installation, examples, tests,
-  vignette re-building, and the PDF and HTML manuals all pass. 1487 tests
-  pass; the 3 skips are the `expect_snapshot()` diagnostics that testthat
-  skips on CRAN by default.
-* `_R_CHECK_DEPENDS_ONLY_=true R CMD check`, Suggested packages absent:
-  status OK, no notes. Examples, tests, and vignettes all complete. 1083
-  tests pass; 66 of the 68 skips are backend tests whose optional package is
-  missing, and the other 2 are the same snapshot skips as above.
+* `R CMD check --as-cran`, every Suggested package installed. Installation,
+  examples, tests, vignette re-building, and the PDF and HTML manuals all
+  pass. Every skipped test is an `expect_snapshot()` diagnostic, which
+  testthat skips on CRAN by default.
+* `_R_CHECK_DEPENDS_ONLY_=true _R_CHECK_FORCE_SUGGESTS_=false R CMD check
+  --as-cran`, Suggested packages withheld. Examples, tests, and vignettes all
+  complete. Every skipped test is either one of those same snapshot
+  diagnostics or a test whose optional backend is absent, and the testthat
+  output names the missing package for each of the latter.
+  `_R_CHECK_FORCE_SUGGESTS_=false` is what stops `--as-cran` from treating the
+  deliberately withheld packages as a failure.
+
+Neither entry quotes a test count, deliberately. A count changes whenever a
+test is added, so one written here is accurate on the day it is measured and
+silently wrong afterwards, and a stale count reads exactly like a fresh one.
+What these runs are offered as evidence for is the status line above and the
+account of the skips, and both of those survive a test being added.
 
 ## Optional backends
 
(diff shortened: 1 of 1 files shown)


sayuks:
## Review record

Reviewed against the merge-base with `origin/main` (`78e1f5e`) through
`mattpocock-skills:code-review`, both axes, three rounds. Standards sources:
`AGENTS.md`, `CONTEXT.md`, `design/architecture.md` (including *Answering a
review*), `design/agents/code-review.md`, and the ADRs. Spec sources: #65 and
its pre-implementation comment, #112, #288.

### Round 1 — four findings, all answered in 18569f4

**Standards and Spec both raised the same hard finding.** `data.table` was
listed among the Suggests of which "each backend adds one optional lazy path".
It adds none: raw `data.table` input reaches the local backend as a data frame
subclass (#176), which is what its `backend` job proves, per the comment above
`optional_suggest_spec()` and *Release matrix* in `AGENTS.md`. The claim also
sat oddly beside "local data frames need none of them", a data.table being one.
The sentence now says which of the two things each package adds.

**The DBI paragraph re-derived an argument another file owns** — why DBI is
declared despite dbplyr importing it, which is `AGENTS.md`'s under *Dependency
metadata* and the helper's own comment. *Answering a review*'s second category.
It now states the fact and stops.

**"both runs report it and nothing else"** read past the dependency-only run's
`checking package dependencies ... INFO`. Narrowed to the note, which is what
the heading above it already scopes.

**The rationale paragraph** lost the half addressed to a maintainer rather than
to a CRAN reviewer.

### Round 2 — one finding, answered in ab4a86b

Round 1's DBI answer made a pre-existing falsehood visible: the dependency-only
bullet was labelled "Suggested packages withheld" while the section two below
it said a Suggest was installed. Six are — DBI and tidyr via dbplyr's Imports,
tibble via dplyr's, knitr via quarto → rmarkdown, plus testthat and quarto
themselves. The bullet now names the library the run used instead of claiming a
withholding that did not happen.

### Round 3 — clean, both axes

Round 2's answer was verified by recomputing the recursive
`Depends`/`Imports`/`LinkingTo` closure of the Imports plus testthat and quarto
against the library the run used: closure 62, installed 62, `setdiff` empty.
The Suggests absent are exactly the seven the check's own dependency `INFO`
names — arrow, data.table, dtplyr, duckdb, quartabs, RSQLite, spelling.

Every other claim the file makes was checked against the two logs: the single
NOTE in each run, the environment line, all 13 full-run skips sitting in
`test_that()` blocks that contain `expect_snapshot()` with no direct
`skip_on_cran()` anywhere, the dependency-only run's 215 skips decomposing as
38 + 39 + 8 + 84 + 39 + 7 with each category naming its package, and five
`.qmd` vignettes with no committed `.html`.

PR #496, opened 2026-09-07
State the record's first answer once, where the code decides it (#494)

### LINKED ISSUE #494: The first of `?last_sent_queries`'s four answers states a condition the code does not use
## Blocks

#40.

## The false clause

`R/sent-queries.R:100`, the first bullet of the *Four answers* section:

> - Nothing has been recorded in this session -- **no Margin verb has begun** --
>   and the call is refused with a `"marginplyr_error"`.

An unaudited `inspect_grouping()` is enough to move the answer off that branch,
and `CONTEXT.md`'s *Sent query* entry says `inspect_grouping()` "compiles a plan
without a Margin operation" — so it is not a Margin verb:

```
fresh session:              No Margin operation has run in this session.
                            i Run a Margin verb or `inspect_grouping()` first.

after inspect_grouping():   The last Margin operation was not audited.
                            i Set `options(marginplyr.audit_sql = TRUE)` ...
```

The diagnostic itself names `inspect_grouping()`. So does the same page's own
opening at `R/sent-queries.R:62-64`, which lists `inspect_grouping()` among the
calls that keep a record. Only the bullet is wrong.

## The second half: three copies, and the canonical one drifted

The *Four answers* block is written out three times:

| where | the first bullet says |
|---|---|
| `R/sent-queries.R:100` — the reference | "no Margin verb has begun" |
| `R/marginplyr-package.R:136` | "neither a Margin verb nor `inspect_grouping()` has run" |
| `vignettes/database_backends.qmd:711` | "neither a Margin verb nor `inspect_grouping()` has run" |

`R/marginplyr-package.R:109-148` restates the whole record contract from
`?last_sent_queries` — the two-column tibble, the `"result"` promise, "recorded
before it is sent", the data-frame/dtplyr/Arrow zero-row case, and the four
answers — several sentences verbatim.

`design/architecture.md` *Answering a review* names this as a finding in its own
right: prose that "duplicates an argument another file owns and can drift from
it". It has now drifted, and the copy that drifted is the canonical reference
that #23's User Story 35 makes the source of truth: "I want function references
to be the canonical source of user-visible rules, so that duplicated tables
cannot drift."

## Acceptance criteria

- [ ] `R/sent-queries.R`'s first answer states the condition the code uses. A
      claim found false is deleted rather than restated
      (`design/architecture.md` *Answering a review*).
- [ ] The three copies are reduced to one owner plus references to it, or the
      ticket records why three copies is the right answer here. Note that
      `R/marginplyr-package.R`'s package-doc page and the vignette have
      different audiences from the reference, so "delete both" is not
      automatically right — what is not acceptable is a third independent
      statement of the same contract.
- [ ] Nothing asserts these three stay in step today. Whatever the fix, say at
      the surviving site what would catch the next drift, or accept that
      nothing does and record that.

## How it was found

The Standards (prose) axis of #40's readiness review, against `b3190d2`, and

sayuks:
Closes #494.

## What was wrong

`?last_sent_queries`'s first answer said the record is empty when "no Margin
verb has begun". The code's condition is `sent_queries$recorded`, set by
`reset_sent_queries()`, which every exported function taking `.grouping` calls
— `inspect_grouping()` among them. An unaudited `inspect_grouping()` alone
therefore moves the answer to the second, and the diagnostic itself says so.

## What changed

- **The false clause is deleted, not restated.** The bullet now refers to the
  list of calls the page's description already carries, so the page holds one
  enumeration rather than two, and a call that keeps a record cannot be in one
  and missing from the other.
- **One owner, two references.** `?marginplyr` and *Record the SQL marginplyr
  sends* in the database-backends guide each restated the whole record
  contract — the tibble's columns, the `"result"` promise,
  recorded-before-sent, the zero-row case, and the four answers. Each now keeps
  what its own audience needs (the package page its one-capture argument, the
  guide its walkthrough of the rows it just printed) and points at the
  reference for the contract.
- **Two gates, where nothing held the page to the code before.**
  `test-sent-queries.R` runs every entry point out of `forwarded_verbs` against
  a session with an empty record and asserts the answer it moves to is the
  unaudited one; its control now reads the refusal's message, both answers
  being marginplyr errors. `test-documentation.R` holds the page's enumeration
  to `verbs_taking(".grouping")`, resolving each verb to the topic documenting
  it so that the `summarise_with_margins()` alias is named by its own page.
  Neither can read whether a sentence about the record is true, and the comment
  above the block says so.

## Checks

`lintr::lint_package()`, `roxygen2::roxygenise()`, `spelling::spell_check_package()`,
`verify-context-budget.R`, `verify-must-error.R`, and the full suite under
`NOT_CRAN=true` all pass locally.

### DIFF
--- a/R/marginplyr-package.R
+++ b/R/marginplyr-package.R
@@ -106,46 +106,14 @@
 #' options(old)
 #' ```
 #'
-#' The record is a tibble of two character columns -- `purpose`, what the
-#' query was sent for, and `sql`, the statement as it was rendered -- with one
-#' row per query, in the order the queries were sent. It belongs to a single
-#' call, being emptied at the start of every one.
-#'
-#' `"result"` is the one `purpose` promised: the query a Margin verb returns
-#' unexecuted, which your own [dplyr::collect()] is still what runs. Every
-#' other row is a query marginplyr sends for its own reasons, and the names
-#' those rows carry are not fixed.
-#'
-#' A query is recorded before it is sent, so the record holds what was sent
-#' rather than what succeeded, and a call that fails leaves readable every
-#' query it had already sent.
-#'
-#' What is recorded is the SQL marginplyr sent, not the execution marginplyr
-#' caused. An audited call on a data frame, a dtplyr table, or an Arrow table
-#' therefore records nothing, and that zero-row answer is correct rather than
-#' a hole.
-#'
 #' There is one capture, `dbplyr::sql_render()`. It renders in your own
 #' session and sends nothing the call did not already send, so what
 #' *[When marginplyr queries your data][summarize_with_margins]* enumerates is
 #' the same whether or not you are recording.
 #'
-#' [last_sent_queries()] gives four answers, and each of the three that could
-#' look like an empty record is told apart from the others:
-#'
-#' - Nothing has been recorded in this session -- neither a Margin verb nor
-#'   [inspect_grouping()] has run -- and the call is refused with a
-#'   `"marginplyr_error"`.
-#' - The last call was not audited, the option being unset or holding a value
-#'   other than `TRUE` when the call began, and the call is refused with a
-#'   `"marginplyr_error"` naming `marginplyr.audit_sql`. The option is read
-#'   once, as the call begins; setting it afterwards does not audit a call
-#'   already made.
-#' - The call was audited and sent nothing: a zero-row tibble, the only answer
-#'   with zero rows.
-#' - A statement had no SQL form on this backend -- a translation dbplyr
-#'   refuses when the query is rendered -- and its row holds `NA` in `sql`.
-#'   The call itself is unaffected.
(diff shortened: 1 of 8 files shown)


sayuks:
## Code review — b3190d2...d3e3b5f

### Standards

No findings. The documentation ownership, ADR amendment, test seams, comments, and generated Rd changes conform to the applicable repository standards. git diff --check is clean; generated man files were excluded from review as required.

### Spec

No findings. The canonical reference now uses the actual recorded-state condition and includes inspect_grouping() in the governing call list. The other pages retain audience-specific context and point to that owner. The two new gates document and exercise the remaining drift boundary.

Checks: targeted test-sent-queries.R and test-documentation.R passed (159 assertions total).

Summary: Standards 0 findings (no worst issue); Spec 0 findings (no worst issue).

PR #497, opened 2026-09-07
Qualify internal heads in deferred expressions (#491)

### LINKED ISSUE #491: A factor grouping dimension on dtplyr fails with `could not find function "encode_factor_for_margin"`
## Blocks

#40.

## What happens

Every Margin verb fails on a `dtplyr` input whose Grouping plan names a factor
or ordered-factor dimension, whenever `.margin_label` is not `NULL` — which is
the default:

```r
library(marginplyr)
library(dplyr)

df <- data.frame(g = factor(c("lo", "hi")), v = 1:2)
collect(summarize_with_margins(dtplyr::lazy_dt(df), s = sum(v), .grouping = rollup(g)))
#> Error in encode_factor_for_margin(g, missing_sentinel = "..marginplyr_missing_factor_code",  :
#>   could not find function "encode_factor_for_margin"
```

All four verbs, measured against the installed 0.1.0 at `b3190d2`:

| call | result |
|---|---|
| `summarize_with_margins()` | `ERR: could not find function "encode_factor_for_margin"` |
| `expand_with_margins()` | same |
| `nest_with_margins()` | same |
| `nest_by_with_margins()` | same |
| `summarize_with_margins(.margin_label = NULL)` | OK |
| a character dimension instead of a factor | OK |
| the same factor on the local backend | OK |

## What is claimed

`R/summarize_with_margins.R:222` — rendered into `man/summarize_with_margins.Rd`
and inherited by all four verbs:

> Factor and ordered-factor columns are the one exception, on the backends
> whose factors marginplyr can rebuild: local data frames, `dtplyr`, and
> DuckDB. There it decomposes them to insert `.margin_label` and rebuilds them
> itself, so their levels and ordering are preserved

#23's User Story 2 asks for the same thing: "I want Parent shares to work with
double, logical, Date, factor, and character grouping dimensions according to
their documented contracts, so that grouping-key storage type does not change
correctness."

## Cause

Two sites inject an *unqualified* internal symbol as the head of an expression
`dtplyr` defers:

- `R/margin-label.R:437` — `rlang::expr(encode_factor_for_margin(...))`
- `R/factor.R:133` — `rlang::expr(reconstruct_factor_vector(...))`

Neither function is exported. On the local backend dplyr evaluates the
expression where the symbol resolves; `dtplyr` translates it into a
`data.table` call evaluated in the environment the pipeline was **written** in,
where it does not:

```
build=globalenv collect=globalenv -> ERR
build=globalenv collect=namespace -> ERR
build=namespace collect=globalenv -> OK
build=namespace collect=namespace -> OK
```

The comment at `R/factor.R:127-130` explains why every *argument* is injected
as a value rather than named — a source column could shadow the local (#76).
The function head was not given the same treatment.

## Why no gate catches it

`testthat::test_env("marginplyr")` is `env_clone(asNamespace("marginplyr"))`:

```
$ Rscript -e 'e <- testthat::test_env("marginplyr"); cat(environmentName(parent.env(e)), exists("encode_factor_for_margin", envir = e), "\n")'
imports:marginplyr TRUE
```

so every internal symbol resolves in the environment the tests write
`lazy_dt()` in. `tests/testthat/test-margin-label.R:124` ("dtplyr obeys the
eight-case contract as the local backend does") exercises exactly the failing
combination and passes. No Rd example or vignette covers it either:
`retail_sales` carries no factor column, and the only `factor()` near a dtplyr
chunk (`vignettes/get_started.qmd:575`) is on the local path.

This is why the full suite, the `Live dtplyr` release-matrix job, and both
tarball checks are all green on a tree where the call errors for every user.

## Acceptance criteria

- [ ] `summarize_with_margins()`, `expand_with_margins()`, `nest_with_margins()`,
      and `nest_by_with_margins()` all return the documented result for a
      factor and for an ordered-factor dimension on a `dtplyr` input with the
      default `.margin_label`, with levels and ordering preserved.
- [ ] A regression test that **fails without the fix**. A test written in the
      ordinary way cannot: it must build the `lazy_dt()` input and collect it
      in an environment that does not inherit `asNamespace("marginplyr")` —
      e.g. `local(..., envir = new.env(parent = globalenv()))`. State that
      requirement at the test, because the next reader will otherwise simplify
      it back into a passing no-op.
- [ ] Whatever other unqualified internal symbol is injected into a deferred
      expression is found and fixed in the same pass, not only these two. Scan
      for `rlang::expr(` and `rlang::call2(` heads in `R/`.
- [ ] The test obeys `AGENTS.md` *Release matrix*: it may require `dtplyr` and
      no other member of `optional_backends()`.

## How it was found

The Docs & Tests audit of #40's readiness review, against `b3190d2`, and

sayuks:
Closes #491.

## What was wrong

Every Margin verb failed on a `dtplyr` input whose Grouping plan named a
factor or ordered-factor dimension with a non-missing `.margin_label` —
the default:

```
Error in encode_factor_for_margin(g, ...) :
  could not find function "encode_factor_for_margin"
```

Two sites injected an unexported internal function as the *bare* head of
an expression `dtplyr` defers: `R/margin-label.R` (`encode_factor_for_margin`)
and `R/factor.R` (`reconstruct_factor_vector`). dplyr resolves such a head
below the quosure’s own environment, so the local backend never showed it;
`dtplyr` evaluates its translation in the environment the caller wrote the
pipeline in, where the name is not bound.

## The fix

`share_private_call()` already built a `marginplyr:::` head for exactly this
reason. It moves to `R/utils.R` as `marginplyr_private_call()`, gains the
header its contract needs, and is used at both sites. A scan over every
`rlang::expr()` and `rlang::call2()` reachable from a loaded namespace found
no third unqualified internal head.

## Why no gate caught it

`testthat::test_env("marginplyr")` is `env_clone(asNamespace("marginplyr"))`,
so a test written in the ordinary way builds its `lazy_dt()` where every
internal symbol resolves — which is why the existing eight-case dtplyr
contract test exercises the failing combination and stayed green.

The regression test therefore evaluates in an environment parented *below*
`package:marginplyr` on the search path. `globalenv()`, which is what a user
has, is not enough on its own: `pkgload::load_all()` attaches every internal
name to `package:marginplyr`, so under `devtools::test()` an environment
above it resolves them all. Starting below withholds them either way, and
still reaches `utils::head()`, which dtplyr’s own translation calls.

It covers all four verbs against a factor and an ordered-factor dimension,
and requires `dtplyr` and no other member of `optional_backends()`.

## Checks

- `lintr::lint_package()` — clean
- full suite with `NOT_CRAN=true` — 7013 pass, 0 fail, 0 skip
- `verify-suite-coverage.R`, `verify-verifier-invocation.R`,
  `verify-context-budget.R` — pass
- `roxygen2::roxygenise()` — no diff

### DIFF
--- a/R/factor.R
+++ b/R/factor.R
@@ -127,15 +127,19 @@ reconstruct_factor.data.frame <- function(data,
   # against the data mask first, so a source column called `new_levels`,
   # `ord`, or `missing_sentinel` would supply the argument instead of the
   # local, silently rebuilding the factor from the wrong levels.
+  #
+  # The head is qualified rather than bare for the separate reason
+  # `marginplyr_private_call()` gives: the dtplyr method below shares this
+  # body, and dtplyr defers the expression into an environment where a bare
+  # name is not bound (#491).
   dplyr::mutate(
     .data = data,
-    "{col}" := !!rlang::expr(
-      reconstruct_factor_vector(
-        !!margin_column_pronoun(col),
-        new_levels = !!new_levels,
-        ordered = !!ord,
-        missing_sentinel = !!missing_sentinel
-      )
+    "{col}" := !!rlang::call2(
+      marginplyr_private_call("reconstruct_factor_vector"),
+      margin_column_pronoun(col),
+      new_levels = new_levels,
+      ordered = ord,
+      missing_sentinel = missing_sentinel
     )
   )
 }
--- a/R/factor.R
+++ b/R/factor.R
@@ -128,10 +128,9 @@ reconstruct_factor.data.frame <- function(data,
   # `ord`, or `missing_sentinel` would supply the argument instead of the
   # local, silently rebuilding the factor from the wrong levels.
   #
-  # The head is qualified rather than bare for the separate reason
-  # `marginplyr_private_call()` gives: the dtplyr method below shares this
-  # body, and dtplyr defers the expression into an environment where a bare
-  # name is not bound (#491).
+  # The head is qualified for a separate reason: the dtplyr method below shares
+  # this body, and dtplyr defers the expression
+  # (`marginplyr_private_call()`, #491).
   dplyr::mutate(
     .data = data,
     "{col}" := !!rlang::call2(
--- a/R/margin-label.R
+++ b/R/margin-label.R
@@ -433,9 +433,8 @@ label_margin_branch <- function(.data,
       encoded_factors,
       function(info) {
         col <- info$col
-        # The head is qualified for the reason `marginplyr_private_call()`
-        # gives: dtplyr defers this expression into an environment where a
-        # bare name is not bound (#491).
+        # The head is qualified because dtplyr defers this expression
+        # (`marginplyr_private_call()`, #491).
         rlang::call2(
           marginplyr_private_call("encode_factor_for_margin"),
           margin_column_pronoun(col),
--- a/R/margin-label.R
+++ b/R/margin-label.R
@@ -433,12 +433,14 @@ label_margin_branch <- function(.data,
       encoded_factors,
       function(info) {
         col <- info$col
-        rlang::expr(
-          encode_factor_for_margin(
-            !!margin_column_pronoun(col),
-            missing_sentinel = !!sentinels[[col]],
-            preserve_missing_value = TRUE
-          )
+        # The head is qualified for the reason `marginplyr_private_call()`
+        # gives: dtplyr defers this expression into an environment where a
+        # bare name is not bound (#491).
+        rlang::call2(
+          marginplyr_private_call("encode_factor_for_margin"),
+          margin_column_pronoun(col),
+          missing_sentinel = sentinels[[col]],
+          preserve_missing_value = TRUE
         )
       }
     )
--- a/R/share.R
+++ b/R/share.R
@@ -971,7 +971,7 @@ wrap_share_sources <- function(dots,
         source_summaries
       )
       wrapped <- rlang::call2(
-        share_private_call("check_share_across"),
+        marginplyr_private_call("check_share_across"),
         expr,
         share_outputs = share_outputs,
         share_kinds = share_kinds
@@ -980,7 +980,7 @@ wrap_share_sources <- function(dots,
       check <- checks[[1L]]
       is_dtplyr <- identical(backend_kind, "dtplyr")
       wrapped <- rlang::call2(
-        share_private_call(if (is_dtplyr) {
+        marginplyr_private_call(if (is_dtplyr) {
           "check_dtplyr_share_source"
         } else {
           "check_share_scalar"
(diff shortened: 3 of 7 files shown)


sayuks:
## Review round 1

(b3190d2), two axes in parallel.

### Standards — 4 findings, all fixed in dfa38d4

| finding | what became of it |
|---|---|
| The property is asserted by no gate; the repo has three of this shape in `test-utils.R` and this would be a fourth | Fixed. `test-utils.R` now carries "no deferred expression names an internal function bare", reading the loaded namespace through the shared `namespace_functions()`/`visit_calls()`, with the witness test each gate here carries. Verified end to end: against `origin/main`'s `R/factor.R` and `R/margin-label.R` it names `label_margin_branch`, `reconstruct_factor.data.frame`, and `reconstruct_factor.dtplyr_step`. |
| Both site comments restate `marginplyr_private_call()`'s header beside the call to it (`AGENTS.md` *Code comments*) | Fixed. Each now says which way its own site falls under the citation and stops. |
| Mysterious Name: `factor_user_env_call()` builds no user-env call | Fixed — `factor_call_with_input()`. `marginplyr_private_call()` keeps its inherited name: what it returns is a `:::` call, and the five pre-existing `R/share.R` sites read the same. |
| `sort(as.character(...))` drops `NA` from both sides | Fixed — `na.last = TRUE`. |

`walks_matching()`'s header and `helper-namespace-walk.R`'s list of the
gates reading it both counted three; both now count four.

### Spec (#491) — 0 findings

All four acceptance criteria verified rather than taken on trust:

1. All four verbs × {factor, ordered factor} on dtplyr with the default
   `.margin_label`, levels and ordering preserved, values matching the
   local backend.
2. The regression test fails without the fix under **both** regimes —
   `pkgload::load_all()` and a pre-fix tree installed to a separate
   library. Reverting only `R/factor.R` also fails it, so it covers both
   fixed sites and not just the reproducer's.
3. An independent AST walk over `R/` (skipping `!!` subtrees; covering
   `rlang::expr/quo/exprs/quos`, `quote`, `bquote`, and string-headed
   `rlang::call2`) found exactly the two reported heads on `origin/main`
   and none on `HEAD`. There was no third site.
4. Guards on `dtplyr` alone; passes with every other Suggest hidden.

One note it raised, not a finding: moving `share_private_call()` to
`R/utils.R` as `marginplyr_private_call()` was not asked for by the
issue. It is what keeps the `:::` builder from having a second copy, and
is now also what the new gate's remedy names.

### Checks

`lintr::lint_package()` clean; full suite 7022 pass / 0 fail / 0 skip
under `NOT_CRAN=true`; `verify-suite-coverage.R` 803 tests all covered;
`verify-verifier-invocation.R` and `verify-context-budget.R` pass;
`roxygen2::roxygenise()` no diff.

sayuks:
## Review round 2

(`b3190d2`), two axes in parallel, at `dfa38d4`.

### Standards — 0 findings

The review-response commit preserves the repository standards: the new loaded-namespace gate has both an enumeration/remedy witness and shape witnesses; its shared-helper inventory is updated; the two site comments state only the local consequence of the helper contract; and the dtplyr test uses the registered optional-Suggest guard. Generated artifacts are untouched. No documented-standard breach or material Fowler-baseline smell remains.

### Spec (#491) — 0 findings

All four acceptance criteria are implemented: the two deferred internal heads are qualified; the regression test exercises every public verb for factor and ordered-factor dimensions outside the namespace while preserving the documented factor properties; the loaded-namespace gate covers the required `rlang::expr()` / `rlang::call2()` scan; and the test requires only `dtplyr`. Consolidating the existing share helper into `marginplyr_private_call()` is necessary to keep that qualified-call construction single-sourced, not scope creep.

### Checks

- `testthat::test_local(".", filter = "margin-label")` — pass
- `pkgload::load_all(".", quiet = TRUE); lintr::lint_package()` — clean
- `verify-suite-coverage.R`, `verify-verifier-invocation.R`, `verify-context-budget.R` — pass

PR #500, opened 2026-09-07
Fix portable window ordering

### LINKED ISSUE #493: `.sort = "none"` clears a window ordering on the portable UNION ALL path, where the reference says it clears none
## Blocks

#40.

## What the reference promises

`R/summarize_with_margins.R:312`, rendered into
`man/summarize_with_margins.Rd`'s *Margin order* section and inherited by every
verb:

> asking for a Margin order discards any window ordering the input carried.
> `.sort = "none"` records no order and clears none.

## What happens

`tests/testthat/test-margin-order.R:698` asserts exactly that sentence. Running
it verbatim with only the simulated dialect changed:

```r
library(marginplyr)
dat <- data.frame(region = c("East", "West"), units = c(1, 2))

probe <- function(sim) {
  remote <- dbplyr::window_order(dbplyr::tbl_lazy(dat, con = sim), region)
  q <- summarize_with_margins(
    remote, units = sum(units, na.rm = TRUE),
    .grouping = grouping_spec(region), .sort = "none"
  )
  w <- suppressWarnings(dbplyr::sql_render(dplyr::mutate(q, running = cumsum(units))))
  grepl("OVER (ORDER BY", w, fixed = TRUE)
}

probe(dbplyr::simulate_postgres())  #> TRUE
probe(dbplyr::simulate_sqlite())    #> FALSE
probe(dbplyr::simulate_mysql())     #> FALSE
```

On the portable `UNION ALL` path the window ordering the input carried is gone,
and a window function written over the result has no order. On live SQLite a
following `cumsum()` then raises dbplyr's own

> Windowed expression `SUM(`s`)` does not have explicit order.

`expand_with_margins()` behaves as documented on both paths; this is specific to
`summarize_with_margins()` on the `UNION ALL` adapter — that is, on every
backend *Database backend coverage* lists as portable.

## Why no gate catches it

Both window-ordering tests — `tests/testthat/test-margin-order.R:408` and
`:698` — use `dbplyr::simulate_postgres()`, a native `GROUPING SETS` dialect,
which is the one path where the claim holds. The portable path is never
asserted and behaves the opposite way.

`dbplyr::simulate_sqlite()` opens no connection and reads no optional package
beyond what `AGENTS.md` *Release matrix* already exempts, so a mirror test costs
nothing and satisfies `verify-suite-coverage.R`.

## The decision this needs

1. **Fix the adapter** so `.sort = "none"` carries the input's window ordering
   through the union, making the sentence true everywhere. Whether dbplyr can
   express that across a `UNION ALL` without an ordering subquery per branch is
   the open question, and it is what makes this `ready-for-human`.
2. **Narrow the sentence** to say that a portable-path result carries no window
   ordering, and say so where *Database backend coverage* draws the same line.
   Then `.sort = "none"` stops being described as inert.

Option 2 is a smaller change and is honest; option 1 is what the current
sentence promises. Nothing here returns a wrong value — the failure mode is a
non-deterministic window function downstream, which dbplyr warns about on the
backends that warn at all.

## Acceptance criteria

- [ ] The disposition above is recorded on this ticket before implementation.
- [ ] A test asserts the portable path's behaviour, whichever way it is
      settled, alongside the existing `simulate_postgres()` one — so the pair
      records that the two paths were compared.
- [ ] `?summarize_with_margins` *Margin order* and the code agree.

## How it was found

The Docs & Tests audit of #40's readiness review, against `b3190d2`, and

sayuks:
## Summary

- preserve usable input window ordering through the portable UNION ALL summary path
- keep only terms that remain in the summary result, matching dbplyr native behavior
- cover native and portable rollup paths without adding a physical ORDER BY

## Verification

- `pkgload::load_all(".", quiet = TRUE); lintr::lint_package()`
- `devtools::test()` (7,030 passed)

Closes #493

### DIFF
--- a/R/grouping-adapter-union.R
+++ b/R/grouping-adapter-union.R
@@ -386,7 +386,8 @@ summarize_margin_union <- function(.data,
                                    reserved_names,
                                    set_id_name = NULL,
                                    set_id_is_internal = FALSE,
-                                   call = NULL) {
+                                   call = NULL,
+                                   input_window_order = list()) {
   dots <- summaries$dots
   group_vars <- unique(c(plan$by, plan$dimensions))
   key_names <- new_margin_internal_names(
@@ -523,7 +524,28 @@ summarize_margin_union <- function(.data,
     plan$set_ids
   )
 
-  combine_margin_branches(branches)
+  restore_input_window_order(
+    combine_margin_branches(branches),
+    input_window_order
+  )
+}
+
+# Restores the input's usable dbplyr window ordering after `UNION ALL` has
+# dropped it. ADR 0018's `.sort = "none"` amendment and #493 decide the
+# contract; terms the result cannot name follow the native adapter and vanish.
+restore_input_window_order <- function(result, input_window_order) {
+  if (length(input_window_order) == 0L) {
+    return(result)
+  }
+  result_names <- get_col_names(result, dplyr::everything())
+  usable <- Filter(
+    function(term) all(all.vars(term) %in% result_names),
+    input_window_order
+  )
+  if (length(usable) == 0L) {
+    return(result)
+  }
+  dbplyr::window_order(result, !!!usable)
 }
 
 # `backend` is the operation's own, and it sits among the required arguments
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -1058,6 +1058,11 @@ stage_margin_summaries <- function(operation,
           set_id_is_internal = set_id_is_internal
         )
       } else {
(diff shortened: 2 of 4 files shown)


sayuks:
Review: `b3755c7...7cd06c1`.

- Standards: no findings.
- Spec: the initial review found that the portable test used one grouping set. The implementation now uses `rollup(region)` for both simulator paths, verifies no physical `ORDER BY` under `.sort = "none"`, and the follow-up review found no remaining issues.

Verification: `lintr::lint_package()` and `devtools::test()` (7,030 passed).

PR #509, opened 2026-09-09
Clarify Arrow input contract

sayuks:
## Summary
- document the verb-specific contract for every Arrow input category
- refuse direct `RecordBatchReader` inputs and reader-backed Arrow queries before reusable Margin branches can return incomplete results
- keep reader-backed queries inspectable without consumption and track direct-reader inspection in #510
- align general non-tabular input diagnostics and add public-seam regression coverage

## Verification
- full testthat suite
- targeted Arrow admission, backend, and utility suites
- `jarl check .`
- `lintr::lint_package()` after `pkgload::load_all()`
- roxygen documentation generation
- context-budget, must-error, verifier-invocation, document-reference, and suite-coverage verifiers
- `git diff --check`

### DIFF
--- a/R/expand_with_margins.R
+++ b/R/expand_with_margins.R
@@ -98,7 +98,7 @@ expand_with_margins <- function(.data,
   with_margin_error_call(
     {
       assert_margin_input(.data)
-      assert_lazy_table(.data)
+      assert_reusable_margin_input(.data)
     },
     call = call
   )
--- a/R/expand_with_margins.R
+++ b/R/expand_with_margins.R
@@ -32,6 +32,7 @@
 #' @inheritSection summarize_with_margins Grouping set identifiers
 #' @inheritSection summarize_with_margins Margin order
 #' @inheritSection summarize_with_margins Display labels and grouping identity
+#' @inheritSection summarize_with_margins Arrow inputs
 #' @inheritSection summarize_with_margins Database backend coverage
 #' @inheritSection summarize_with_margins When marginplyr queries your data
 #' @return An ungrouped data frame, or a lazy table when `.data` is lazy. Its
--- a/R/grouping-backend.R
+++ b/R/grouping-backend.R
@@ -13,10 +13,6 @@ arrow_input_classes <- function() {
 # methods. Queries keep their left source in `.data` and any right sources in
 # the join and union nodes; walking all three is what prevents wrapping a
 # one-shot reader in a query from bypassing the reusable-input refusal.
-#
-# This reproduces Arrow's source graph rather than calling its unexported
-# `all_sources()` helper. The public-verb tests pin each node shape so a change
-# in Arrow's query representation fails at the refusal boundary.
 arrow_input_has_reader_source <- function(.data) {
   if (is.null(.data)) {
     return(FALSE)
--- a/R/grouping-backend.R
+++ b/R/grouping-backend.R
@@ -1,10 +1,43 @@
 # Named because two readings need it and a second copy of the vector is what
 # would drift: the kind below, and the branch summary's Absorbing-backend
 # handler, which has only `.data` to decide from and must not answer for a
-# warning some other backend raised (ADR 0025). `RecordBatchReader` is
-# deliberately absent, as it is from the kind: no verb accepts one.
+# warning some other backend raised (ADR 0025). A direct `RecordBatchReader` is
+# deliberately absent because no verb accepts it. A query backed by one still
+# has the class below; each verb applies its narrower reader-source rule.
 arrow_input_classes <- function() {
   c("arrow_dplyr_query", "Table", "RecordBatch", "Dataset")
+}
+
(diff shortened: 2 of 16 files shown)


sayuks:
Review of 2292947 (merge-base with main)

Standards
- No documented-standard violations.
- Observation (judgement call): the Arrow input builders in test-verb-argument-admission.R overlap helper-arrow-shapes.R. A future input-shape change could update only one suite.

Spec
- Medium: public docs group non-tabular Arrow objects together, but the requested all-types audit has no direct rejection coverage for Scanner, Scalar, Array, ChunkedArray, Schema, Field, or DataType.
- Low: docs promise both nesting verbs reject Arrow inputs, while the new matrix covers nest_with_margins() but not nest_by_with_margins().
- No scope creep found.

sayuks:
Review follow-up: the missing nest_by_with_margins() coverage from Spec finding 2 is fixed in bdea765. Spec finding 1 remains a design question: non-tabular Arrow objects already use the same general admission error as R vectors and matrices; RecordBatchReader intentionally has a distinct error because it is a valid Arrow dplyr source with a one-pass/materialization constraint.

sayuks:
Review follow-up

- Addressed the non-tabular Arrow coverage finding by documenting and testing Scanner, Scalar, Array, ChunkedArray, Schema, Field, and DataType through the shared admission refusal.
- Replaced the generic as.data.frame()/dplyr::tbl() remedy with a truthful destination contract, because those functions do not convert every rejected Arrow object.
- Confirmed the RecordBatchReader remedy with a two-batch fixture: refusal leaves all 5 rows unconsumed, as_arrow_table() materializes all batches, and the converted input returns complete group and grand-total aggregates.
- The dedicated RecordBatchReader refusal remains: Margin operations build multiple grouping-set branches from one reusable source, which a one-shot reader cannot provide.
- The nest_by_with_margins() matrix finding was addressed in bdea765.

sayuks:
Review of 601b7ec (merge-base with `main`)

## Standards

- **Hard violation:** the new amendment in `design/adr/0012-distinguish-factor-na-levels-from-missing-margin-values.md` re-argues the change and repeats measured evidence. Under `design/architecture.md` (*Repository placement*), an ADR amendment records the current decision and what changed, while the ticket owns the reason; the dated investigation note owns the reproduction evidence. Keep the contract change and cite the ticket/PR and investigation note instead.
- **Judgement call — Duplicated Code:** `tests/testthat/test-verb-argument-admission.R` repeats the same multi-batch `RecordBatchReader` factory in two adjacent tests. A local helper would prevent the batch boundary and expected row count from drifting.

## Spec

- **Implemented-looking but wrong:** the documented multi-branch reason does not justify rejecting `RecordBatchReader` in `inspect_grouping()`, which compiles and formats a plan but executes no grouping-set branches. Raw readers currently fail later in the selection proxy, but wrapping the reader in an unexecuted Arrow query allows inspection without consuming it. A schema-only admission path should be evaluated; this is a new issue candidate, and the public explanation should distinguish inspection from summary/expansion.
- **Missing/partial:** the requested concrete wrong-result example is only described in prose. There is no runnable two-query `union_all()` reproduction or regression test. Without one, an Arrow release that makes the reader safely shareable would not show that the refusal premise has disappeared.
- **Missing/partial:** the general-input regression covers matrix/list/NULL but not the explicitly requested character and numeric vectors. Both currently reach the unified general error, but that contract is not pinned by the new test.

No scope creep found.

Summary: Standards has 1 hard violation and 1 judgement call; its worst issue is duplicated ticket/note-owned reasoning in the ADR. Spec has 3 findings; its worst issue is the unsupported rationale for refusing `RecordBatchReader` in `inspect_grouping()`.

sayuks:
Review of 736a384 and follow-up 0b47ae8 (merge-base with `main`)

## Standards

- Hard: `arrow_input_has_reader_source()` initially described what another test covered, contrary to the repository's code-comment rule. Fixed in 0b47ae8 by leaving only the function contract and source-graph invariant.
- Judgement call — Duplicated Code: direct-reader and reader-query refusal tests repeated the same verb matrix and assertions. Fixed in 0b47ae8 by combining them into one input-shape matrix.
- Follow-up verification found 0 remaining Standards findings.

## Spec

- No missing or partial requirements.
- No scope creep.
- No implemented-looking-but-wrong behaviour found.
- The review confirmed the full input/verb contract, reader-source traversal, non-consumption guarantees, multi-branch reproduction, conversion remedies, and the #510 split.

Summary: Standards raised 2 findings, both resolved (0 remaining); Spec raised 0 findings.

sayuks:
CI follow-up: the working-tree failure came from one stale expand-operation expectation for the previous RecordBatchReader diagnostic. Commit 1355f7e aligns that regression test with the new reader-dependent refusal message. The targeted tests, full local test suite, both required linters, and the complete PR check matrix now pass, including Live arrow and Suite coverage by construction.

PR #517, opened 2026-09-09
Align inspect_grouping() with non-operation semantics

### LINKED ISSUE #513: Align inspect_grouping() failures with its non-operation semantics
Part of #512.

## Problem

`inspect_grouping()` rejects a dtplyr step rooted at `dtplyr::lazy_dt(immutable = FALSE)` through the mutable-step guard shared with Margin operations. Its diagnostic explains the rejection by saying that a Margin verb builds one branch per grouping set and data.table writes each branch to the caller's table.

That causal explanation is false for inspection. `inspect_grouping()` is documented to compile and format a Grouping plan without executing a Margin operation, and it builds no grouping-set branch. Bypassing only `mutable_dtplyr_step()` lets the real inspection path return the expected plan without mutating the source table.

The related `last_sent_queries()` unaudited-call diagnostic also says “The last Margin operation was not audited” after `inspect_grouping()`, even though the accessor deliberately tracks inspection as a non-operation call.

## Evidence

- `R/grouping-plan.R:197-207` applies the mutable-step guard on the shared preparation path.
- `R/grouping-backend.R:122-134` explains the refusal only in terms of branch construction and reference writes.
- `R/inspect-grouping.R:3-5` says inspection executes no Margin operation.
- `R/inspect-grouping.R:147-175` shows that inspection only prepares and formats the plan.
- ADR 0029 deliberately put the guard where `prepare_margin_operation()` and `inspect_grouping()` both reach it; its scope must be amended.
- ADR 0020's `RecordBatchReader` amendment and #510 establish the applicable boundary: inspection may accept an input that is unsafe only when reused across Margin branches because inspection builds no branch.
- `R/sent-queries.R:183-198` calls the most recent tracked call a Margin operation.
- `CONTEXT.md`, *Sent query*, explicitly says `inspect_grouping()` compiles a plan without a Margin operation.
- Existing snapshots pin the inconsistent mutable-step message for `inspect_grouping()`.

## Scope

- Mutable-step validation placement or scoping for `inspect_grouping()`.
- The `last_sent_queries()` diagnostic vocabulary after inspection.
- ADR 0029 and any documentation that claims every `.grouping` entry point refuses a mutable step.
- Structural tests that distinguish the five branch-building Margin verbs from `inspect_grouping()`.

## Non-goals

- Accepting mutable dtplyr roots in `summarize_with_margins()`, `summarise_with_margins()`, `expand_with_margins()`, `nest_with_margins()`, or `nest_by_with_margins()`.
- Fixing dtplyr's upstream multi-reference mutation hazard.
- Copying the caller's table implicitly.
- Changing sent-query recording behavior.

## Proposed direction

- Allow `inspect_grouping()` to prepare and format a plan from a mutable dtplyr root.
- Keep the conservative root-based refusal unchanged for every operation that builds Margin branches.
- Amend ADR 0029 to state the operation boundary instead of the current shared-entry-point boundary.
- Use “last tracked call” or another term covering the documented call set in `last_sent_queries()` diagnostics.

## Acceptance criteria

- [ ] `inspect_grouping()` accepts a root and a derived dtplyr step whose root was built with `immutable = FALSE`.
- [ ] The returned plan matches the plan for the equivalent immutable input.
- [ ] Inspection leaves the caller's `data.table` unchanged in names, columns, rows, and values.
- [ ] Inspection invokes no execution entry point catalogued by `test-query-policy.R`.
- [ ] All five branch-building Margin verbs continue to reject mutable roots before any branch is built.
- [ ] The mutable-step coverage is derived structurally so a future public Margin verb cannot arrive untested.
- [ ] `last_sent_queries()` no longer calls an `inspect_grouping()` call a Margin operation.
- [ ] ADR 0029, snapshots, reference text, and tests state the same boundary.

## Relevant files, tests, and docs

- `R/grouping-plan.R`
- `R/grouping-backend.R`
- `R/inspect-grouping.R`
- `R/sent-queries.R`
- `CONTEXT.md`
- `design/adr/0029-refuse-a-step-a-backend-would-write-to.md`
- `design/adr/0020-ask-before-reading-a-lazy-input.md`
- `tests/testthat/helper-margin-verbs.R`
- `tests/testthat/test-grouping-backends.R`
- `tests/testthat/test-inspect-grouping.R`
- `tests/testthat/test-query-policy.R`
- `tests/testthat/test-sent-queries.R`
- `tests/testthat/_snaps/grouping-backends.md`

## Dependencies

No implementation blocker. The inspection boundary is decided in #512. Related upstream work remains tracked separately in #468 and does not block this issue.

sayuks:
Closes #513

## Summary
- allow inspect_grouping() to inspect mutable dtplyr roots without changing the caller's data.table
- keep all five branch-building Margin verbs structurally covered by the mutable-step refusal
- describe unaudited inspection as the last tracked call
- amend ADR 0029 and the generated reference documentation

## Verification
- full testthat suite
- jarl check .
- lintr::lint_package()
- context-budget and document-reference verifiers

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -192,7 +192,10 @@ same field value and fall on opposite sides. A Margin verb refuses one before
 any branch is built, because it builds one branch per grouping set from the
 same step and data.table writes each of them to the caller's table by
 reference — leaving a result in which every row carries the Margin label, and
-a table whose columns were dropped or whose names were permuted.
+a table whose columns were dropped or whose names were permuted. Inspection
+builds no branch and accepts one; where a typed selection needs a proxy, it is
+compiled against an isolated zero-row root so the caller's table stays
+unchanged.
 _Avoid_: Mutable input, in-place backend
 
 **Sent query**:
--- a/R/backend-metadata.R
+++ b/R/backend-metadata.R
@@ -21,6 +21,20 @@ zero_row_dtplyr_proxy_input <- function(.data) {
   step
 }
 
+# The typed selection proxy for Mutable-step inspection. The registered dtplyr
+# method evaluates only the isolated zero-row root above and reaches none of
+# ADR 0020's catalogued execution entry points; calling the method directly is
+# what avoids routing the step through `collect()` or `as_tibble()`.
+mutable_dtplyr_selection_proxy <- function(.data) {
+  proxy <- utils::head(zero_row_dtplyr_proxy_input(.data), n = 0L)
+  as_data_table <- getS3method(
+    "as.data.table",
+    "dtplyr_step",
+    envir = asNamespace("data.table")
+  )
+  as_data_table(proxy)
+}
+
 grouping_selection_proxy <- function(.data,
                                      backend = grouping_backend(.data)) {
   if (identical(backend$kind, "arrow")) {
--- a/R/backend-metadata.R
+++ b/R/backend-metadata.R
@@ -4,6 +4,23 @@ get_col_names <- function(data, ...) {
   as.character(dplyr::tbl_vars(selected))
 }
 
+# A copy of a Mutable step rooted in a zero-row table, used only to acquire an
+# inspection selection proxy. The derived step may already hold a reference-
+# writing `:=` call, so changing its root's permission is too late; replacing
+# the copied root's source gives that call an isolated schema-only table to
+# write to. ADR 0029 records why Margin operations refuse rather than use this
+# rewrite.
+zero_row_dtplyr_proxy_input <- function(.data) {
+  stopifnot(inherits(.data, "dtplyr_step"))
+  step <- .data
+  if (inherits(step[["parent"]], "dtplyr_step")) {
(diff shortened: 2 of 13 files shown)


sayuks:
Review record for 8aef909 (merge-base with main: 7f00983).

## Standards

Round 1 raised one judgement-call finding: the Margin-verb forwarder subset was duplicated across two tests. It was centralized in tests/testthat/helper-margin-verbs.R in 8aef909. Follow-up review found the duplication resolved and no hard standards violations or new smells.

## Spec

Round 1 raised two findings: typed Mutable-step inspection still reached a catalogued execution entry point, and the name-only optimization changed DuckDB inspection's selection-proxy record. In 8aef909, typed root/derived inspection is covered against the execution catalog, and the special name-only path is limited to Mutable inspection with a regression test preserving DuckDB's selection_proxy row. Follow-up review found both resolved and no new spec regression.

Summary: Standards 0 open findings; Spec 0 open findings.

PR #519, opened 2026-09-09
Clarify grouped-input remediation

### LINKED ISSUE #515: Make grouped-input remediation explicit about lost grouping semantics
Part of #512.

## Problem

The shared grouped-input validation uses the same remediation — “Call `dplyr::ungroup()` first” — for three different conditions:

1. row-wise input;
2. grouped input created with `.drop = FALSE`; and
3. grouped input that also supplies `.by`.

The third is valid when the caller is deliberately replacing the existing groups with `.by`. The first two do not have the same semantics.

For `.drop = FALSE`, literal ungrouping removes the existing grouping columns as implicit fixed keys and removes the empty-level behavior that triggered the refusal. Adding the former keys back with `.by` still does not restore empty groups. The error disappears because the operation's meaning changed.

For `rowwise(data, account)`, literal ungrouping also discards the row-wise identifier grouping. Whether the intended rewrite is an ungrouped calculation or `.by = account` cannot be inferred from the call, so the diagnostic must not imply that ungrouping preserves meaning.

## Evidence

- `R/grouping-plan.R:68-96` emits the same remediation from all three validation branches.
- `R/summarize_with_margins.R`, *Grouped and row-wise inputs*, defines existing groups as implicit fixed keys.
- `tests/testthat/test-grouping-interface.R:990-1028` asserts the refusals but does not execute the remediation.
- The audit reproduced `.drop = FALSE` with a declared but empty factor level: literal `ungroup()` removed the fixed key from the result; `ungroup()` plus `.by` retained the key but not the empty level.
- ADR 0011 assigns missing-key completion to explicit pre-operation `tidyr::complete()` or key-relation workflows and rejects a hidden completion interface.

## Scope

- The three grouped/row-wise diagnostics in `normalize_grouping_input()`.
- Documentation of replacement groups, discarded row-wise grouping, and empty-key completion.
- Public-seam remediation tests for all affected Margin verbs and `inspect_grouping()` where appropriate.

## Non-goals

- Supporting `.drop = FALSE` directly.
- Adding `.complete`, `.fill`, `.empty`, or a completion verb.
- Inferring whether a caller wants to retain row-wise identifier columns.
- Changing the valid grouped-input-plus-`.by` replacement semantics.

## Proposed direction

- Keep “ungroup first” for grouped input plus `.by`, explicitly as replacement of existing groups.
- For `.drop = FALSE`, explain that marginplyr cannot preserve empty groups. Present removal of empty-group semantics and explicit pre-operation completion as distinct caller choices; do not name bare `ungroup()` as a semantics-preserving fix.
- For row-wise input, explain that `ungroup()` discards row-wise grouping. Put a documentation example showing how named row-wise identifiers can be retained explicitly with `.by` after ungrouping.

## Acceptance criteria

- [ ] The `.drop = FALSE` diagnostic no longer presents bare `ungroup()` as an unconditional remedy.
- [ ] Its documentation distinguishes discarding empty groups from completing keys before the Margin operation and points to the existing ADR 0011 workflow.
- [ ] A test applies each documented `.drop = FALSE` route and checks fixed keys, empty keys, and aggregate values.
- [ ] The row-wise diagnostic makes the loss of row-wise grouping explicit.
- [ ] Documentation demonstrates the `.by` rewrite for a row-wise input with identifier columns.
- [ ] A test distinguishes literal ungrouping from ungrouping plus explicit `.by`.
- [ ] Grouped input plus `.by` retains its current valid replacement guidance and behavior.
- [ ] All affected errors remain `marginplyr_error` conditions blamed on the public call.

## Relevant files, tests, and docs

- `R/grouping-plan.R`
- `R/summarize_with_margins.R`
- `vignettes/get_started.qmd`
- `vignettes/completing_keys.qmd`
- `design/adr/0011-keep-key-completion-outside-margin-operations.md`
- `tests/testthat/test-grouping-interface.R`
- `tests/testthat/test-inspect-grouping.R`
- `tests/testthat/helper-margin-verbs.R`

## Dependencies

No implementation blocker. The grouped-input remediation decisions are recorded in #512.

sayuks:
Closes #515.\n\n## Summary\n- distinguish replacement, discarded row-wise grouping, and empty-key completion in grouped-input diagnostics\n- document the row-wise .by rewrite and ADR 0011 completion workflow\n- add public-seam remediation regression coverage\n\n## Validation\n- devtools::test(reporter = "summary")\n- jarl check .\n- lintr::lint_package()\n- Rscript .github/scripts/verify-context-budget.R\n- Rscript .github/scripts/verify-must-error.R\n- Rscript .github/scripts/verify-verifier-invocation.R

### DIFF
--- a/R/grouping-plan.R
+++ b/R/grouping-plan.R
@@ -71,14 +71,19 @@ normalize_grouping_input <- function(.data, by_quo) {
   if (inherits(.data, "rowwise_df")) {
     abort_marginplyr(c(
       "{.fun rowwise} input is not supported.",
-      i = "Call {.fun dplyr::ungroup} first."
+      i = "Calling {.fun dplyr::ungroup} discards row-wise grouping."
     ))
   }
 
   if (!dplyr::group_by_drop_default(.data)) {
     abort_marginplyr(c(
       "Grouped input created with {.code .drop = FALSE} is not supported.",
-      i = "Call {.fun dplyr::ungroup} first."
+      i = "marginplyr cannot preserve empty groups.",
+      i = "Call {.fun dplyr::ungroup} to discard empty groups.",
+      i = paste0(
+        "To retain empty keys, complete the input before the ",
+        "Margin operation."
+      )
     ))
   }
 
@@ -86,7 +91,10 @@ normalize_grouping_input <- function(.data, by_quo) {
   if (length(input_groups) > 0L && !rlang::quo_is_null(by_quo)) {
     abort_marginplyr(c(
       "Can't supply {.arg .by} when {.arg .data} is grouped.",
-      i = "Call {.fun dplyr::ungroup} first."
+      i = paste0(
+        "Call {.fun dplyr::ungroup} first to replace the existing ",
+        "groups with {.arg .by}."
+      )
     ))
   }
 
--- a/R/summarize_with_margins.R
+++ b/R/summarize_with_margins.R
@@ -156,9 +156,12 @@
 #' also supply `.by`; call [dplyr::ungroup()] first when replacing the existing
 #' groups. A grouping column also cannot appear in `.grouping`, because one
 #' column cannot be both a fixed key and a dimension that can be rolled up.
-#' Grouped local data created with `.drop = FALSE` is rejected because empty
-#' factor groups do not have a consistent equivalent in SQL and other lazy
-#' backends.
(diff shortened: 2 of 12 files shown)


sayuks:
## Code review

Reviewed commit `6e4a878` against merge-base `591d280` (`git diff main...HEAD`).

- Standards: 0 findings
- Spec: 0 findings

No findings required a disposition.

PR #532, opened 2026-09-10
Fix typed-missing SQL dimensions across UNION ALL order

### LINKED ISSUE #527: RSQLite loses typed-missing dimension types when a Grand total branch is first
## Problem

On RSQLite's portable `UNION ALL` path, reordering semantically equivalent
grouping-set occurrences can change an all-missing Margin dimension's
collected type. When the Grand total branch is first, a source `character`
dimension is collected as `logical`; with the detail branch first, it remains
`character`.

This violates the documented typed-missing Margin-label contract and makes a
Grouping-plan ordering choice silently alter the public result schema.

## Minimal reproduction

```r
on.exit(DBI::dbDisconnect(con))

remote <- dplyr::copy_to(
  con,
  data.frame(a = character()),
  "type_repro",
  temporary = TRUE
)

detail_first <- summarize_with_margins(
  remote,
  .grouping = grouping_sets(grouping_set(a), grouping_set()),
  .margin_label = NULL
) |>
  dplyr::collect()

total_first <- summarize_with_margins(
  remote,
  .grouping = grouping_sets(grouping_set(), grouping_set(a)),
  .margin_label = NULL
) |>
  dplyr::collect()

c(detail_first = typeof(detail_first$a), total_first = typeof(total_first$a))
# detail_first total_first
#   "character"   "logical"
```

The failure is deterministic. It also occurs with a one-row
`NA_character_` source, integer and double all-missing dimensions,
`.margin_label = NA_character_`, and `expand_with_margins()`. A non-missing
source value or non-missing Margin label hides it.

## Contract

- `R/summarize_with_margins.R` says that `NULL` and `NA_character_` insert
  typed missing values and that the portable `UNION ALL` path has the same
  semantics as confirmed native SQL.
- `R/grouping-spec.R` defines `grouping_sets()` as a union.
- ADR 0012 requires an omitted dimension to receive a typed missing value.
- Any remedy must preserve ADR 0020's ban on unrequested reads of lazy inputs.

## Cause boundary established by the investigation

The generic `sql` backend supplies no source prototype. An omitted dimension
therefore becomes bare SQL `NULL`, and `summarize_margin_union()` combines
branches in plan order. RSQLite gives the compound result source-column
affinity when the pass-through branch is first, but reports an all-NULL column
as logical when the `NULL` branch is first. `CAST(NULL AS TEXT)` alone did not
restore the collected type in a direct probe.

## Acceptance criteria

- Add a public-seam regression test on live RSQLite for both orders of
  `grouping_sets(grouping_set(a), grouping_set())`.
- For zero-row and one-row all-missing character sources, both orders collect
  the dimension as `character` with equal values.
- Cover both `NULL` and `NA_character_` Margin labels and both
  `summarize_with_margins()` and `expand_with_margins()`.
- Do not compare occurrence identifiers or unspecified physical order.
- Preserve the no-unrequested-read query policy in ADR 0020 and the existing
  semantics on local, dtplyr, DuckDB, and Arrow paths.
- Keep the regression at the public verb seam; do not assert opaque Margin
  operation internals.

## Existing-test gap

The closest live RSQLite test uses non-missing character values and a
detail-first rollup. Simulated fallback tests inspect SQL shape without
collecting an all-missing result, and the live SQLite Margin-label test uses a
non-missing `"Total"` label. The full existing test suite passes while this
reproducer fails.

## Severity

Medium. Values and row counts remain correct, but the public result schema is
silently corrupted and can break downstream string operations, type-stable
joins, schema assertions, and empty-result pipelines.

sayuks:
Closes #527.\n\nAdds a zero-row SQL type anchor so generic SQL UNION ALL results retain source dimension types regardless of grouping-set occurrence order. Includes live RSQLite regression coverage for both public verbs, both typed-missing labels, and zero-row / all-missing character inputs.


sayuks:
Two-axis review completed against main...HEAD at 56c7f81.

- Standards: no documented-standard violations. One low-confidence Duplicated Code judgement was considered; the summary and expansion call sites deliberately retain distinct source-column choices, so no change was made.
- Spec: no findings; all Issue #527 acceptance criteria are covered.

Validated with testthat::test_local("."), jarl check ., pkgload::load_all(".", quiet = TRUE); lintr::lint_package(), git diff --check, and the context-budget verifier.

PR #535, opened 2026-09-10
test: add share conservation metamorphic coverage

### LINKED ISSUE #530: Add share-conservation metamorphic tests
## Problem

Existing share tests assert selected expected ratios, but do not state the
broader conservation and scale-invariance properties implied by the Parent and
Total share contracts. A defect in denominator partitions or join keys can
therefore evade a small collection of hand-authored ratios.

The investigation in #528 checked these properties on deterministic generated
inputs without finding a violation.

## Scope

Add public-seam metamorphic tests showing that:

- Total shares sum to one within each fixed-key and grouping-occurrence
  denominator partition;
- Parent shares sum to one beneath each immediate parent in a pure rollup;
- multiplying every positive source measure by the same nonzero constant does
  not change Parent or Total shares; and
- duplicating every source row doubles additive summaries but does not change
  the shares.

Include fixed and variable missing keys and duplicate rows. Use positive
integer measures for conservation assertions so zero and missing-denominator
contracts remain separate example-based cases.

## Acceptance criteria

- Tests exercise the public share helpers through
  `summarize_with_margins()`.
- Local, dtplyr, RSQLite, and DuckDB execution paths are represented; Arrow is
  excluded because the public contract rejects share operations there.
- An RSQLite test uses `.check_share_source = FALSE` only after the test has
  established the source as an eligible `sum(integer)` expression.
- Occurrence and fixed-key partitions are explicit in the assertions.
- No individual test requires more than one member of `optional_backends()`.
- Tests are deterministic, and failures identify the relation and backend.

## Evidence

See relation families 18--21 and Recommended durable metamorphic tests in
#528.

sayuks:
Closes #530

### DIFF
--- a/tests/testthat/test-metamorphic-margin-semantics.R
+++ b/tests/testthat/test-metamorphic-margin-semantics.R
@@ -582,3 +582,183 @@ test_that("RSQLite preserves typed missing values across grouping-set order", {
     }
   }
 })
+
+# The positive integer measure keeps conservation separate from the missing- and
+# zero-denominator examples in `test-share.R`. The first and last rows are a
+# duplicate, and both the fixed and variable keys carry a missing value.
+share_metamorphic_input <- function() {
+  data.frame(
+    fixed = c("north", "north", NA_character_, NA_character_, "north"),
+    region = c("east", NA_character_, "west", "west", "east"),
+    store = c("a", "a", NA_character_, "b", "a"),
+    measure = c(2L, 3L, 5L, 7L, 2L)
+  )
+}
+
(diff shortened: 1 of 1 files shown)


sayuks:
## Code review

Reviewed snapshot: 0c11993 with git diff main...0c11993.

## Standards

- Fixed in d4c6644: the RSQLite source-type precondition did not include the metamorphic failure context. Its assertion now carries source case, relation, API, and backend.

## Spec

- No findings.

Summary: Standards raised one finding, fixed in d4c6644; Spec raised no findings.

PR #544, opened 2026-09-12
Fix Arrow Margin ordering with duplicate grouping bits

### LINKED ISSUE #541: Make Arrow Margin ordering robust to duplicate Grouping-bit expressions
## Problem

Arrow rejects an opted-in Margin order when two plan columns have identical
varying Grouping-bit expressions. The smallest supported case is a true
two-column composite dimension.

The public contract in `CONTEXT.md`, ADR 0018, and #85 says that an opted-in
Margin order works across supported grouping structures and backends. #85
specifically requires a composite dimension's columns to order together; its
columns share one Grouping bit. Issue #87 implemented that contract, but its
Arrow test uses `rollup(c(region, store))`, which creates two scalar dimensions
rather than the documented composite spelling
`rollup(grouping_set(region, store))`.

## Minimal reproduction

```r
library(marginplyr)

x <- arrow::Table$create(
  data.frame(a = character(), b = character())
)

expand_with_margins(
  x,
  .grouping = rollup(grouping_set(a, b)),
  .sort = "last"
) |>
  dplyr::collect()
```

Expected: a zero-row collected result, as local execution returns.

Actual:

```text
Invalid: Invalid sort key column: Multiple matches for
FieldRef.Name(dplyr::if_else(.data[["..marginplyr_sort_1"]] %in% 2L, 1L, 0L))
```

The failure also occurs with `.sort = "first"`, populated input, and
`summarize_with_margins()`. `.id` and caller-visible Grouping-bit summaries are
not required.

## Diagnosis

`margin_order_terms()` in `R/margin-operation.R` emits one bit term per plan
column. Columns in a composite dimension have the same omitted-occurrence ids,
so their expressions deparse to the same Arrow field name. Arrow materializes
both expressions and refuses the ambiguous sort field. Explicit grouping sets
can trigger the same failure when independently declared dimensions happen to
have the same nonconstant inclusion mask.

## Scope

- Affected: `summarize_with_margins()`, `summarise_with_margins()`, and
  `expand_with_margins()`; Arrow; `.sort = "first"` or `"last"`.
- Unaffected: `.sort = "none"`; local, dtplyr, RSQLite, and DuckDB in the
  property sweep; Arrow plans whose varying bit expressions are unique.
- Workaround: omit `.sort`, or collect and order after the Margin operation.
- Severity: medium; a documented backend/API/option combination is unusable,
  but ordering is opt-in and has a workaround.

## Why the current suite misses it

The Arrow test named `Arrow orders a composite factor dimension under
expansion` uses `rollup(c(region, store))`. Public inspection shows three
rollup sets and two scalar dimensions. The true composite spelling produces
two sets and the duplicate expression that Arrow rejects. No existing test
generates arbitrary inclusion masks with repeated nonconstant bit patterns.

## Acceptance criteria

- [ ] `.sort = "first"` and `"last"` execute on Arrow for
      `rollup(grouping_set(a, b))` through both expansion and summary.
- [ ] The collected order agrees with the local Margin-order contract on empty
      and populated inputs.
- [ ] An explicit `grouping_sets()` case covers two independently declared
      dimensions with the same nonconstant inclusion mask.
- [ ] The result remains lazy until the caller collects it.
- [ ] Internal sort fields remain unambiguous and absent from the public
      result.
- [ ] Existing native and portable ordering behavior is unchanged on other
      backends.
- [ ] Focused public-seam regression coverage uses the documented nested
      `grouping_set()` composite spelling.

No blockers are known.

sayuks:
Closes #541\n\n## Verification\n\n- Focused Arrow Margin-order regression test\n- Review-ready check: cfb7e6081619ddacbb60ad7e87016bb74a59ddde\n  - full testthat suite: passed\n  - jarl: passed\n  - package-aware lintr: passed\n  - source-tarball R CMD check: 0 errors, 0 warnings, 0 notes\n\n## Review\n\nPending two-axis review.

### DIFF
--- a/R/margin-operation.R
+++ b/R/margin-operation.R
@@ -504,9 +504,15 @@ margin_order_terms <- function(plan,
     recursive = FALSE
   )
 
+  seen_margin_ids <- list()
   for (dimension in plan$dimensions) {
-    bit <- margin_grouping_bit_expr(plan, dimension, sort_id)
-    if (!is.null(bit)) {
+    margin_ids <- margin_grouping_bit_ids(plan, dimension)
+    if (
+      !is.null(sort_id) &&
+        !is.null(margin_ids) &&
+        !any(vapply(seen_margin_ids, identical, logical(1), margin_ids))
+    ) {
+      bit <- margin_grouping_bit_expr(sort_id, margin_ids)
       terms <- c(terms, list(
         if (identical(sort, "first")) {
           rlang::expr(dplyr::desc(!!bit))
@@ -514,6 +520,7 @@ margin_order_terms <- function(plan,
           bit
         }
       ))
+      seen_margin_ids <- c(seen_margin_ids, list(margin_ids))
     }
     terms <- c(
       terms,
@@ -553,13 +560,9 @@ margin_missing_last_expr <- function(column) {
   ))
 }
 
-# One dimension's Grouping bit, read from the Grouping set identifier the
-# adapter left in the result. `NULL` when the plan makes the bit constant, so
-# that a term with nothing to order by is not emitted.
-margin_grouping_bit_expr <- function(plan, dimension, sort_id) {
-  if (is.null(sort_id)) {
-    return(NULL)
-  }
+# The Grouping-set occurrences where one dimension is absent. `NULL` when the
+# plan makes its Grouping bit constant, so a key term would order nothing.
+margin_grouping_bit_ids <- function(plan, dimension) {
   margin_ids <- as.integer(
     plan$set_ids[plan$grouping_masks[, dimension] == 1L]
   )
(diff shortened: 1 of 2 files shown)


sayuks:
## Two-axis review

Reviewed cfb7e6081619ddacbb60ad7e87016bb74a59ddde against main with git diff main...HEAD.

- Standards: 0 findings. The change follows ADR 0018: equal nonconstant Grouping-bit occurrence sets contribute one redundant-free sort term; no lazy-input execution policy or documented coding standard is breached.
- Spec: 0 findings. The regression coverage exercises the documented composite spelling, same-mask explicit grouping sets, empty/populated inputs, both sort choices, summary/expansion, lazy results, local-order agreement, and no public internal fields.

No findings required a disposition.

PR #547, opened 2026-09-12
Test local NaN Parent-share numerators

### LINKED ISSUE #545: Test local NaN share numerators normalize to NA_real_
## Problem

The current local Parent-share value test does not distinguish `NaN` from
`NA_real_`, so it survives this mutation in the R-side share value expression
built by `apply_joined_shares()`:

```r
is.na(source) | is.na(denominator) | denominator == 0
```

becomes:

```r
is.na(source) & is.na(denominator) | denominator == 0
```

For a local `NaN` numerator and a finite nonzero denominator, the production
expression returns the promised `NA_real_`; the mutant divides instead and
returns `NaN`. Both values satisfy `is.na()`, so the existing suite passes.

This is a test gap, not a known production defect: the unmutated implementation
currently returns the contract value.

## Contract

- ADR 0010 states that a missing numerator produces `NA_real_` and explicitly
  treats local `NaN` as missing.
- `?share_of_parent` repeats that missing numerators and local `NaN` follow the
  missing-value rule.
- #27 required local `NaN` to follow the documented Parent-share value table.

## Minimal reproducer

```r
result <- summarize_with_margins(
  data.frame(group = c("a", "b")),
  total = if (dplyr::n() == 1L) NaN else 2,
  share = share_of_parent(total),
  .grouping = rollup(group),
  .margin_label = NULL
)

detail <- result$share[!is.na(result$group)]
```

Observed output:

- unmutated implementation: `c(NA_real_, NA_real_)`;
- mutant: `c(NaN, NaN)`.

The closest existing test, "Parent-share values handle missing, zero, negative,
and empty summaries", creates `NA_real_` rather than `NaN` and asserts only
`is.na()`. Those choices admit the mutant.

## Acceptance criteria

- Add a local Parent-share test beside the existing value-rule test.
- Use the minimal two-row scenario above, or an equally small public-interface
  scenario, with `NaN` only in detail source summaries and a finite parent
  source summary.
- Assert exact `NA_real_` output, for example with
  `expect_identical(detail, c(NA_real_, NA_real_))`; the test must reject
  `NaN`, not merely accept any missing double.
- Confirm that the test fails under the first-`|` to `&` mutant described
  above and passes against the unmutated implementation.
- Do not duplicate the same case for Total shares unless the implementation
  paths have first become distinct; both helpers currently share this value
  expression.
- Run the repository's Review-ready check for the tests-only commit.

## Scope and severity

Tests only; no production-code change is expected. Severity is medium: callers
can observe `NaN` versus `NA_real_` through `is.nan()`, serialization, and
downstream reporting even though both satisfy `is.na()`.

sayuks:
Closes #545

## Verification

- Focused share tests: passed (924 assertions).
- Mutation check: the specified `|` to `&` mutant fails the new test with `NaN` where `NA_real_` is expected.
- Review-ready check: `cd2b7b11b08e842ece89f8e2631dc1c75c8c0e00`
  - full testthat suite: passed
  - jarl: passed
  - package-aware lintr: passed
  - source-tarball R CMD check: 0 errors, 1 warning; failed only because Bioconductor package-index requests timed out. The warning is external to this tests-only change.

### DIFF
--- a/tests/testthat/test-share.R
+++ b/tests/testthat/test-share.R
@@ -324,6 +324,19 @@ test_that(paste0(
   expect_identical(partitioned$share, double())
 })
 
+test_that("local NaN Parent-share numerators normalize to NA_real_", {
+  result <- summarize_with_margins(
+    data.frame(group = c("a", "b")),
+    total = if (dplyr::n() == 1L) NaN else 2,
+    share = share_of_parent(total),
+    .grouping = rollup(group),
+    .margin_label = NULL
+  )
+
+  detail <- result$share[!is.na(result$group)]
+  expect_identical(detail, c(NA_real_, NA_real_))
+})
+
 test_that(paste0(
   "Parent matching ignores display labels and preserves ",
   "expression order"


sayuks:
## Code review — `main...cd2b7b1`

### Standards

No findings. The test-only hunk follows the applicable repository standards; no documented-standard breach or baseline smell was found.

### Spec

No findings. The new test is adjacent to the existing Parent-share value-rule test, uses the requested minimal two-row public-interface scenario, asserts exact `NA_real_`, and adds neither Total-share duplication nor production changes.

Summary: Standards 0 findings (none); Spec 0 findings (none).

Issue #551, opened 2026-09-12
Make Parent occurrence discovery scale linearly for rollups


sayuks:
## Problem Statement

Parent-share planning discovers the Parent occurrence for every child occurrence by comparing it with every Grouping-set occurrence and performing a set membership test for each candidate. For the accepted pure-rollup shape, the compiled occurrences are already ordered from detailed to coarse, so this all-pairs search does work the structure does not require.

The 2026-09-12 performance investigation isolated the discovery step from backend work. Its median increased from 0.00042 seconds at 17 occurrences to 0.113 seconds at 257 occurrences, consistent with the implementation's approximately `O(S^2 * D)` candidate and membership work.

## Solution

Derive each distinct rollup occurrence's Parent in a forward pass over the already ordered compiled plan, skipping retained duplicate occurrences that are not strictly less detailed. Preserve the existing Parent semantics and backend-independent interface while removing the all-pairs candidate search.

## User Stories

1. As a user with a very wide rollup, I want Parent-target planning to follow the rollup sequence, so that planning does not dominate a small result.
2. As a user retaining duplicate Grouping-set occurrences, I want every occurrence to select the same strictly less detailed Parent it selects today, so that occurrence identity remains stable.
3. As a user on any supported backend, I want the same Parent-target mapping, so that optimizing the backend-independent planner does not alter results.
4. As a maintainer, I want the algorithmic improvement expressed through deterministic structure or operation counts, so that no wall-clock threshold enters the test suite.
5. As a maintainer, I want this optimization isolated from lazy SQL mapping construction, so that each scalability problem can be reviewed and reverted independently.

## Implementation Decisions

- Parent discovery continues to consume the compiled pure-rollup plan and return one Parent occurrence identifier or missing value per occurrence.
- The algorithm uses the plan's detailed-to-coarse ordering and treats only a strictly less detailed distinct occurrence as a Parent.
- Retained duplicate occurrences are skipped as Parent candidates when they represent the same Grouping set.
- Backend adapters and Parent denominator query construction are unchanged by this ticket.
- No public API, error contract, or result ordering changes.

## Testing Decisions

- Preserve existing public Parent-share semantic coverage across local and lazy backends.
- Add focused deterministic cases for single-occurrence rollups, repeated occurrences, composite rollup units, and the Grand total occurrence.
- Verify candidate visits with scratch instrumentation or another stable algorithmic count when validating the rewrite.
- Do not add elapsed-time assertions to ordinary tests.

[...]

PR #552, opened 2026-09-12
Bound lazy Parent-share query growth

### LINKED ISSUE #550: Stop Parent-share SQL from repeating staged Margin summaries
## Problem Statement

A Parent share on a lazy SQL backend builds one mapping branch for each child Grouping-set occurrence. Each branch refers to the staged Margin summary independently, so dbplyr renders the complete staged summary into every branch. The query therefore grows faster than the Parent mapping it represents, even though the backend may later factor the repeated aggregate into a common physical plan.

The 2026-09-12 performance investigation reproduced this on live DuckDB rollups. From 5 to 33 Grouping-set occurrences, rendered SQL grew from 7.4 KB to 911 KB and the number of staged `GROUPING SETS` copies grew from 5 to 33. The corresponding Total share retained two staged-summary references.

## Solution

Represent the Parent occurrence mapping for SQL backends as one small, non-materializing inline plan relation and join it to the staged summary once. Keep the staged summary reference count constant as the rollup grows. The returned value remains a composable lazy query, with no caller-data materialization introduced during construction.

## User Stories

1. As a user summarizing a deep hierarchy on a SQL backend, I want Parent-share query construction to grow with the mapping rather than repeat the complete aggregate, so that query rendering and submission remain practical.
2. As a user working with a remote database, I want Parent shares to remain lazy, so that constructing the operation never executes or materializes my result unexpectedly.
3. As a user combining Parent shares with later dplyr verbs, I want the returned relation to remain composable, so that the optimization does not change the public workflow.
4. As a user relying on duplicate-safe Grouping-set identity, I want the inline mapping to preserve occurrence identifiers exactly, so that Parent denominators remain correct.
5. As a user of either native or portable SQL aggregation, I want the same Parent-share semantics, so that backend strategy changes query shape but not results.
6. As a maintainer, I want a structural regression assertion rather than a machine-dependent timing threshold, so that the test protects the owned cost pattern without becoming flaky.
7. As a maintainer, I want the backend-specific mapping decision recorded with the existing Parent-share adapter decision, so that a later refactor does not restore repeated staged queries accidentally.
8. As a maintainer investigating future performance changes, I want the dated measurements and reproducible scratch harness retained, so that later results can be compared with the original evidence.

## Implementation Decisions

- SQL Parent mapping may differ from row-matched mapping by constructing a non-materializing inline relation containing child occurrence identifiers and their Parent occurrence identifiers.
- The inline relation is joined to the staged summary once to obtain all denominator rows. It does not use `compute()` or any other execution/materialization entry point.
- The complete staged summary may appear a constant number of times in the final rendered statement, matching the qualitative growth of Total shares; it must not appear once per Grouping-set occurrence.
- Ratio calculation, denominator reuse, one-join-per-share-kind behavior, cleanup, and common Margin finalization remain shared.
- Local, dtplyr, and other row-matched behavior remain unchanged.
- The existing Parent-share adapter ADR receives a dated amendment because this deliberately relaxes its earlier decision that every backend shares the same mapping construction.
- The dated performance investigation and its scratch harness are included with the implementation pull request but are not wired into ordinary tests or CI.

## Testing Decisions

- Test through the public summary operation and its rendered lazy result, not by asserting the private query-node representation.
- Exercise both a live DuckDB native-aggregation path and an RSQLite portable-aggregation path.
- Compare Parent-share results with the established local semantics across representative rollups and missing grouping values.
- Assert that staged-summary definitions or source-reference families remain constant as Grouping-set occurrence count grows, that Parent mapping still uses one join, and that construction does not materialize the result.
- Reuse the repository's existing sent-query and query-policy seams where they express the contract.
- Do not assert SQL byte totals, elapsed time, or backend plan text verbatim.

## Acceptance Criteria

- [ ] Default Parent-share SQL no longer repeats the complete staged Margin summary once per child occurrence.
- [ ] The staged-summary definition/reference family is bounded independently of rollup size on both native DuckDB and portable RSQLite cases.
- [ ] Parent-share results remain equivalent to local results for the covered rollups and missing-value cases.
- [ ] Same-kind Parent shares still use one denominator mapping and one join.
- [ ] Lazy construction adds no caller-data materialization or execution beyond the already permitted typed metadata proxy.
- [ ] The Parent-share adapter decision records the SQL inline-plan exception and why a custom always-CTE query node was not chosen.
- [ ] The dated investigation report and reproducible scratch harness are committed without adding a runtime dependency or CI job.
- [ ] Package-affecting local checks and the repository's two-axis code review pass before merge.

## Blocked by

None (can start immediately).

## Out of Scope

- Replacing the separate all-pairs Parent occurrence discovery algorithm.
- Adding wall-clock thresholds to CRAN checks or ordinary unit tests.
- Ranking backend engines or claiming behavior for unavailable network databases.
- Introducing an eagerly materialized temporary table.

## Further Notes

The alternative of forcing CTE rendering through a custom dbplyr lazy-query node can reduce the aggregate to one textual definition, but it carries broader dbplyr-internal and dialect compatibility risk. The selected inline-plan design targets the confirmed superlinear repetition while preserving normal lazy composition.

sayuks:
## Summary

- carry the SQL Parent occurrence mapping as one non-materializing inline relation instead of one staged-summary branch per child occurrence
- keep the staged summary at two definitions across rollup sizes on native DuckDB and at two portable branch families on RSQLite
- preserve one final Parent-share join, lazy construction, and local/dtplyr mapping behavior
- amend the Parent-share adapter decision and architecture description
- retain the dated performance investigation and reproducible scratch harness

## Performance evidence

The live DuckDB `D = 4, 8, 16, 32` rollup series changed from 7.4, 25.5, 135, and 911 KB of Parent-share SQL to 4.9, 9.7, 24.8, and 76.8 KB. `GROUPING SETS` definitions and CTE scans remain at two throughout; the physical base scan remains one. Elapsed observations remain investigation evidence only and are not test thresholds.

## Testing

- Review-ready check: `be10e7f578c7d11f7a7b95cb2b358d8f5b32f58c`
  - full testthat suite: passed
  - jarl: passed
  - package-aware lintr: passed
  - source-tarball R CMD check: 0 errors, 0 warnings, 0 notes
- focused `test-share-backends.R`, `test-share.R`, `test-branch-union.R`, and `test-query-policy.R`: passed

## Scope

The separate all-pairs Parent occurrence discovery cost remains tracked by #551 and is not changed here.

Closes #550

### DIFF
--- a/R/share.R
+++ b/R/share.R
@@ -2922,14 +2922,11 @@ build_lazy_parent_mapping <- function(result,
                                       plan,
                                       set_id_name,
                                       used_names) {
-  group_vars <- unique(c(plan$by, plan$dimensions))
-  key_exprs <- lapply(group_vars, margin_column_pronoun)
-  names(key_exprs) <- group_vars
-  denominator_exprs <- lapply(
-    sources,
-    function(source) margin_column_pronoun(source)
+  mapping_exprs <- parent_mapping_exprs(
+    plan,
+    sources = sources,
+    denominator_names = denominator_names
   )
-  names(denominator_exprs) <- unname(denominator_names[sources])
 
   mappings <- lapply(
     child_ids,
@@ -2945,13 +2942,25 @@ build_lazy_parent_mapping <- function(result,
       )
       dplyr::transmute(
         parent_rows,
-        !!!key_exprs,
+        !!!mapping_exprs$keys,
         !!!child_id_expr,
-        !!!denominator_exprs
+        !!!mapping_exprs$denominators
       )
     }
   )
   combine_margin_branches(mappings)
+}
+
+# The staged-summary projections both Parent mapping strategies share. The
+# caller inserts the child occurrence identifier between the keys and the
+# denominators because its expression differs between the two strategies.
+parent_mapping_exprs <- function(plan, sources, denominator_names) {
+  group_vars <- unique(c(plan$by, plan$dimensions))
+  key_exprs <- lapply(group_vars, margin_column_pronoun)
+  names(key_exprs) <- group_vars
+  denominator_exprs <- lapply(sources, margin_column_pronoun)
+  names(denominator_exprs) <- unname(denominator_names[sources])
+  list(keys = key_exprs, denominators = denominator_exprs)
 }
 
 # A SQL Parent mapping is a relation carried by the query, not one branch per
@@ -2989,20 +2998,20 @@ build_dbplyr_parent_mapping <- function(result,
     by = stats::setNames(parent_id_name, set_id_name)
   )
 
-  group_vars <- unique(c(plan$by, plan$dimensions))
-  key_exprs <- lapply(group_vars, margin_column_pronoun)
-  names(key_exprs) <- group_vars
+  mapping_exprs <- parent_mapping_exprs(
+    plan,
+    sources = sources,
+    denominator_names = denominator_names
+  )
   child_id_expr <- stats::setNames(
     list(margin_column_pronoun(child_id_name)),
     set_id_name
   )
-  denominator_exprs <- lapply(sources, margin_column_pronoun)
-  names(denominator_exprs) <- unname(denominator_names[sources])
   dplyr::transmute(
     parent_rows,
-    !!!key_exprs,
+    !!!mapping_exprs$keys,
     !!!child_id_expr,
-    !!!denominator_exprs
+    !!!mapping_exprs$denominators
   )
 }
 
--- a/R/share.R
+++ b/R/share.R
@@ -2071,7 +2071,8 @@ execute_row_matched_shares <- function(operation,
     plan = operation$plan,
     set_id_name = set_id_name,
     kind = kind,
-    sql_join = FALSE
+    sql_join = FALSE,
+    parent_mapping_builder = build_lazy_parent_mapping
   )
 }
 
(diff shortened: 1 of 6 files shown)


sayuks:
## Code review

Fixed point: `main` at `325573f3ae2837dac4e2f99ba43537da572fdab2`  
Reviewed snapshot: `51b869bd734e1176df1285d22a5287ee4baf2d57`

### Standards

Two findings:

1. The investigation and ADR used `Grouping-set` outside a House term, contrary to `CONTEXT.md`. Fixed to `grouping-set` in `be10e7f578c7d11f7a7b95cb2b358d8f5b32f58c`.
2. The two Parent mapping strategies repeated their shared key and denominator projection preparation. Extracted the private `parent_mapping_exprs()` preparation seam in the same commit.

No further lazy-input execution, adapter-selection, or R comment-contract finding was raised.

### Spec

No findings. The review found no missing or partial requirement, scope creep, or implemented-but-wrong behavior against #550.

### Verification of dispositions

`test-share-backends.R` and package-aware lintr passed after the fixes. The complete review-ready check passed on `be10e7f578c7d11f7a7b95cb2b358d8f5b32f58c`: full testthat, jarl, and package-aware lintr passed; source-tarball `R CMD check --as-cran` completed with 0 errors, 0 warnings, and 0 notes.

